PR #4, opened 2024-09-02
Add Rent sysvar type

Ozodimgba:
Added `Rent` sysvar, still missing some implementations like default, calculate burn...


febo:
Looking good, I think you are currently missing adding the module declaration to `sysvars/mod.rs`

Ozodimgba:
Oh I thought I added that. i was rushing out as communicated

PR #7, opened 2024-09-05
fix rent min balance: account for storage overhead

billythedummy:
Was under-reporting due to account overhead.

Reference: https://github.com/anza-xyz/agave/blob/c8685ce0e1bb9b26014f1024de2cd2b8c308cbde/sdk/program/src/rent.rs#L74-L78


febo (approved):
Thanks!

PR #8, opened 2024-09-05
CPI first pass impl

billythedummy:
Tested out with some actual programs CPI-ing system program.

Feel free to critique design

### DIFF
--- a/sdk/src/cpi.rs
+++ b/sdk/src/cpi.rs
@@ -1,0 +1,140 @@
+// Reference: https://github.com/anza-xyz/agave/blob/e207c6e0eaf8e1657fbfaff07da05ca6a7928349/programs/sbf/rust/ro_modify/src/lib.rs#L105-L129
+
+use crate::{
+    account_info::AccountInfo, instruction::Instruction, program_error::ProgramError,
+    syscalls::sol_invoke_signed_c,
+};
+
+/// The struct expected to be pointed to by `sol_invoke_signed_c()`'s first arg.
+///
+/// The u64 fields are raw pointers.
+///
+/// DO NOT EXPOSE THIS STRUCT -
+/// to ensure pointers are valid upon use, the scope of this struct should
+/// only be limited to the stack where sol_invoke_signed_c happens and then
+/// discarded immediately after
+#[derive(Debug)]
+#[repr(C)]
+struct SolInstruction {
+    program_id_addr: u64,
+    accounts_addr: u64,
+    accounts_len: usize,
+    data_addr: u64,
+    data_len: usize,
+}
+
+// Reference: https://github.com/anza-xyz/agave/blob/c8685ce0e1bb9b26014f1024de2cd2b8c308cbde/programs/sbf/rust/ro_modify/src/lib.rs#L122-L132
+impl<'p, 'a, 'd> From<Instruction<'p, 'a, 'd>> for SolInstruction {
+    #[inline]
+    fn from(
+        Instruction {
+            program_id,
+            accounts,
+            data,
+        }: Instruction,
+    ) -> Self {
+        Self {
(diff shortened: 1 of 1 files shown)


sdk/src/cpi.rs
```diff
@@ -0,0 +1,140 @@
...
+            rent_epoch: 0, // dont care
+        }
+    }
+}
+
+pub fn invoke_signed<'a, const MAX_ACCOUNTS_INCL_PROGRAM: usize>(
+    ix: Instruction,
+    program_account_info: &'a AccountInfo,
+    account_infos: impl IntoIterator<Item = &'a AccountInfo>,
+    signers: &[&[&[u8]]],
```
febo, line 111: Wwe can't pass the rust references directly to the C code, since it expects the length of each individual seed unfortunately. I started adding signer "types" on this PR: https://github.com/febo/pinocchio/pull/11/files#diff-c14b0096da794786d77f37e2667c1f5cf800e7bf7875f8e99059e14857a03cf4R107-R127

sdk/src/cpi.rs
```diff
@@ -0,0 +1,140 @@
...
+                Ok((arr, len + 1))
+            },
+        )?;
+    let sol_ix = SolInstruction::from(ix);
+    let result = unsafe {
+        sol_invoke_signed_c(
+            &sol_ix as *const _ as *const _,
+            &sol_account_infos as *const _ as *const _,
+            sol_account_infos_len,
+            signers as *const _ as *const _,
```
febo, line 132: We can't pass the rust references directly to the C code, since it expects the length of each individual seed unfortunately. I started adding signer "types" on this PR: https://github.com/febo/pinocchio/pull/11/files#diff-c14b0096da794786d77f37e2667c1f5cf800e7bf7875f8e99059e14857a03cf4R107-R127

sdk/src/cpi.rs
```diff
@@ -0,0 +1,140 @@
...
+            owner_addr: value.owner().as_ptr() as u64,
+            is_signer: value.is_signer(),
+            is_writable: value.is_writable(),
+            executable: value.executable(),
+            rent_epoch: 0, // dont care
+        }
+    }
+}
+
+pub fn invoke_signed<'a, const MAX_ACCOUNTS_INCL_PROGRAM: usize>(
```
febo, line 107: I think we need to add the borrow checks: https://github.com/anza-xyz/agave/blob/master/sdk/program/src/program.rs#L255-L269

We could have an `unsafe *_unchecked` variation without the checks.

febo (commented):
Looking good, thanks for preparing this!

I had started to add the types required for CPI in a separate PR (#11) so perhaps we can combine the effort.

sdk/src/cpi.rs
billythedummy: btw do you know why the https://github.com/anza-xyz/agave/blob/master/sdk/program/src/program.rs#L255-L269 is doing this n^2 thing of looking for the corresponding AccountInfo for a given AccountMeta? Aren't both the AccountInfo and AccountMeta array supposed to be in the same order?

sdk/src/cpi.rs
febo: For reference: https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/cpi.rs#L596-L610

billythedummy:
> @billythedummy What do you think if we use the types from #11? We could merge #11 and then rebase this PR.

Sure thing, please go ahead with any changes you're planning to make and feel free to merge or adapt all my open PRs; unfortunately I won't have a lot of time to dedicate to this until at least the end of breakpoint, apologies

febo:
Closing this as the changes suggested on this PR were added to #11 – thank you for your contribution @billythedummy.

PR #9, opened 2024-09-05
find_program_address()

billythedummy:


sdk/src/pubkey.rs
```diff
@@ -13,3 +13,21 @@ pub fn log_pubkey(pubkey: &Pubkey) {
...
+            seeds as *const _ as *const u8,
+            seeds.len() as u64,
+            program_id as *const _ as *const u8,
+            &mut bytes as *mut _ as *mut u8,
+            &mut bump_seed as *mut _,
+        )
+    };
+    match result {
+        crate::entrypoint::SUCCESS => (bytes, bump_seed),
+        _ => panic!("PDA cannot be found"),
```
febo, line 31: Should we use the same message from `solana_program`, e.g., `"Unable to find a viable program address bump seed"`?

febo (commented):
Looks great, thanks! I have one minor nit.

Also, I was wondering whehter it is worth adding a `try_find_program_address` or not, which returns a `Result` instead of panic. We could either do it in this PR or in a separate one.

sdk/src/pubkey.rs
Ozodimgba: I think we should keep error logs consistent

Ozodimgba:
I think we should add it in this PR and  think we should try and keep things as they are and only reinvent the wheel where absolutely necessary, so we don't inadvertently steepen the learning curve for devs. Docs are also necessary as we build

febo (approved):
Approved - the little nits will be done on a separate PR.

PR #10, opened 2024-09-05
Make `AccountMeta.pubkey` public

billythedummy:
As a lib consumer, there was no way to construct an `AccountMeta` struct previously because there were no constructor functions and AccountMeta.pubkey was private


febo:
Closing this as #11 already incorporate changes to the `AccountMeta` type. In general, we should avoid exposing raw pointer fields since they do not have an explicit associated lifetime.

PR #25, opened 2024-10-11
docs: Using Github callout in README

kquirapas:
# Change

Used a Github callout in the README

Screenshot 2024-10-12 at 7 48 56 AM

Can change it to the one you most prefer from the ones below:

Screenshot 2024-10-12 at 7 49 42 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ pub fn process_instruction(
 }
 ```
 
-> ⚠️ **Note:**
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


kquirapas:
A

> [!WARNING]
> Body

> [!IMPORTANT]
> Body

Might be suitable

febo:
It currently does not use the GitHub annotation since the `README.md` is the same one that gets publishes on [crates.io](https://crates.io/crates/pinocchio), which does not support it. If we were to use it, it will not be formatted and it will render as `[!WARNING]`.

febo:
Closing this as it seems we cannot use it at the moment.

PR #26, opened 2024-10-12
Implement token programs

Jac0xb:


febo:
@Jac0xb Should we close this one since it seems that #31 already have these?

sonicfromnewyoke:
@febo I really like ATA-part of that MR, so in case you decide to close it, I would like to open a dedicated MR with ATA-related logic/improvements

febo:
> @febo I really like ATA-part of that MR, so in case you decide to close it, I would like to open a dedicated MR with ATA-related logic/improvements

Good point! So we could re-purpose this PR to include only the ATA, or close it and create one just for ATA.

cc: @Jac0xb

febo:
Closed in favour of #36

PR #31, opened 2024-10-30
Added Optimized Spl-token program

L0STE:
Added most of the instruction from spl-token program using unsafe rust for optimization w/ @deanmlittle 

Some example on how the instruction are constructed: 
- For data (trasfer_checked in this occasion): 
```
let mut instruction_data = MaybeUninit::<[u8; 10]>::uninit();

// Populate data
unsafe {
    let ptr = instruction_data.as_mut_ptr() as *mut u8;
    // Set discriminator as u8 at offset [0]
    *ptr = 12;
    // Set amount as u64 at offset [1]
    *(ptr.add(1) as *mut u64) = self.amount;
    // Set decimals as u8 at offset [9]
    *ptr.add(9) = self.decimals;
}

    program_id: &crate::ID,
    accounts: &account_metas,
    data: unsafe { &instruction_data.assume_init() },
};
```
- For accounts (transfer_checked): 
```
let account_metas: [AccountMeta; 4] = [
    AccountMeta::writable(self.from.key()),
    AccountMeta::readonly(self.mint.key()),
    AccountMeta::readonly_signer(self.authority.key()),
];
```

To test it out and as an example on how to integrate, I created a repo that have all the CPI to this program and all the tests that actually look if the data is saved correctly at this link: https://github.com/L0STE/pinocchio-spl-examples


programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,66 @@
...
+pub struct Mint(*const u8);
+
+impl Mint {
+    pub const LEN: usize = 82;
+
+    pub fn from_account_info_unchecked(account_info: &AccountInfo) -> Self {
+        unsafe { Self(account_info.borrow_data_unchecked().as_ptr()) }
+    }
+
+    pub fn from_account_info(account_info: &AccountInfo) -> Self {
```
febo, line 11: I am not sure about this pattern. The `borrow_data_unchecked` does not keep track of the borrows but this is not obvious from the method name `from_account_info`.

I wonder if this should use `try_borrow_data` instead, although I see that the `unchecked` in this case refers to the length and owner checks.

programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /// Use this when you know the account will have a mint authority and you want to skip the Option check.
+    #[inline(always)]
+    pub fn mint_authority_unchecked(&self) -> Pubkey {
+        unsafe { *(self.0.add(4) as *const Pubkey) }
+    }
+
+    pub fn supply(&self) -> u64 {
+        unsafe { *(self.0.add(36) as *const u64) }
```
febo, line 37: I think this results in UB since the pointer won't be aligned to `mem::align_of::<u64>()`.

programs/token/src/state/token.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Use this when you know the AccountInfo will have the correct owner and account space
+    #[inline(always)]
+    pub fn from_account_info_unchecked(account_info: &AccountInfo) -> Self {
+        unsafe { Self(account_info.borrow_data_unchecked().as_ptr()) }
+    }
+
+    pub fn from_account_info(account_info: &AccountInfo) -> Self {
+        assert_eq!(account_info.data_len(), Self::LEN);
+        assert_eq!(account_info.owner(), &ID);
+        Self::from_account_info_unchecked(account_info)
```
febo, line 24: Same here about using `try_borrow_data` instead.

programs/token/src/state/token.rs
```diff
@@ -0,0 +1,103 @@
...
+            Some(self.native_amount_unchecked())
+        } else {
+            None
+        }
+    }
+
+    /// Use this when you know the account is native and you want to skip the Option check.
+    #[inline(always)]
+    pub fn native_amount_unchecked(&self) -> u64 {
+        unsafe { *(self.0.add(113) as *const u64) }
```
febo, line 78: Same here about UB since the pointer won't be aligned.

programs/token/src/state/token.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+    
+    pub fn delegated_amount(&self) -> u64 {
+        unsafe { *(self.0.add(121) as *const u64) }
```
febo, line 82: Pointer alignment here.

programs/token/src/instructions/transfer_checked.rs
```diff
@@ -0,0 +1,76 @@
...
+        // -  [0]: instruction discriminator 
+        // -  [1..9]: amount 
+        // -  [9]: decimals
+        let mut instruction_data = MaybeUninit::<[u8; 10]>::uninit();
+
+        // Populate data
+        unsafe {
+            let ptr = instruction_data.as_mut_ptr() as *mut u8;
+            // Set discriminator as u8 at offset [0]
+            *ptr = 12;
```
febo, line 57: Unfortunately this is UB. The reason is that when you assign a value to `*ptr`, it will call `drop` on the previous value so creating a reference to uninitialized memory is UB:

https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html#method.as_mut_ptr

I think it would be fine to do the static allocation of the array:
```
let mut instruction_data = [0u8; 10];
```

programs/token/src/instructions/set_authority.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub account: &'a AccountInfo,
+
+    /// Authority of the Account.
+    pub authority: &'a AccountInfo,
+
+    /// The type of authority to update.
+    pub authority_type: AuthorityType,
+
+    /// The new authority
+    pub new_authority: Option<Pubkey>,
```
febo, line 33: I think it would be better to use a reference to `Pubkey` to avoid one extra copy when setting the value.

programs/token/src/instructions/initialize_mint_2.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// ### Accounts:
+///   0. `[WRITABLE]` Mint account
+pub struct InitilizeMint2<'a> {
+    /// Mint Account.
+    pub mint: &'a AccountInfo,
+    /// Decimals.
+    pub decimals: u8,
+    /// Mint Authority.
+    pub mint_authority: Pubkey,
```
febo, line 17: Here as well, it would be better to use a reference to `Pubkey` instead to avoid an extra copy.

programs/token/src/instructions/initialize_mint_2.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// Freeze Authority.
+    pub freeze_authority: Option<Pubkey>,
```
febo, line 19: And here, reference to `Pubkey`.

programs/token/src/instructions/initialize_mint.rs
```diff
@@ -0,0 +1,80 @@
...
+///   1. `[]` Rent sysvar
+pub struct InitilizeMint<'a> {
+    /// Mint Account.
+    pub mint: &'a AccountInfo,
+    /// Rent sysvar Account.
+    pub rent_sysvar: &'a AccountInfo,
+    /// Decimals.
+    pub decimals: u8,
+    /// Mint Authority.
+    pub mint_authority: Pubkey,
```
febo, line 24: Same here, reference to `Pubkey`.

programs/token/src/instructions/initialize_mint.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// Freeze Authority.
+    pub freeze_authority: Option<Pubkey>,
```
febo, line 26: And here as well, reference to `Pubkey`.

programs/token/src/instructions/initialize_account_3.rs
```diff
@@ -0,0 +1,60 @@
...
+/// ### Accounts:
+///   0. `[WRITE]`  The account to initialize.
+///   1. `[]` The mint this account will be associated with.
+pub struct InitilizeAccount3<'a> {
+    /// New Account.
+    pub token: &'a AccountInfo,
+    /// Mint Account.
+    pub mint: &'a AccountInfo,
+    /// Owner of the new Account.
+    pub owner: Pubkey
```
febo, line 18: Same here, reference to `Pubkey`.

programs/token/src/instructions/initialize_account_2.rs
```diff
@@ -0,0 +1,64 @@
...
+///   3. `[]` Rent sysvar
+pub struct InitilizeAccount2<'a> {
+    /// New Account.
+    pub token: &'a AccountInfo,
+    /// Mint Account.
+    pub mint: &'a AccountInfo,
+    /// Rent Sysvar Account
+    pub rent_sysvar: &'a AccountInfo,
+    /// Owner of the new Account.
+    pub owner: Pubkey
```
febo, line 21: Same here, reference to `Pubkey`.

febo:
This is great! I left a few comments, but overall looks very good. The main one is about UB when you get a pointer from a `MaybeUninit`. We can only do that if the value is initialized. I think it would be fine to just replace the use of `MaybeUninit` for static allocated arrays – the compiler will probably optimize the initialization.

Also, could you please run `cargo fmt` and `clippy`? Otherwise CI will keep complaining. 😀

programs/token/src/state/mint.rs
deanmlittle: I'm really not sure what good counting borrows does us when the only way to modify this segment of memory is by CPI, and CPIs already don't care about us pretending using memory we don't own is safe.

febo: Yeah, I see your point. I am thinking on the case of the `CloseAccount` instruction for example. You could get the pointer to the token account and then do a CPI to close the account. After the CPI, that pointer is technically invalid, since the account has been resized to `0`.

The "checked" version of the `invoke_signed`, which is the one that the instruction builder is using, checks the borrow flags: https://github.com/febo/pinocchio/blob/main/sdk/pinocchio/src/program.rs#L91-L97

So if you have an "active" borrow, you will get a `ProgramError::AccountBorrowFailed` error before the CPI. I think we should offer a method that allows to bypass the borrow check and let the client code decide when that is safe, but I wonder if this behaviour should be behind a method name that makes that explicit.

L0STE:
Writing here a new comment addressing some of the changes:

- **State Getters UB**
Added 2 more tests [`token_getter`](https://github.com/L0STE/pinocchio-spl-examples/blob/a6b40b34640fdf1c7959d3858b8d3f28fce844d1/programs/pinocchio_spl_examples/src/tests/mod.rs#L2090) and [`mint_getter`](https://github.com/L0STE/pinocchio-spl-examples/blob/a6b40b34640fdf1c7959d3858b8d3f28fce844d1/programs/pinocchio_spl_examples/src/tests/mod.rs#L2017) to test out against UB for the state getters. 

Not sure if you were referring to something else but happy to change it if that's the case! 

- **Pubkey deferencing**
Good catch on the Pubkey dereferencing, didn't think about that. Changed all the function + tests CPI to suits that changes and seeing if any UB got introduced but seems everything is good on that front

- **UB from pointers from a MaybeUninit**
Instead of getting rid of the MaybeUninit and using the zeroed out array, we would like to propose this alternative solution: 
```
let mut instruction_data = MaybeUninit::<[u8; 10]>::uninit();

// Populate data
unsafe {
    let buf = instruction_data.assume_init_mut();
    
    // Set discriminator as u8 at offset [0]
    buf[0] = 12;
    // Set amount as u64 (convert to bytes and copy into buffer)
    buf[1..9].copy_from_slice(&self.amount.to_le_bytes());
    // Set decimals as u8 at offset [9]
    buf[9] = self.decimals;
}

    program_id: &crate::ID,
    accounts: &account_metas,
    data: unsafe { &instruction_data.assume_init() },
};
```
If this is satisfying I'll change the instructions and run additional tests

febo:
> Writing here a new comment addressing some of the changes:
> 
> * **State Getters UB**
>   Added 2 more tests [`token_getter`](https://github.com/L0STE/pinocchio-spl-examples/blob/a6b40b34640fdf1c7959d3858b8d3f28fce844d1/programs/pinocchio_spl_examples/src/tests/mod.rs#L2090) and [`mint_getter`](https://github.com/L0STE/pinocchio-spl-examples/blob/a6b40b34640fdf1c7959d3858b8d3f28fce844d1/programs/pinocchio_spl_examples/src/tests/mod.rs#L2017) to test out against UB for the state getters.
> 
> Not sure if you were referring to something else but happy to change it if that's the case!

I think what is tricky of UB is that it might work, but "break" in a different setting (arch or compiler version). From the docs, *when a raw pointer is dereferenced (using the `*` operator), it must be non-null and aligned*. I flagged a couple of values where the pointer won't be aligned to the type, which might work now but could break in the future.

A workaround would be to define the `u64` types as byte arrays, e.g.:
```rust
pub struct Mint {
    pub mint_authority_present: [u8; 4],

    pub mint_authority: Pubkey,

    pub supply: [u8; 8],

    pub decimals: u8,

    pub is_initialized: bool,

    pub freeze_authority_present: [u8; 4],

    pub freeze_authority: Pubkey,
}

impl Mint {
    pub fn supply(&self) -> u64 {
        u64::from_le_bytes(self.supply)
    }
}
```
You can then cast the raw pointer into this type to access its fields. The "penalty" from the `from_le_bytes` should be relatively small. The other alternative would be to use [`core::ptr::read_unaligned`](https://doc.rust-lang.org/std/ptr/fn.read_unaligned.html) &mdash; not sure if this one would perform better than the `from_le_bytes`.

> * **Pubkey deferencing**
>   Good catch on the Pubkey dereferencing, didn't think about that. Changed all the function + tests CPI to suits that changes and seeing if any UB got introduced but seems everything is good on that front

Great! 👍

> * **UB from pointers from a MaybeUninit**
>   Instead of getting rid of the MaybeUninit and using the zeroed out array, we would like to propose this alternative solution:
> 
> ```
> let mut instruction_data = MaybeUninit::<[u8; 10]>::uninit();
> 
> // Populate data
> unsafe {
>     let buf = instruction_data.assume_init_mut();
>     
>     // Set discriminator as u8 at offset [0]
>     buf[0] = 12;
>     // Set amount as u64 (convert to bytes and copy into buffer)
>     buf[1..9].copy_from_slice(&self.amount.to_le_bytes());
>     // Set decimals as u8 at offset [9]
>     buf[9] = self.decimals;
> }
> 
> let instruction = Instruction {
>     program_id: &crate::ID,
>     accounts: &account_metas,
>     data: unsafe { &instruction_data.assume_init() },
> };
> ```
> 
> If this is satisfying I'll change the instructions and run additional tests

I think this is still UB: https://doc.rust-lang.org/beta/std/mem/union.MaybeUninit.html#method.assume_init_mut

* From the docs: *Calling this* [`assume_init_mut`] *when the content is not yet fully initialized causes undefined behavior: it is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. For instance, .assume_init_mut() cannot be used to initialize a MaybeUninit.*

I think the compiler will do a good job in optimizing this even with a "normal" zeroed static array since the array gets populated immediately after.

deanmlittle:
> "I think the compiler will do a good job in optimizing this even with a "normal" zeroed static array since the array gets populated immediately after."

It will not. It will zero allocate this which costs CUs we really don't need to be burning when we're about to immediately overwrite that entire section of memory.

febo:
> > "I think the compiler will do a good job in optimizing this even with a "normal" zeroed static array since the array gets populated immediately after."
> 
> It will not. It will zero allocate this which costs CUs we really don't need to be burning when we're about to immediately overwrite that entire section of memory.

What if we use `write` instead? We could define a helper function/macro that uses `write` instead of using the `copy_from_slice` for byte arrays.
```rust
const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
let mut instruction_data = [UNINIT_BYTE; 10];

// Populate data
unsafe {
    // Set discriminator as u8 at offset [0]
    instruction_data[0].write(12);
    // Set amount as u64 from offset [1]
    // TODO: `write_bytes` needs to be defined
    write_bytes(&mut instruction_data[1..], &self.amount.to_le_bytes());
    // Set decimals as u8 at offset [9]
    instruction_data[9].write(self.decimals);
}

    program_id: &crate::ID,
    accounts: &account_metas,
    data: unsafe { &*instruction_data.as_ptr() },
};
```

programs/token/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#![no_std]
```
febo, line 1: 👍

febo:
@L0STE @deanmlittle I did a simple benchmark to see the difference between using a zeroed initialized array and variations of `MaybeUninit` that we have been discussing. I used the System `transfer` instruction, so `12` bytes of instruction data.

The zeroed initialized array used `+3` CU when compared to the versions using `MaybeUninit`. I tested 2 variations using `MaybeUninit`:

- the "original" one, which took a mutable pointer from the uninitialized array
- the last one that we discussed, which uses `write` to initialize the values before taking a pointer from the array

Both performed the same on this small test, no difference on CU consumption. Here is how the version with the `write` looks like:
```rust
const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
let mut instruction_data = [UNINIT_BYTE; 12];

write_bytes(&mut instruction_data, &[2, 0, 0, 0]);
write_bytes(&mut instruction_data[4..], &self.lamports.to_le_bytes());

    program_id: &crate::ID,
    accounts: &account_metas,
    data: unsafe { from_raw_parts(instruction_data.as_ptr() as _, 12) },
};
```
And the `write_bytes` helper:
```rust
fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
    for (d, s) in destination.iter_mut().zip(source.iter()) {
        d.write(*s);
    }
}
```

What do you think?

L0STE:
> @L0STE @deanmlittle I did a simple benchmark to see the difference between using a zeroed initialized array and variations of `MaybeUninit` that we have been discussing. I used the System `transfer` instruction, so `12` bytes of instruction data.

Thank you so much for doing this! It was on my list of todos for the weekend to see the impact of the changes but this helps a lot speeding up things. Will wrap up the PR this weekend so we can move forward! 

Thanks for providing your expertise and time!

L0STE:
**Update on the final PR:** 

**- General nit fixes:**
Deleted the tests folder, forgot to delete that when I moved the tests to the `spl-pinocchio-example` repo + ran cargo fmt and clip so now should be good to go on that point.

**- Alignment Issues**
Decided to opt for `unsafe { core::ptr::read_unaligned(self.0.add(64) as *const u64) }` as a solution since I tested it and didn't add any CUs.

**- `try_borrow_data()` for Mint and Token**
We opted to change the `from_account_info_unchecked` function to use `try_borrow_data()` as suggested but we added an unsafe function `from_account_info_unchecked_unsafe` that doesn't track the borrow cause using `try_borrow_data()` added 6CUs so it might be a good idea to just give the user the possibility of choosing.
```
/// # Safety
/// Use this when you don't care to keep track of the borrows &
/// the AccountInfo will have the correct owner and account space
#[inline(always)]
pub unsafe fn from_account_info_unchecked_unsafe(account_info: &AccountInfo) -> Self {
    Self(account_info.borrow_data_unchecked().as_ptr())
}

/// # Safety
/// Use this when you don't care to keep track of the borrows
pub unsafe fn from_account_info_unsafe(account_info: &AccountInfo) -> Self {
    assert_eq!(account_info.data_len(), Self::LEN);
    assert_eq!(account_info.owner(), &ID);
    Self::from_account_info_unchecked_unsafe(account_info)
}

/// Use this when you know the AccountInfo will have the correct owner and account space
#[inline(always)]
pub fn from_account_info_unchecked(account_info: &AccountInfo) -> Result<Self, ProgramError> {
    let data = account_info.try_borrow_data()?;
    Ok(Self(data.as_ptr()))
}

pub fn from_account_info(account_info: &AccountInfo) -> Result<Self, ProgramError> {
    }
    if account_info.owner() != &ID {
    }
    Self::from_account_info_unchecked(account_info)
}
 ```
 
 - `write_bytes()` helper
 I added the `write_bytes()` helper in the lib.rs so we could reference it for every instruction. Might make sense to add it in the pinocchio general sdk but didn't want to make the choice myself. So leaving this discussion here for you guys to decide on

**Comment by febo** [comment]
Amazing! 🙌

> **- Alignment Issues** Decided to opt for `unsafe { core::ptr::read_unaligned(self.0.add(64) as *const u64) }` as a solution since I tested it and didn't add any CUs.

Great!

> **- `try_borrow_data()` for Mint and Token** We opted to change the `from_account_info_unchecked` function to use `try_borrow_data()` as suggested but we added an unsafe function `from_account_info_unchecked_unsafe` that doesn't track the borrow cause using `try_borrow_data()` added 6CUs so it might be a good idea to just give the user the possibility of choosing.

I think we might have an issue here with the fact that you are taking a raw pointer and the borrow going out of scope. I will have a closer look at the code to confirm this.

> * `write_bytes()` helper
>   I added the `write_bytes()` helper in the lib.rs so we could reference it for every instruction. Might make sense to add it in the pinocchio general sdk but didn't want to make the choice myself. So leaving this discussion here for you guys to decide on

Read my mind. 😀 I think we can move it to an "utility" crate (e.g., `pinocchio-invoke`). There was also a suggestion of adding an `Invoke` trait to instruction `struct`s (I guess it would facilitate passing them as arguments or create CPI "bundles"), so this trait could also be part of this new crate.

I will review the PR shortly - thanks for addressing the comments!

File: programs/token/src/instructions/initialize_mint.rs [hunk 15]
```diff
@@ -0,0 +1,74 @@
...
+        // Set decimals as u8 at offset [1]
+        write_bytes(&mut instruction_data[1..2], &[self.decimals]);
+        // Set mint_authority as Pubkey at offset [2..34]
+        write_bytes(&mut instruction_data[2..34], self.mint_authority);
+        // Set COption & freeze_authority at offset [34..67]
+        if let Some(freeze_auth) = self.freeze_authority {
+            write_bytes(&mut instruction_data[34..35], &[1]);
+            write_bytes(&mut instruction_data[35..], freeze_auth);
+        } else {
+            write_bytes(&mut instruction_data[34..], &[0; 33]);
```
febo, line 63: I think you only need to write one `0` value to represent `None`, so `&[0]`.

programs/token/src/instructions/initialize_mint_2.rs
```diff
@@ -0,0 +1,68 @@
...
+        // Set decimals as u8 at offset [1]
+        write_bytes(&mut instruction_data[1..2], &[self.decimals]);
+        // Set mint_authority as Pubkey at offset [2..34]
+        write_bytes(&mut instruction_data[2..34], self.mint_authority);
+        // Set COption & freeze_authority at offset [34..67]
+        if let Some(freeze_auth) = self.freeze_authority {
+            write_bytes(&mut instruction_data[34..35], &[1]);
+            write_bytes(&mut instruction_data[35..], freeze_auth);
+        } else {
+            write_bytes(&mut instruction_data[34..], &[0; 33]);
```
febo, line 57: Same here, `&[0]`.

programs/token/src/instructions/set_authority.rs
```diff
@@ -0,0 +1,80 @@
...
+        // Set discriminator as u8 at offset [0]
+        write_bytes(&mut instruction_data, &[6]);
+        // Set authority_type as u8 at offset [1]
+        write_bytes(&mut instruction_data[1..2], &[self.authority_type as u8]);
+        // Set new_authority as [u8; 32] at offset [2..35]
+        if let Some(new_authority) = self.new_authority {
+            write_bytes(&mut instruction_data[2..3], &[1]);
+            write_bytes(&mut instruction_data[3..], new_authority);
+        } else {
+            write_bytes(&mut instruction_data[2..], &[0; 33]);
```
febo, line 69: Same here, `&[0]`.

programs/token/src/state/token.rs
```diff
@@ -0,0 +1,119 @@
...
+        assert_eq!(account_info.data_len(), Self::LEN);
+        assert_eq!(account_info.owner(), &ID);
+        Self::from_account_info_unchecked_unsafe(account_info)
+    }
+
+    /// Use this when you know the AccountInfo will have the correct owner and account space
+    #[inline(always)]
+    pub fn from_account_info_unchecked(account_info: &AccountInfo) -> Result<Self, ProgramError> {
+        let data = account_info.try_borrow_data()?;
+        Ok(Self(data.as_ptr()))
```
febo, line 31: Same here about the drop of the `Ref<[u8]>`.

programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,91 @@
...
+        assert_eq!(account_info.data_len(), Self::LEN);
+        assert_eq!(account_info.owner(), &ID);
+        Self::from_account_info_unchecked_unsafe(account_info)
+    }
+
+    /// Use this when you know the AccountInfo will have the correct owner and account space
+    #[inline(always)]
+    pub fn from_account_info_unchecked(account_info: &AccountInfo) -> Result<Self, ProgramError> {
+        let data = account_info.try_borrow_data()?;
+        Ok(Self(data.as_ptr()))
```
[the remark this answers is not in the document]
febo: Actually, maybe all we need is to `map` the `Ref<[u8]>` to a `Ref<Mint>`:
```rust
Ref::map(account_info.try_borrow_data()?, |data| Self(data.as_ptr()))
```
The return type of the method then becomes `Result<Ref<Mint>, ProgramError>`.

febo (commented):
This is super close, just a couple of things.

programs/token/src/state/mint.rs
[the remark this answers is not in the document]
Aursen: > Actually, maybe all we need is to `map` the `Ref<[u8]>` to a `Ref<Mint>`:
> 
> ```rust
> Ref::map(account_info.try_borrow_data()?, |data| Self(data.as_ptr()))
> ```
> 
> The return type of the method then becomes `Result<Ref<Mint>, ProgramError>`.

I think it's better too, more convenient for the user

febo:
@L0STE I had a second look at the `*_unchecked` and `*_unsafe` helpers – in particular to see if we can simplify the `*_unchecked_unsafe` naming.

Wonder if the following makes sense:

- `from_account_info(account_info: &AccountInfo)`: does the owner/length validation + returns a `Ref<T>`

- `from_account_info_unchecked(account_info: &AccountInfo)`: does the owner/length validation + returns a `T` (unchecked borrow)

- `from_bytes(bytes: &[u8])`: returns `T` (the caller controls whether to pass an unchecked borrow or not.

programs/token/src/instructions/initialize_mint.rs
L0STE: Good point actually, since the validation is only done on the first byte

programs/token/src/state/mint.rs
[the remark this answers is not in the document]
L0STE: Not sure if this is still correct, but I needed to add an unsafe cast because the compiler was yelling at me so I went with this for the code solution at the end:
```rust
pub fn from_account_info(account_info: &AccountInfo) -> Result<Ref<Mint>, ProgramError> {
    if account_info.data_len() != Self::LEN { return Err(ProgramError::InvalidAccountData) }
    if account_info.owner() != &ID { return Err(ProgramError::InvalidAccountData) }
    Ok(Ref::map(account_info.try_borrow_data()?, |data| {
        unsafe { &*(data.as_ptr() as *const Mint) }
    }))
}
 ```

**Comment by L0STE** [comment]
> @L0STE I had a second look at the `*_unchecked` and `*_unsafe` helpers – in particular to see if we can simplify the `*_unchecked_unsafe` naming.
> 
> Wonder if the following makes sense:
> 
> * `from_account_info(account_info: &AccountInfo)`: does the owner/length validation + returns a `Ref<T>`
> * `from_account_info_unchecked(account_info: &AccountInfo)`: does the owner/length validation + returns a `T` (unchecked borrow)
> * `from_bytes(bytes: &[u8])`: returns `T` (the caller controls whether to pass an unchecked borrow or not).

I think this totally make sense, this is my interpretation of this, let me know If I'm close: 
```rust
/// Performs owner and length validation on `AccountInfo` and returns a `Ref<T>` for safe borrowing.
pub fn from_account_info(account_info: &AccountInfo) -> Result<Ref<Mint>, ProgramError> {
    if account_info.data_len() != Self::LEN { return Err(ProgramError::InvalidAccountData) }
    if account_info.owner() != &ID { return Err(ProgramError::InvalidAccountData) }
    Ok(Ref::map(account_info.try_borrow_data()?, |data| {
        unsafe { &*(data.as_ptr() as *const Mint) }
    }))
}

/// # Safety
/// Performs owner and length validation on `AccountInfo` but performs unchecked borrowing and 
/// returns a `T` directly.
#[inline(always)]
pub unsafe fn from_account_info_unchecked(account_info: &AccountInfo) -> Result<Mint, ProgramError> {
    if account_info.data_len() != Self::LEN { return Err(ProgramError::InvalidAccountData) }
    if account_info.owner() != &ID { return Err(ProgramError::InvalidAccountData) }
    Ok(Self::from_bytes(account_info.borrow_data_unchecked().as_ref()))
}

/// # Safety
/// Constructs a `T` directly from a byte slice. The caller must ensure that `bytes` contains a 
/// valid representation of `T`.
pub unsafe fn from_bytes(bytes: &[u8]) -> Self {
    core::ptr::read(bytes.as_ptr() as *const Mint)
}
```

programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,84 @@
...
+        if account_info.data_len() != Self::LEN { return Err(ProgramError::InvalidAccountData) }
+        if account_info.owner() != &ID { return Err(ProgramError::InvalidAccountData) }
+        Ok(Self::from_bytes(account_info.borrow_data_unchecked().as_ref()))
+    }
+
+    /// # Safety
+    /// Constructs a `T` directly from a byte slice. The caller must ensure that `bytes` contains a 
+    /// valid representation of `T`.
+    pub unsafe fn from_bytes(bytes: &[u8]) -> Self {
+        core::ptr::read(bytes.as_ptr() as *const Mint)
```
febo, line 32: I think this should be `Self(bytes.as_ptr())`.

programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /// # Safety
+    /// Performs owner and length validation on `AccountInfo` but performs unchecked borrowing and 
+    /// returns a `T` directly.
+    #[inline(always)]
+    pub unsafe fn from_account_info_unchecked(account_info: &AccountInfo) -> Result<Mint, ProgramError> {
+        if account_info.data_len() != Self::LEN { return Err(ProgramError::InvalidAccountData) }
+        if account_info.owner() != &ID { return Err(ProgramError::InvalidAccountData) }
+        Ok(Self::from_bytes(account_info.borrow_data_unchecked().as_ref()))
```
febo, line 25: I think you can remove the `.as_ref()` call.

programs/token/src/state/token.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /// # Safety
+    /// Performs owner and length validation on `AccountInfo` but performs unchecked borrowing and 
+    /// returns a `T` directly.
+    #[inline(always)]
+    pub unsafe fn from_account_info_unchecked(account_info: &AccountInfo) -> Result<TokenAccount, ProgramError> {
+        if account_info.data_len() != Self::LEN { return Err(ProgramError::InvalidAccountData) }
+        if account_info.owner() != &ID { return Err(ProgramError::InvalidAccountData) }
+        Ok(Self::from_bytes(account_info.borrow_data_unchecked().as_ref()))
```
febo, line 27: Same here, you probably can remove the `.as_ref()` call.

febo (approved):
Looks great!

programs/token/src/instructions/approve.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        // Set discriminator as u8 at offset [0]
+        write_bytes(&mut instruction_data, &[4]);
+        // Set amount as u64 at offset [1..9]
+        write_bytes(&mut instruction_data[1..], &self.amount.to_le_bytes());
+
+        let instruction = Instruction {
+            program_id: &crate::ID,
+            accounts: &account_metas,
+            data: unsafe { from_raw_parts(instruction_data.as_ptr() as _, 9) },
```
alessandrod, line 56: why not just slice here?

programs/token/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+
+pinocchio_pubkey::declare_id!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
+
+use core::mem::MaybeUninit;
+
+const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
+
+#[inline(always)]
+fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
+    for (d, s) in destination.iter_mut().zip(source.iter()) {
```
alessandrod, line 14: did you check that this actually desugars to memcpy and not to a bunch of assignments?

programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Performs owner and length validation on `AccountInfo` and returns a `Ref<T>` for safe borrowing.
+    pub fn from_account_info(account_info: &AccountInfo) -> Result<Ref<Mint>, ProgramError> {
+        if account_info.data_len() != Self::LEN {
+            return Err(ProgramError::InvalidAccountData);
+        }
+        if account_info.owner() != &ID {
+            return Err(ProgramError::InvalidAccountData);
+        }
+        Ok(Ref::map(account_info.try_borrow_data()?, |data| unsafe {
+            &*(data.as_ptr() as *const Mint)
```
alessandrod, line 22: Is this guaranteed to have the right alignment? otherwise it's UB

programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,98 @@
...
+            Some(self.mint_authority_unchecked())
+        } else {
+            None
+        }
+    }
+
+    /// Use this when you know the account will have a mint authority and you want to skip the Option check.
+    #[inline(always)]
+    pub fn mint_authority_unchecked(&self) -> Pubkey {
+        unsafe { *(self.0.add(4) as *const Pubkey) }
```
alessandrod, line 65: Shouldn't this return by ref? Saves you a 32 byte copy

programs/token/src/state/mint.rs
febo: Yes, the data comes from the `AccountInfo` (pointer from the entrypoint input), which is aligned.

alessandrod: Uh I don't think that account pointers have a defined alignment, and the alignment isn't guaranteed to be mem::align_of::<Mint>() anyway?

febo: They should have, no?

https://solana.com/docs/programs/faq#input-parameter-serialization

Every account data starts from an 8-bytes aligned address. The `Mint` in this case is a struct that holds a pointer, so it should be aligned.

programs/token/src/state/token.rs
```diff
@@ -0,0 +1,128 @@
...
+    pubkey::Pubkey,
+};
+
+pub struct TokenAccount(*const u8);
+
+impl TokenAccount {
+    pub const LEN: usize = 165;
+
+    /// Performs owner and length validation on `AccountInfo` and returns a `Ref<T>` for safe borrowing.
+    pub fn from_account_info(
```
alessandrod, line 16: can't I do the following:

```
let token = TokenAccount::from_account_info(empty_account); // no unsafe
token.amount() // <- this is UB
token.authority() // <- this returns random garbage
if token.is_native() { // this is UB and could do anything
    ...
} else {
    ...
}
```

and the same with an account that doesn't respect the expected account data layout

alessandrod: nevermind I'm extremely dumb you _are_ checking the len I need to sleep

programs/token/src/state/mint.rs
alessandrod: Oh yes you're returning a ref to `Mint(ptr)` which is not the thing that is actually stored in the account, I missed that you had the pointer wrapper type

programs/token/src/state/mint.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// # Safety
+    /// Constructs a `T` directly from a byte slice. The caller must ensure that `bytes` contains a
+    /// valid representation of `T`.
+    pub unsafe fn from_bytes(bytes: &[u8]) -> Self {
+        Self(bytes.as_ptr())
+    }
+
+    #[inline(always)]
+    pub fn has_mint_authority(&self) -> bool {
+        unsafe { *(self.0 as *const bool) }
```
alessandrod, line 51: hopefully whatever is storing this isn't storing bool and is storing u8, and you should also load as u8 and explicitly check for != 0

programs/token/src/instructions/approve.rs
febo: It is an array of `[MaybeUninit<u8>]` so we need to cast it to `[u8]`.

programs/token/src/lib.rs
febo: I think we probably can replace this by a `memcpy`. It is using an iteration to call `write` since each position is a `MaybeUninit<u8>` but since we know the size in advance, we might be able to use `ptr::copy_nonoverlapping` instead.

PR #36, opened 2024-11-11
feat : Add token program 2022 

Nagaprasadvr:
- [ ] Add token program 2022 ix (extensions included)


Jac0xb:
I think there should be de-duping with the other sdk with instruction/CPI calling as much as possible to avoid bytecode and logic bloat.

Perelyn-sama:
Hi, is there anything holding this back? And if yes, what can I do to help?

Nagaprasadvr:
> Hi, is there anything holding this back? And if yes, what can I do to help?

Working on adding all extensions

programs/token/src/extensions/metadata.rs
```diff
@@ -0,0 +1,53 @@
...
+        if !account_info.is_owned_by(&TOKEN_2022_PROGRAM_ID) {
+            return Err(ProgramError::InvalidAccountOwner);
+        }
+
+        let acc_data_bytes = account_info.try_borrow_data()?;
+        let acc_data_bytes = acc_data_bytes.as_ref();
+
+        get_extension_from_bytes::<Self>(acc_data_bytes).ok_or(ProgramError::InvalidAccountData)
+    }
+}
```
Taylor123, line 47: is this missing the discriminator within LEN and deserialization? Should be `[112, 132, 90, 90, 11, 88, 157, 87]` for TokenMetadata.

Taylor123:
I've got a need for Metadata, Groups, and Members. Doesn't look like the latter 2 are supported. Will start on adding Group extensions if no one else is working on it.

Nagaprasadvr:
> I've got a need for Metadata, Groups, and Members. Doesn't look like the latter 2 are supported. Will start on adding Group extensions if no one else is working on it.

Please Go ahead, also I need to fix TokenMetadata it wasn't deserializing

programs/token/src/extensions/metadata.rs
Nagaprasadvr: Does it have 8 bytes acc discriminator ? need to check

L0STE:
Hey @Nagaprasadvr, I'm going to turn the whole thing into `no-std` because most of the program we write with pinocchio uses that.

Additionally, let me know what is missing to push this over the finish line since I need to use it for a project I'm working on and can push this to the finish line pretty fast

Nagaprasadvr:
> Hey @Nagaprasadvr, I'm going to turn the whole thing into `no-std` because most of the program we write with pinocchio uses that.
> 
> Additionally, let me know what is missing to push this over the finish line since I need to use it for a project I'm working on and can push this to the finish line pretty fast

hey , its still using no_std but TokenMetadata is using let me see if we can use no_std for that , additionally we are trying to cover all extensions , i guess only few are left i will let u know the list

L0STE:
> let me see if we can use no_std for that 

Don't worry, I already did the `no_std` implementation! Wrapping up and testing it but should be ready by eotd

Nagaprasadvr:
> > let me see if we can use no_std for that
> 
> Don't worry, I already did the `no_std` implementation! Wrapping up and testing it but should be ready by eotd

Oh cool thats great

L0STE:
Opened a PR for the `no-std` version. We probably will need an helper for deserializing `additional_metadata` if it's important for someone

Nagaprasadvr:
> hi @Nagaprasadvr 👋 what still isn't implemented? i can take 1-2 tasks to speed-up the process slightly

Hey u can check TokenMetadat ixs , it's not done , also transfer hook extension ty

Nagaprasadvr:
> > hi @Nagaprasadvr 👋 what still isn't implemented? i can take 1-2 tasks to speed-up the process slightly
> 
> Hey u can check TokenMetadat ixs , it's not done , also transfer hook extension ty

Hey @sonicfromnewyoke TokenMetadata ixs work is already starter can u please not work on that instead u could take Transfer hook

sonicfromnewyoke:
> > > hi @Nagaprasadvr 👋 what still isn't implemented? i can take 1-2 tasks to speed-up the process slightly
> > 
> > 
> > Hey u can check TokenMetadat ixs , it's not done , also transfer hook extension ty
> 
> Hey @sonicfromnewyoke TokenMetadata ixs work is already started can u please not work on that instead u could take Transfer hook

roger 🫡

programs/token/src/state/token.rs
```diff
@@ -86,9 +90,13 @@ impl TokenAccount {
         if account_info.data_len() != Self::LEN {
```
sonicfromnewyoke, line 90: `LEN` of the token account fluctuates depending on extensions added. This check always validates account size against the `165 bytes`.

For example, this account `6k7LYSfeaajmLpC4gbz2pn69YtBoXwZPfgb13E3NRSkt`  has `182 bytes` size and will fail this check

programs/token/src/state/mint.rs
```diff
@@ -47,7 +47,10 @@ impl Mint {
         if account_info.data_len() != Self::LEN {
```
sonicfromnewyoke, line 47: `LEN` of the mint fluctuates depending on extensions added. This check always validates mint size against the `82 bytes`.

For example, this mint `Ddm4DTxNZxABUYm2A87TFLY6GDG2ktM2eJhGZS3EbzHM`  has `530 bytes` size and will fail this check

programs/token/src/lib.rs
```diff
@@ -1,14 +1,39 @@
...
 
+pub mod extensions;
 pub mod instructions;
 pub mod state;
 
-pinocchio_pubkey::declare_id!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
+use pinocchio::pubkey::Pubkey;
+use pinocchio_pubkey::pubkey;
+
+pub const LEGACY_TOKEN_PROGRAM_ID: Pubkey = pubkey!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
```
sonicfromnewyoke, line 10: wdyt about using `TOKEN_PROGRAM_ID` as a name, without `LEGACY` prefix?

programs/token/src/state/mint.rs
Nagaprasadvr: > `LEN` of the mint fluctuates depending on extensions added. This check always validates mint size against the `82 bytes`.
> 
> For example, this mint `Ddm4DTxNZxABUYm2A87TFLY6GDG2ktM2eJhGZS3EbzHM` has `530 bytes` size and will fail this check

makes sense we should just check for minimum len

kevinheavey:
I strongly encourage not making these the same crate, as spl-token-2022 has a way higher API surface area, so it will have far more major version bumps than spl-token

programs/token/src/state/mint.rs
```diff
@@ -68,13 +71,19 @@ impl Mint {
     pub unsafe fn from_account_info_unchecked(
         account_info: &AccountInfo,
     ) -> Result<&Self, ProgramError> {
-        if account_info.data_len() != Self::LEN {
+        if account_info.data_len() < Self::LEN {
             return Err(ProgramError::InvalidAccountData);
         }
-        if account_info.owner() != &ID {
+
+        if !account_info.is_owned_by(&TOKEN_2022_PROGRAM_ID)
```
febo, line 78: It will be more efficient for this to be specific to one token program, since a pubkey comparison consumes `12` CUs. This is another motivation to have separate crates.

programs/token/src/instructions/approve.rs
```diff
@@ -28,11 +30,15 @@ pub struct Approve<'a> {
 
 impl Approve<'_> {
     #[inline(always)]
-    pub fn invoke(&self) -> ProgramResult {
-        self.invoke_signed(&[])
+    pub fn invoke(&self, token_program: TokenProgramVariant) -> ProgramResult {
```
febo, line 33: Did you consider adding the `token_program` as an account to the struct? That way you don't need to statically "choose" the program, it will depend on the token program account.

febo:
> I strongly encourage not making these the same crate, as spl-token-2022 has a way higher API surface area, so it will have far more major version bumps than spl-token

Agree with this – it will be better to have this as a separate crate. The other question is whether it should support both token programs or not. I would avoid having pubkey comparisons hard-coded on it – e.g.,:
```rust
if !account_info.is_owned_by(&TOKEN_2022_PROGRAM_ID) && !account_info.is_owned_by(&TOKEN_PROGRAM_ID)
```
This will consume extra `12` CUs everytime. We could have a generic parameter that specify the "target" token program or require a parameter with the token program pubkey to match.

Nagaprasadvr:
> > I strongly encourage not making these the same crate, as spl-token-2022 has a way higher API surface area, so it will have far more major version bumps than spl-token
> 
> Agree with this – it will be better to have this as a separate crate. The other question is whether it should support both token programs or not. I would avoid having pubkey comparisons hard-coded on it – e.g.,:
> 
> ```rust
> if !account_info.is_owned_by(&TOKEN_2022_PROGRAM_ID)
>     && !account_info.is_owned_by(&TOKEN_PROGRAM_ID) {
> ```
> 
> This will consume extra `12` CUs every time. We could have a generic parameter that specify the "target" token program or require a parameter with the token program pubkey to match, so that there is only one comparison needed.

so u mean having all the shared ixs (like approve , transfer .. ) in token crate supporting both token and token2022 program and then just exposing extensions in a separate crate ?

Nagaprasadvr:
> > > I strongly encourage not making these the same crate, as spl-token-2022 has a way higher API surface area, so it will have far more major version bumps than spl-token
> > 
> > 
> > Agree with this – it will be better to have this as a separate crate. The other question is whether it should support both token programs or not. I would avoid having pubkey comparisons hard-coded on it – e.g.,:
> > ```rust
> > if !account_info.is_owned_by(&TOKEN_2022_PROGRAM_ID)
> >     && !account_info.is_owned_by(&TOKEN_PROGRAM_ID) {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will consume extra `12` CUs every time. We could have a generic parameter that specify the "target" token program or require a parameter with the token program pubkey to match, so that there is only one comparison needed.
> 
> so u mean having all the shared ixs (like approve , transfer .. ) in token crate supporting both token and token2022 program and then just exposing extensions in a token-2022 crate?

as u mentioned we can pass token_program pubkey and use it for matching for all cpi apis

Nagaprasadvr:
> Has there been any other movement on this? Would be great to get it in. Happy to help push things over the finish line.

Just thinking about removing shared ixs between token and token-2022 by passing token program as an account to invoke ixs 

By doing this we will only have extensions inside token-2022

febo:
Sorry, the delay is on me. I am currently refactoring a bit the CPI functions (PR #154) since we can use less CUs on these clients. I am close to finish it and then I will have a look at this PR.

Nagaprasadvr:
> @Nagaprasadvr Is this suppose to cover all Token 2022 instructions? E.g., I can't see `withdraw_excess_lamports`.

Yeah right ig i might be missing something, i only looked at extensions will check other ixs supported and add them

lainhathoang:
Is it ready to use? I am eagerly waiting for the Token-2022 Pinocchio program

Nagaprasadvr:
> Is it ready to use? I am eagerly waiting for the Token-2022 Pinocchio program

Well this is just an api to token-2022 , we haven't developed Pinocchio version of token-2022 but soon

febo:
@Nagaprasadvr Thanks for the PR, this looks very good. To help speed up the review, could we please break it down into separate PRs so we can get things merged iteratively?

My initial idea is to have:
1. A PR with the "code" program instructions + state – these in general are all the same. Ideally this should not be a "copy-and-paste" from the token one, we should try to refactor in a way that we reuse the same ones. We can add the program ID as a `const` parameter or pass the `AccountInfo` of the program to use.
2. A separate PR for each of the extensions. This way we can review them individually and they can go in independently.

Happy to help you do that if needed.

Nagaprasadvr:
> @Nagaprasadvr Thanks for the PR, this looks very good. To help speed up the review, could we please break it down into separate PRs so we can get things merged iteratively?
> 
> My initial idea is to have:
> 
> 1. A PR with the "code" program instructions + state – these in general are all the same. Ideally this should not be a "copy-and-paste" from the token one, we should try to refactor in a way that we reuse the same ones. We can add the program ID as a `const` parameter or pass the `AccountInfo` of the program to use.
> 2. A separate PR for each of the extensions. This way we can review them individually and they can go in independently.
> 
> Happy to help you do that if needed.

Sounds good ser , i will start working on this soon , ig we could send a const param for common ixs

Taylor123:
@Nagaprasadvr How's progress on the requested changes? Anything we can divide and conquer amongst the community?

Nagaprasadvr:
> @Nagaprasadvr How's progress on the requested changes? Anything we can divide and conquer amongst the community?

Discussing with febo on how to combine core ixs for both legacy and token2022

programs/token-2022/src/extensions/transfer_fee.rs
```diff
@@ -0,0 +1,467 @@
...
+        };
+
+        invoke_signed(&instruction, &acc_infos, signers)
+    }
+}
+
+/// Harvest withheld tokens to mint accounts.
+pub struct HarvestWithheldTokensToMint<'a, const ACCOUNTS_LEN: usize> {
+    /// Mint account (must include the `TransferFeeConfig` extension)
+    mint: &'a AccountInfo,
```
kz2d, line 348: Hi. As i understand this should be `pub`

Nagaprasadvr:
closing this , planning to get everyone who contributed (if they are willing) to this and make sep prs for each extension

PR #37, opened 2024-11-12
add IxData struct to handle write and read bytes

Nagaprasadvr:
- [x] Add a struct to handle writing and reading bytes 
- [x] add methods to handle optional pubkey writing to ix buffer
- [x] this will help to avoid few cus if we have multiple optional fileds in ix data and we have to check each parent field if its optional and then write to buffer accordingly

### DIFF
--- a/programs/token/src/instructions/approve.rs
+++ b/programs/token/src/instructions/approve.rs
@@ -47,7 +47,6 @@ impl<'a> Approve<'a> {
         // Set discriminator as u8 at offset [0]
         ix_data.write_bytes(&[4]);
         // Set amount as u64 at offset [1..9]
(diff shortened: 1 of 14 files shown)


febo:
@Nagaprasadvr Thanks for your contribution. While the idea of a helper struct is very nice, unfortunately it will use more compute units than the "manual" way due to the capacity checks and incrementing the size counter. At the same time the main benefit it provides is that we don't need to manually keep track of the size, but it is probably something that we will be doing anyway because we need to determine the size to allocated the array.

The overall idea of the instruction builders is to be as lightweight as possible, since they will be used in your program on-chain.

Nagaprasadvr:
Makes sense thanks for the review closing thks !

PR #42, opened 2024-11-17
A close macro implementation for AccountInfo

L0STE:
PRd this macro implementation invented by @deanmlittle (known as `based_close`) in the Pinocchio AccountInfo implementation to use it more easily.

I added 2 different function + docs:
- `.close()`: the implementation with pointer of the underlying idea
- `.optimized_close()`: the implementation with assembly of the underlying idea (marked the implementation as unsafe function since it uses `asm_experimental_arch`

Some data about the CU usage: 
- `.close()`: takes 60 CUs
- `.optimized_close()`: takes 7 CUs

**Edit**: now close has 7CUs as well (thanks to @febo's comment) and `optimized_close()` is now called `assembly_close()`

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -398,11 +398,21 @@ impl AccountInfo {
     /// referenced by `AccountInfo` fields. It should only be called for
     /// instances of `AccountInfo` that were created by the runtime and received
     /// in the `process_instruction` entrypoint of a program.
+    ///
+    /// This method is unsafe because it does not check if the account data is already
+    /// borrowed. It should only be called when the account is not being used.
     #[inline(always)]
-    pub fn close(&self) {
-        unsafe {
-            *(self.borrow_mut_data_unchecked().as_ptr().sub(48) as *mut [u8; 48]) = [0u8; 48];
-        }
+    pub unsafe fn close(&self) {
+        // Zero out the account owner. While the field is a `Pubkey`, it is quicker
+        // to zero the 32 bytes as 4 x `u64`s.
+        *(self.data_ptr().sub(48) as *mut u64) = 0u64;
+        *(self.data_ptr().sub(40) as *mut u64) = 0u64;
+        *(self.data_ptr().sub(32) as *mut u64) = 0u64;
+        *(self.data_ptr().sub(16) as *mut u64) = 0u64;
+        // Zero the account lamports.
+        (*self.raw).lamports = 0;
+        // Zero the account data length.
+        (*self.raw).data_len = 0;
     }
 
     /// Zero out the the account's data_len, lamports and owner fields, effectively
@@ -429,13 +439,17 @@ impl AccountInfo {
     /// This method set the variable as 0 making sure that we're actually zeroing out
     /// the bytes.
     #[inline(always)]
-    pub unsafe fn optimized_close(&self) {
+    pub unsafe fn assembly_close(&self) {
         #[cfg(target_os = "solana")]
         unsafe {
             let var = 0u64;
             core::arch::asm!(
+                // Zero the data lenght.
                 "stxdw [{0}-8], {1}",
+                // Zero the account lamports.
                 "stxdw [{0}-16], {1}",
+                // Zero out the account owner. While the field is a `Pubkey`, it is quicker
+                // to zero the 32 bytes as 4 x `u64`s.
                 "stxdw [{0}-24], {1}",
                 "stxdw [{0}-32], {1}",
                 "stxdw [{0}-40], {1}",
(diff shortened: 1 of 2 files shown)


sdk/pinocchio/src/account_info.rs
```diff
@@ -379,6 +378,75 @@ impl AccountInfo {
...
+    /// This method makes assumptions about the layout and location of memory
+    /// referenced by `AccountInfo` fields. It should only be called for
+    /// instances of `AccountInfo` that were created by the runtime and received
+    /// in the `process_instruction` entrypoint of a program.
+    #[inline(always)]
+    pub fn close(&self) {
+        unsafe {
+            *(self.borrow_mut_data_unchecked().as_ptr().sub(48) as *mut [u8; 48]) = [0u8; 48];
+        }
+    }
```
febo, line 406:
```rust
 pub unsafe fn close(&self) {
        // Zero out the account owner. While the field is a `Pubkey`, it is quicker
        // to zero the 32 bytes as 4 x u64s than to zero the 32 bytes.
        *(self.data_ptr().sub(48) as *mut u64) = 0u64;
        *(self.data_ptr().sub(40) as *mut u64) = 0u64;
        *(self.data_ptr().sub(32) as *mut u64) = 0u64;
        *(self.data_ptr().sub(16) as *mut u64) = 0u64;
        // Zero the account lamports.
        (*self.raw).lamports = 0;
        // Zero the account data length.
        (*self.raw).data_len = 0;
    }
```

sdk/pinocchio/src/account_info.rs
```diff
@@ -379,6 +378,75 @@ impl AccountInfo {
         Ok(())
     }
 
+    /// Zero out the the account's data_len, lamports and owner fields, effectively
```
febo, line 381: This method probably needs to be marked `unsafe` since it does not check whether the account is already borrowed or not. We should add a comment along the lines:
```rust
/// # Safety
///
/// This method is unsafe because it does not check if the account data is already
/// borrowed. It should only be called when the account is not being used.
```

sdk/pinocchio/src/account_info.rs
febo: With this version, I measured `7` CUs.

sdk/pinocchio/src/account_info.rs
```diff
@@ -379,6 +378,89 @@ impl AccountInfo {
...
+    ///
+    /// This method makes assumptions about the layout and location of memory
+    /// referenced by `AccountInfo` fields. It should only be called for
+    /// instances of `AccountInfo` that were created by the runtime and received
+    /// in the `process_instruction` entrypoint of a program.
+    ///
+    /// This method is unsafe because it does not check if the account data is already
+    /// borrowed. It should only be called when the account is not being used.
+    #[inline(always)]
+    pub unsafe fn close(&self) {
```
febo, line 405: Should we rename this one to `close_unchecked` and add a "safe" version for `close`?

```rust
#[inline(always)]
pub fn close(&self) -> ProgramResult {
    {
        let _ = self.try_borrow_mut_data()?;
    }

    unsafe { self.close_unchecked(); }

    Ok(())
}

File: sdk/pinocchio/src/account_info.rs [hunk 4]
```diff
@@ -379,6 +378,89 @@ impl AccountInfo {
         Ok(())
     }
 
+    /// Zero out the the account's data_len, lamports and owner fields, effectively
+    /// closing the account.
+    ///
+    /// Note: This doesn't protect against future reinitialization of the account
```
febo, line 384: I think we can remove the `"Note:"` from the start of this sentence.

sdk/pinocchio/src/account_info.rs
```diff
@@ -379,6 +378,89 @@ impl AccountInfo {
...
+    /// since the account_data will need to be zeroed out as well. Or if the attacker
+    /// has access to the keypair of the account that we're trying to close, they can
+    /// just add the lenght, lamports and owner back before the data is wiped out from
+    /// the ledger.
+    ///
+    /// Note: This works because the 48 bytes before the account data are:
```
febo, line 390: Same here, I think we can remove the `"Note:"` from the start of this sentence.

sdk/pinocchio/src/account_info.rs
```diff
@@ -379,6 +378,89 @@ impl AccountInfo {
...
+    ///
+    /// This method makes assumptions about the layout and location of memory
+    /// referenced by `AccountInfo` fields. It should only be called for
+    /// instances of `AccountInfo` that were created by the runtime and received
+    /// in the `process_instruction` entrypoint of a program.
+    ///
+    /// This method set the variable as 0 making sure that we're actually zeroing out
+    /// the bytes.
+    #[inline(always)]
+    pub unsafe fn assembly_close(&self) {
```
febo, line 442: Given that they behave the same, do you think we need to have 2 "copies" of `close`? It might be clearer to keep just one of them, so there is no ambiguity in which one to use.

febo (commented):
Looks great – left a couple of suggestions.

febo (approved):

PR #51, opened 2024-12-12
log: Add precision support to `str`

febo:
### Problem

Precision formatting was added to `Logger` but it is currently ignored by the `str` implementation.

### Solution

Add support for precision when logging `str` values. This was done as separate `Argument` variants to encode the truncate position: `TruncateEnd` and `TruncateStart`. There are 3 cases for truncation:

1. The log buffer is large enough to hold the entire `str`: the whole `str` is logged.
2. The log buffer is smaller than the `str`, but large enough to hold a "`...`" slice and part of the `str`: the slice and part of the `str` are logged.
3. The log buffer is smaller than both the `str` and the "`...`" slice: the buffer is filled with `.` characters and the last character is set to `@`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,14 +38,14 @@ version = "0.6.0"
 
 [[package]]
 name = "pinocchio-log"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "pinocchio-log-macro",
 ]
 
 [[package]]
 name = "pinocchio-log-macro"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "quote",
  "regex",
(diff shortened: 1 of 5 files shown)


LucasSte:
When handling strings, the argument for precision shouldn't be a `u8`. We may want to write more than 255 bytes.

febo:
> When handling strings, the argument for precision shouldn't be a `u8`. We may want to write more than 255 bytes.

Good point! So we should introduce a new `Argument` variant for this and then we can also add an option to specify the truncation location (start or end).

febo:
@LucasSte Added `Argument` variants specifically for truncating and now the value is specified as a `usize`.

LucasSte (approved):

PR #52, opened 2024-12-15
macro: Add truncate formatting

febo:
### Problem

There are two new `Attribute` variants (`TruncateEnd` and `TruncateStart`) that allows specifying a maximum length for `&str` log messages. These are not currently supported by the `log!` macro.

### Solution

Add support to both `TruncateEnd` and `TruncateStart` by introducing two new formatting strings:
- "{:>.*size*}" for `TruncateEnd`
- "{:<.*size*}" for `TruncateStart`

### Examples

```rust
use pinocchio_log::log;

let program_name = "pinocchio-program";

// log output: "...program"
log!(":<.10", program_name);

// log output: "pinocchio-..."
log!(":>.10", program_name);
```

### DIFF
--- a/sdk/log/README.md
+++ b/sdk/log/README.md
@@ -108,9 +108,31 @@ logger.append_with_args(amount, &[Argument::Precision(9)]);
 logger.log()
 ```
 
-## Limitations
+And a maximum length for `&str` types with a truncate strategy:
+```rust
+use pinocchio_log::logger::{Attribute, Logger};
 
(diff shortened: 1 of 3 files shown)


febo:
> This is cool! I definitely like the `.width` specifier, since that lines up with how numbers work. As for the start / end specifier, reusing `<` and `>` is clever, but I'm worried that it could be confusing since the behavior isn't the same as the normal `<` / `>` for alignment in `format!`.
> 
> Unfortunately, the only other characters I can see are `[` and `]`, but I'm not sure that's much better. If you really like that option, then we can do that, otherwise, this looks great!

We tried different options, but in the end `<` and `>` seems to be the better one.

PR #56, opened 2025-01-23
Add close unstable

febo:
### Problem

When PR #55 fixed the last index of the owner `Pubkey` on `close_unchecked`, it pushed the CUs used by the method to over `50`.

### Solution

Reimplement `close_unchecked` using the `sol_memset` syscall, which consumes `14` CUs instead. Since this is still over the original `7` CUs, this PR also re-adds the `close_unstable` which uses ASM – this variant takes `7` CUs.

The `close_unstable` can be enabled by using the feature `"asm"`.

Note: `close_unstable` courtesy of @L0STE and @deanmlittle

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -444,7 +444,7 @@ impl AccountInfo {
     /// closing the account.
     ///
     /// This doesn't protect against future reinitialization of the account
-    /// since the account data will need to be zeroed out as well; otherwise the lenght,
+    /// since the account data will need to be zeroed out as well; otherwise the length,
     /// lamports and owner can be set again before the data is wiped out from
     /// the ledger using the keypair of the account being close.
     ///
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -157,6 +157,7 @@ impl AccountInfo {
 
     /// Changes the owner of the account.
     #[allow(invalid_reference_casting)]
+    #[inline]
     pub fn assign(&self, new_owner: &Pubkey) {
         unsafe {
             core::ptr::write_volatile(&(*self.raw).owner as *const _ as *mut Pubkey, *new_owner);
@@ -169,6 +170,7 @@ impl AccountInfo {
     ///
     /// This method is unsafe because it does not return a `Ref`, thus leaving the borrow
     /// flag untouched. Useful when an instruction has verified non-duplicate accounts.
+    #[inline]
     pub unsafe fn borrow_lamports_unchecked(&self) -> &u64 {
         &(*self.raw).lamports
     }
@@ -180,6 +182,7 @@ impl AccountInfo {
     /// This method is unsafe because it does not return a `Ref`, thus leaving the borrow
     /// flag untouched. Useful when an instruction has verified non-duplicate accounts.
     #[allow(clippy::mut_from_ref)]
+    #[inline]
     pub unsafe fn borrow_mut_lamports_unchecked(&self) -> &mut u64 {
         &mut (*self.raw).lamports
     }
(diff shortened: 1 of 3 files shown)


sdk/pinocchio/src/account_info.rs
```diff
@@ -430,17 +436,45 @@ impl AccountInfo {
         // - 8 bytes for the data_len
         //
         // So we can zero out them directly.
+        #[cfg(target_os = "solana")]
+        sol_memset_(self.data_ptr().sub(48), 0, 48);
```
alessandrod, line 440: you can use slice::fill and can get rid of the cfg attr

sdk/pinocchio/src/account_info.rs
```diff
@@ -430,17 +436,45 @@ impl AccountInfo {
...
+    /// instances of `AccountInfo` that were created by the runtime and received
+    /// in the `process_instruction` entrypoint of a program.
+    #[cfg(feature = "asm")]
+    #[inline(always)]
+    pub unsafe fn close_unstable(&self) {
+        #[cfg(target_os = "solana")]
+        unsafe {
+            let zero = 0u64;
+            core::arch::asm!(
+                "stxdw [{0}-8], {1}",
```
LucasSte, line 467: You can try `stdw [{0}-8], 0` directly here, without an extra register.

sdk/pinocchio/src/account_info.rs
febo: Using `slice::fill` takes `58` CUs – same as the current code. The `sol_memset_` is more efficient in this case.

sdk/pinocchio/src/account_info.rs
febo: As per offline conversation, the `stdw` is not yet available to the compiler – only the vm.

sdk/pinocchio/src/account_info.rs
alessandrod: yes, but why?

If the compiler is desugaring to stores, CUs should be lower than the syscall - it's the whole point of applying the optimization. It might mean that the optimization is wrong, or the syscall is mispriced.

febo: I had the same question. 😊 @LucasSte looked into it and already has a plan to improve it.

LucasSte: Four stores is the threshold to call `memset` from compiler-builtins. This is what causing the CU bump. Both the threshold to call the library in LLVM and the threshold to call the syscall in compiler-builints needs to be revised.

sdk/pinocchio/src/account_info.rs
deanmlittle: Yeah, this has been disabled since like forever. You literally can't deploy a program that contains any of the `st` class insns. If your code never updates the value of r0, you can accomplish the same thing with stxdw[offset], r0 but it's not safe to do so in Rust without some kind of a compiler hint to tell it "please don't reassign r0".

LucasSte: > You literally can't deploy a program that contains any of the st class insns. 

Where did you get this information from? Not only do the class [pass the bytecode verification](https://github.com/anza-xyz/sbpf/blob/946f7d98b1df466af547878ba8f276e7f2642229/src/verifier.rs#L278-L281), but also they are implemented in the [jitter](https://github.com/anza-xyz/sbpf/blob/946f7d98b1df466af547878ba8f276e7f2642229/src/jit.rs#L471-L482) and the [interpreter](https://github.com/anza-xyz/sbpf/blob/946f7d98b1df466af547878ba8f276e7f2642229/src/interpreter.rs#L221-L236).

The fact that the compiler wasn't emitting code for the store immediate class doesn't mean you can't deploy a program with it.

sealdevnut (commented):
confirm

febo:
This is not needed anymore.

Issue #39, opened 2024-11-17
Token program instruction builders dont support token2022


Jac0xb:
### Issue

- Token programs have hard coded legacy token program in the instruction builders.

sonicfromnewyoke:
I think it is ok, and there will be another lib `token-2022` with almost a copy-paste code from the legacy token program with hardcoded `TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb`.

The same approach was used in fndn

Jac0xb:
I do not hold the original libs in high regards. This seems like a poor use of space if I am including these in my onchain program. Almost all the instructions are identical and there could easily be a field in the struct for token program rather than needed to depend on two packages and have conditional logic that duplicates the code for each version.

sonicfromnewyoke:
> I do not hold the original libs in high regards

makes sense

Nagaprasadvr:
Makes sense but this would increase 32 bytes for each common ix wyt @febo

febo:
Since `token-2022` is a superset of the token instructions, perhaps the `token-2022` client should include the token program as a parameter. This way it can be used for both `token` and `token-2022`.

Nagaprasadvr:
so we just move all token ixs to token-2022 client  and add an enum field to mark wither legacy or token-2022 ig

febo:
> so we just move all token ixs to token-2022 client and add an enum field to mark whether legacy or token-2022 ig

Not sure if we need an `enum` field. We could have a field for the token program's `AccountInfo` and use the `key` from there.

Taylor123:
Has any more thought been put into Token2022 support? Would be great to have the token crate support both.

Nagaprasadvr:
> Has any more thought been put into Token2022 support? Would be great to have the token crate support both.

I have a pr tat i was working on https://github.com/anza-xyz/pinocchio/pull/36

alizeeshan1234:
[pinocchioPreReq1.docx](https://github.com/user-attachments/files/21009023/pinocchioPreReq1.docx)

febo:
Support was added in PR #190

PR #58, opened 2025-01-24
feat: add associated token account program

sonicfromnewyoke:
should conflict with #39


Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "programs/system",
     "programs/token",
+    "programs/associated-token",
```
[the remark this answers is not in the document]
febo: Yeah, let's use the same name as the program, so `associated-token-account`.

programs/associated-token/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "pinocchio-associated-token"
```
febo, line 2: Same here, `pinocchio-associated-token-account`.

programs/associated-token/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "Pinocchio helpers to invoke Associated Token program instructions"
```
febo, line 3: nit: Associated Token Account

febo (commented):
Looks good, just small nits.

programs/associated-token/Cargo.toml
sonicfromnewyoke: done

Cargo.toml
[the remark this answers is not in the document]
sonicfromnewyoke: done

programs/associated-token-account/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "pinocchio-associated-token-account"
+description = "Pinocchio helpers to invoke Associated Token Account program instructions"
+version = "0.2.1"
```
febo, line 4: Should we have the version starting from `0.1.0`?

febo (commented):
Added a banner image, which is the one that we can add to the other packages as well. I think we only need to update the version of the crate and it should be good to merge.

programs/associated-token-account/Cargo.toml
sonicfromnewyoke: yeah, makes sense.

done

febo (approved):
👌

PR #62, opened 2025-02-07
ci: Test filter workspace

febo:
This is just a test of the CI.


febo:
The conditional filter was removed since it was not precise – e.g., it did not take into account the dependencies between crates.

PR #64, opened 2025-02-08
sdk: Improve comments

febo:
### Problem

Currently the doc linting fails on the main pinocchio crate (`sdk/pinocchio`).

### Solution

Fix the linting warnings. It also expand the crate comments.


sdk/pinocchio/src/sysvars/fees.rs
```diff
@@ -38,7 +38,10 @@ pub struct FeeRateGovernor {
     pub burn_percent: u8,
 }
 
+/// Default lamports per signature.
 pub const DEFAULT_TARGET_LAMPORTS_PER_SIGNATURE: u64 = 10_000;
+
+/// Default signatures per slot.
 pub const DEFAULT_TARGET_SIGNATURES_PER_SLOT: u64 = 50 * DEFAULT_MS_PER_SLOT;
 
 // Percentage of tx fees to burn...I think this was changed in the anza update...
```
joncinque, line 47: Looks this comment is also a bit strange, since the first one is `//` and the next line is `///` -- either way, the upcoming change rewards the full *priority* fee to validators, not base fee. And this is about base fee, so the burn percent is still valid

sdk/pinocchio/src/program_error.rs
```diff
@@ -3,7 +3,7 @@
 //! Current implementation is based on the `ProgramError` enum from
 //! the Solana SDK:
 //!
-//! https://github.com/anza-xyz/agave/blob/master/sdk/program/src/program_error.rs
+//! <https://github.com/anza-xyz/agave/blob/master/sdk/program/src/program_error.rs>
```
joncinque, line 6: May as well update this link to point to the sdk repo since it's no longer correct :smile:

sdk/pinocchio/src/program.rs
```diff
@@ -166,21 +166,15 @@ pub fn slice_invoke_signed(
...
 ///
 /// If any of the writable accounts passed to the callee contain data that is
 /// borrowed within the calling program, and that data is written to by the
 /// callee, then Rust's aliasing rules will be violated and cause undefined
 /// behavior.
-///
-/// # Important
-///
-/// The accounts on the `account_infos` slice must be in the same order as the
-/// `accounts` field of the `instruction`.
```
joncinque, line 183: Why did you remove this? It's pretty important!

sdk/pinocchio/src/program.rs
```diff
@@ -189,21 +183,15 @@ pub unsafe fn invoke_unchecked(instruction: &Instruction, accounts: &[Account])
...
```
joncinque, line 206: Same here, this is pretty important! You can also add it under the "Safety" section if you'd like

.github/workflows/publish.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         with:
           cargo-cache-key: cargo-publish
           toolchain: test
-          components: semver-checks
+          components: release, semver-checks
```
joncinque, line 64: This is my own ignorance -- does this work as expected? I thought you had to put lists in between brackets, ie. `[release, semver-checks]`

sdk/pinocchio/src/lib.rs
```diff
@@ -1,12 +1,222 @@
-//! A library to build a Solana program in Rust.
+//! # Pinocchio
 //!
-//! This library is intended to be used by on-chain programs only. It provides
-//! a zero-dependency library to minimise dependencies conflits. For off-chain
-//! programs, use instead the [`solana-sdk`] crate, which re-exports all modules
-//! from [`solana-program`].
+//! Pinocchio is a zero-dependency library to create Solana programs in Rust.
+//! It takes advantage of the way SBF loaders serialize the program input parameters
```
joncinque, line 4: nit: let's say `SVM loaders` since we're moving away from using the term `SBF` in most places

sdk/pinocchio/src/lib.rs
```diff
@@ -1,12 +1,222 @@
...
+//! default_panic_handler!();
+//! ```
+//! Any of these macros can be replaced by other implementations and Pinocchio
+//! offers a couple of variants for this.
+//!
+//! ### [`lazy_program_entrypoint!`]
+//!
+//! The [`entrypoint!`] macro looks similar to the "standard" one found in
+//! [`solana-program`](https://docs.rs/solana-program-entrypoint/latest/solana_program_entrypoint/macro.entrypoint.html).
+//! It parsers the whole input and provides the `program_id`, `accounts` and
```
joncinque, line 75: nit typo
```suggestion
//! It parses the whole input and provides the `program_id`, `accounts` and
```

sdk/pinocchio/src/lib.rs
```diff
@@ -1,12 +1,222 @@
...
+//!     Ok(())
+//! }
+//! ```
+//!
+//! The [`InstructionContext`](entrypoint::InstructionContext) provides on-demand
+//! access to the information of the input:
+//!
+//! * [`available()`](entrypoint::InstructionContext::available): number of available
+//!   accounts.
+//! * [`next_account()`](entrypoint::InstructionContext::next_account): parsers the
```
joncinque, line 118: nit typo
```suggestion
//! * [`next_account()`](entrypoint::InstructionContext::next_account): parses the
```

sdk/pinocchio/src/lib.rs
```diff
@@ -1,12 +1,222 @@
...
+//!   next available account (can be used as many times as accounts available).
+//! * [`instruction_data()`](entrypoint::InstructionContext::instruction_data): parsers
```
joncinque, line 120:
```suggestion
//! * [`instruction_data()`](entrypoint::InstructionContext::instruction_data): parses
```

sdk/pinocchio/src/lib.rs
```diff
@@ -1,12 +1,222 @@
...
+//!   the intruction data.
+//! * [`program_id()`](entrypoint::InstructionContext::program_id): parsers the
```
joncinque, line 122:
```suggestion
//! * [`program_id()`](entrypoint::InstructionContext::program_id): parses the
```

sdk/pinocchio/src/lib.rs
```diff
@@ -1,12 +1,222 @@
...
+//! ```ignore
+//! pinocchio = { version = "0.7.0", features = ["std"] }
+//! ```
+//!
+//! Instead of enabling the `std` feature to be able to format log messages with [`msg!`],
+//! it is recommented to use the [`pinocchio-log`](https://crates.io/crates/pinocchio-log)
+//! crate. This crate provides a lightweight `log!` macro with better compute units
+//! consumption than the standard `format!` macro without requiring the `std` library.
+//!
+//! ## Advance entrypoint configuration
```
joncinque, line 183:
```suggestion
//! ## Advanced entrypoint configuration
```

sdk/pinocchio/src/lib.rs
```diff
@@ -1,12 +1,222 @@
...
+//!
+//! The symbols emitted by the entrypoint macros &mdash; program entrypoint, global
+//! allocator and default panic handler &mdash; can only be defined once globally. If
+//! the program crate is also intended to be use as a library, it is common practice
```
joncinque, line 187:
```suggestion
//! the program crate is also intended to be used as a library, it is common practice
```

sdk/pinocchio/src/log.rs
```diff
@@ -28,13 +28,24 @@
...
+///
+/// Supports simple strings of type `&str`. The expression will be passed
+/// directly to [`sol_log`]. This is typically used for logging static strings.
+///
+/// # Examples
+///
+/// ```
+/// use pinocchio::msg;
+///
+/// msg!("verifying multisig");
```
joncinque, line 48:
```suggestion
/// ```
/// use pinocchio::msg;
///
/// msg!("verifying multisig");
/// ```
```

joncinque (commented):
Mostly small things and typos, looks great for the most part!

.github/workflows/publish.yml
febo: Actually, we are not using `[` and `]` in other places since it does not work having them. 😅

sdk/pinocchio/src/lib.rs
febo: I consistently made the same typo over multiple lines. 😅

sdk/pinocchio/src/program.rs
febo: It does not apply to the `*_unchecked` version since it does not validate that all accounts have been passed in, so the order is not important.

sdk/pinocchio/src/program.rs
febo: Same here, `*_unchecked` version does not do this validation so the order is not important.

joncinque: Ah ok, thanks!

sdk/pinocchio/src/program.rs
joncinque: Hold on, what does the syscall do if things are out of order? Does it succeed or fail? `invoke_signed` does the work to put things in the right order, and this does not. If the order doesn't matter, shouldn't we remove that logic in `invoke_signed`?

febo: It succeeds, the syscall finds the accounts by pubkey: https://github.com/anza-xyz/agave/blob/cfaf849b9385cc7f85c8086c334f3f2b893590f6/programs/bpf_loader/src/syscalls/cpi.rs#L917-L919

The order requirement only exists on pinocchio to optimize this check that the SDK does: https://github.com/anza-xyz/solana-sdk/blob/master/cpi/src/lib.rs#L257-L270, which makes multiple comparisons of pubkeys.

That is one of the main differences between the "checked" vs "unchecked" version – in the "checked" one, it validates that all accounts have been passed in and they can be borrowed.

joncinque: Ah gotcha, that makes sense, thanks!

joncinque (approved):
All good with the recent set of changes!

PR #67, opened 2025-02-10
extract pinocchio-cpi crate

kevinheavey:
further to discussion in #63

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,7 +41,15 @@ name = "pinocchio-associated-token-account"
 version = "0.1.0"
 dependencies = [
  "pinocchio",
+ "pinocchio-cpi",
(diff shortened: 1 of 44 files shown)


febo:
This is great! Let move the `instruction.rs` to this crate as well, since those types are only used for CPI. We could probably close #63 as well, since the approach that we will follow is to not include the definitions with the main crate.

kevinheavey:
instruction.rs is also used by syscalls.rs so moving it breaks this PR

kevinheavey:
instruction.rs is used by syscalls.rs so moving it would break this PR

febo:
Closed in favour of #122, which adds a "`cpi`" feature  – the refactoring to use SDK types is described here #111.

PR #68, opened 2025-02-12
update README to point to anza

nickfrosty:
### Summary of changes

Updated the readme to point to the anza-xyz for the status pills


febo (approved):
Thanks!

PR #71, opened 2025-02-21
Create token-2022 program

wjthieme:
Hi @febo. We're just toying around with pinocchio. I know https://github.com/codama-idl/codama/pull/276 is in the works but would be great if we can get the token-22 program as a pinocchio crate in the meantime. Right now this is just an exact clone of the token client. Let me know what you think!


wjthieme:
I see https://github.com/anza-xyz/pinocchio/pull/36 also exists. Happy to help there to see what it would take to get that over the finish line so that an initial version that supports t22 can be published

nabeel99:
happy to help this effort as well.

febo:
Closed in favour of #36

PR #75, opened 2025-02-26
Add comments from Jon

joncinque:
#### Problem

Pinocchio is getting a lot of adoption, but no one from Anza has reviewed it!

#### Summary of changes

Everything looks great overall, but there's a few little things to consider, especially for combining pinocchio with the core types from the sdk.

### DIFF
--- a/sdk/log/crate/src/logger.rs
+++ b/sdk/log/crate/src/logger.rs
@@ -32,6 +32,8 @@ pub struct Logger<const BUFFER: usize> {
     buffer: [MaybeUninit<u8>; BUFFER],
 
     // Remaining space in the buffer.
+    // JC nit: this isn't the remaining space in the buffer, maybe call it `len`
+    // directly?
     offset: usize,
 }
 
@@ -72,6 +74,8 @@ impl<const BUFFER: usize> Logger<BUFFER> {
                 }
             }
         } else {
+            // If someone implements `write_with_args` incorrectly, this can
+            // blow past the offset and cause undefined behavior, correct?
             self.offset += value.write_with_args(&mut self.buffer[self.offset..], args);
         }
 
@@ -91,6 +95,9 @@ impl<const BUFFER: usize> Logger<BUFFER> {
     }
 
     /// Check if the buffer is empty.
+    // JC nit: this was a bit confusing since `slice.is_empty()` means nearly
+    // the opposite, ie. BUFFER == 0 rather than offset == 0. Is this even
+    // used? If not, I'd recommend removing it or coming up with a better name
     #[inline(always)]
     pub fn is_empty(&self) -> bool {
         self.offset == 0
@@ -186,6 +193,10 @@ macro_rules! impl_log_for_unsigned_integer {
                 match *self {
                     // Handle zero as a special case.
                     0 => {
+                        // JC: I got worried that this was actually unsafe, since
+                        // there's no bounds check. Can you add a comment saying
+                        // we just checked that the buffer isn't empty, so this
+                        // is safe?
                         unsafe {
                             buffer.get_unchecked_mut(0).write(*DIGITS.get_unchecked(0));
                         }
@@ -217,6 +228,12 @@ macro_rules! impl_log_for_unsigned_integer {
                         };
 
                         // Number of available digits to write.
+                        //
+                        // JC: `offset` starts at max_digits, so if I write 1
+                        // digit, then offset will be `9`, which means available
+                        // would be `1`, which is not the number of available
+                        // digits, but rather the number of written digits. Am
+                        // I missing something?
                         let mut available = $max_digits - offset;
 
                         if precision > 0 {
@@ -250,6 +267,10 @@ macro_rules! impl_log_for_unsigned_integer {
 
                             #[cfg(target_os = "solana")]
                             {
+                                // JC: Should we do a check on the size so that
+                                // the syscall isn't always used? For example, if
+                                // `written` is less than 32 bytes, then we do a
+                                // normal copy
                                 if precision == 0 {
                                     sol_memcpy_(ptr as *mut _, source as *const _, written as u64);
                                 } else {
@@ -296,6 +317,12 @@ macro_rules! impl_log_for_unsigned_integer {
                         }
 
                         // There might not have been space for all the value.
+                        // JC: since this function uses an intermediate array
+                        // of `MaybeUninit<u8>`, it could be nicer to have this
+                        // function return the `MaybeUninit` array, so that the
+                        // logger decides how to deal with overflow if the
+                        // returned amount is too big. However, that isn't possible
+                        // at compile-time, so maybe this is the best option.
                         if overflow {
                             unsafe {
                                 let last = buffer.get_unchecked_mut(written - 1);
@@ -404,7 +431,7 @@ impl Log for &str {
         //
         // 1. No arguments were provided, so the entire string is copied to the buffer if it fits;
         //    otherwise, the buffer is filled as many characters as possible and the last character
-        //    is set to `TRUCATED`.
+        //    is set to `TRUNCATED`.
         //
         // Then cases only applicable when precision formatting is used:
         //
@@ -416,7 +443,7 @@ impl Log for &str {
         //    returned is `prefix` + number of characters copied.
         //
         // 4. The buffer is smaller than the string and the prefix: the buffer is filled with the
-        //    prefix and the last character is set to `TRUCATED`. The length returned is the length
+        //    prefix and the last character is set to `TRUNCATED`. The length returned is the length
         //    of the buffer.
         //
         // The length of the message is determined by whether a precision formatting was used or
@@ -433,7 +460,11 @@ impl Log for &str {
 
         // No truncate arguments were provided, so the entire string is copied to the buffer if
         // it fits; otherwise, the buffer is filled with as many characters as possible and the
-        // last character is set to `TRUCATED`.
+        // last character is set to `TRUNCATED`.
+        // JC nit: the variable name `offset` for a pointer is a little mean for
+        // readers :sweat-smile:, same with `length` for "length to write", and
+        // `prefix` can in fact also be a `suffix`, so maybe just be totally
+        // clear and call it `written_truncated_slice_length`
         let (offset, source, length, prefix, truncated) = if truncate_end.is_none() {
             let length = core::cmp::min(size, self.len());
             (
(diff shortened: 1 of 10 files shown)


sdk/log/crate/src/logger.rs
```diff
@@ -72,6 +74,8 @@ impl<const BUFFER: usize> Logger<BUFFER> {
                 }
             }
         } else {
+            // If someone implements `write_with_args` incorrectly, this can
+            // blow past the offset and cause undefined behavior, correct?
```
febo, line 78: Correct, good catch!

sdk/log/crate/src/logger.rs
```diff
@@ -217,6 +228,12 @@ macro_rules! impl_log_for_unsigned_integer {
                         };
 
                         // Number of available digits to write.
+                        //
+                        // JC: `offset` starts at max_digits, so if I write 1
+                        // digit, then offset will be `9`, which means available
+                        // would be `1`, which is not the number of available
+                        // digits, but rather the number of written digits. Am
+                        // I missing something?
```
febo, line 236: Yeah, the naming was not very clear. It is the number of "available" digits to write – a better name is `written` as you suggested.

sdk/log/crate/src/logger.rs
```diff
@@ -560,6 +595,8 @@ macro_rules! impl_log_for_slice {
                     }
                 }
 
+                // JC: shouldn't this call `write_with_args` instead? Or should
+                // this function be called `debug`? Or both?
```
febo, line 599: Yeah, this is a bit unclear: when you are logging a slice of `str`, the `str` values are logged in "debug" mode, i.e., the values are between quotes (`"`). This is to mimic the behaviour of `format!` for slices, where arrays are formatted with the debug flag `{:?}`.

sdk/pinocchio/src/instruction.rs
```diff
@@ -87,7 +94,8 @@ impl<'a> From<&'a AccountInfo> for Account<'a> {
             data_len: account.data_len() as u64,
             data: offset(account.raw, 88),
             owner: offset(account.raw, 40),
-            rent_epoch: 0,
+            rent_epoch: 0, // This is probably ok, but the rent epoch is in the
+                           // account info and could be used!
```
febo, line 98: We are not reading the `epoch` value since we can't get it zero-copy – it appears after the variable data of the account.

sdk/pinocchio/src/instruction.rs
```diff
@@ -41,6 +44,10 @@ pub struct ProcessedSiblingInstruction {
 #[derive(Clone)]
 pub struct Account<'a> {
     // Public key of the account.
+    // JC nit: in a few other places, you're using a reference to stand in for
+    // a pointer, ie the `pubkey` in `AccountMeta` -- could we do the same thing
+    // here for all the pointers? That way we could avoid the need for PhantomData.
+    // I just don't think it'll work for data, but the others should work.
```
febo, line 50: Here it is using pointers since we need one for the `data`, so we don't mix the notation on the same type. We would still need the `PhantomData` for the `data` pointer.

sdk/pinocchio/src/instruction.rs
```diff
@@ -171,6 +179,11 @@ impl<'a> From<&'a AccountInfo> for AccountMeta<'a> {
 #[derive(Debug, Clone)]
 pub struct Seed<'a> {
     /// Seed bytes.
+    /// JC nit: same here, since a slice is a pointer followed by a length, maybe
+    /// we can do the same thing here and avoid the phantom data?
+    /// On the flip side, that would make the types work inconsistently on
+    /// non-Solana architectures, ie. a 32-bit build would not be the as a 64-bit
+    /// build. But since Solana is always 64-bit, maybe we're ok?
```
febo, line 186: I will double check, I had problems using slices before and it seems other people as well: https://github.com/cavemanloverboy/solana-nostd-entrypoint/issues/11

sdk/pinocchio/src/program.rs
```diff
@@ -129,6 +139,10 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
 ///
 /// The accounts on the `account_infos` slice must be in the same order as the
 /// `accounts` field of the `instruction`.
+///
+/// JC: I wonder if we can remove this and replace it with an array with the
+/// maximum possible number of instruction accounts -- this uses a huge stack,
+/// but it avoids any dynamic allocation.
```
febo, line 145: Yeah, it could work for the current `64` limit.

sdk/pinocchio/src/account_info.rs
```diff
@@ -118,6 +124,9 @@ impl AccountInfo {
     }
 
     /// Program that owns this account.
+    /// JC nit: *technically*, since this is returning a ref that can get
+    /// modified, we could do the same thing as with lamports / data and track
+    /// borrows, but the unsafe behavior probably isn't such a big deal
```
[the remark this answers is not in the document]
joncinque: yes, correct, which is why I starred the *technically* part. It does break Rust's borrow rules though

sdk/log/crate/src/logger.rs
```diff
@@ -91,6 +95,9 @@ impl<const BUFFER: usize> Logger<BUFFER> {
     }
 
     /// Check if the buffer is empty.
+    // JC nit: this was a bit confusing since `slice.is_empty()` means nearly
+    // the opposite, ie. BUFFER == 0 rather than offset == 0. Is this even
+    // used? If not, I'd recommend removing it or coming up with a better name
```
febo, line 100: Now that we renamed `offset` to `len`, would it make sense to keep it? It is not used directly by the logger but might be useful for clients.

sdk/log/crate/src/logger.rs
```diff
@@ -186,6 +193,10 @@ macro_rules! impl_log_for_unsigned_integer {
                 match *self {
                     // Handle zero as a special case.
                     0 => {
+                        // JC: I got worried that this was actually unsafe, since
+                        // there's no bounds check. Can you add a comment saying
+                        // we just checked that the buffer isn't empty, so this
+                        // is safe?
```
febo, line 199: Done.

sdk/log/crate/src/logger.rs
joncinque: I'd just remove it if it isn't used -- let someone ask for it

sdk/log/crate/src/logger.rs
```diff
@@ -296,6 +317,12 @@ macro_rules! impl_log_for_unsigned_integer {
                         }
 
                         // There might not have been space for all the value.
+                        // JC: since this function uses an intermediate array
+                        // of `MaybeUninit<u8>`, it could be nicer to have this
+                        // function return the `MaybeUninit` array, so that the
+                        // logger decides how to deal with overflow if the
+                        // returned amount is too big. However, that isn't possible
+                        // at compile-time, so maybe this is the best option.
```
febo, line 325: Yeah, another option would be to change the return type for `write_with_args` to indicate the overflow. But since there is a truncate args, perhaps the truncation should be down to the each individual value?

sdk/log/crate/src/logger.rs
```diff
@@ -433,7 +460,11 @@ impl Log for &str {
 
         // No truncate arguments were provided, so the entire string is copied to the buffer if
         // it fits; otherwise, the buffer is filled with as many characters as possible and the
-        // last character is set to `TRUCATED`.
+        // last character is set to `TRUNCATED`.
+        // JC nit: the variable name `offset` for a pointer is a little mean for
+        // readers :sweat-smile:, same with `length` for "length to write", and
+        // `prefix` can in fact also be a `suffix`, so maybe just be totally
+        // clear and call it `written_truncated_slice_length`
```
febo, line 467: Totally agree changed the names of all variables.

sdk/log/crate/src/logger.rs
```diff
@@ -250,6 +267,10 @@ macro_rules! impl_log_for_unsigned_integer {
 
                             #[cfg(target_os = "solana")]
                             {
+                                // JC: Should we do a check on the size so that
+                                // the syscall isn't always used? For example, if
+                                // `written` is less than 32 bytes, then we do a
+                                // normal copy
```
febo, line 273: Spoke to Lucas about this. The compiler will always use the syscall if the size cannot be determined at compile time, so in this case the best we can do is to call the syscall.

sdk/log/crate/src/logger.rs
joncinque: Hm... not sure, we can just stick with this I think

sdk/pinocchio/src/account_info.rs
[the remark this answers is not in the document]
febo: Should we mark this function `unsafe` and explain this edge case?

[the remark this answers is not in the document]
febo: That is a good idea, but I think we still need to keep the `owner` getter. For example, if I just want to print the owner value or add it to an account statte without a specific requirement to which program is the owner, we need a way to get the value.

So we could add an `assert_owner()` and make the `owner()` unsafe.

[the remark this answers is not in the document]
febo: Or we can make the `assign()` to be unsafe.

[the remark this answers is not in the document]
joncinque: So let's summarize. To be perfectly correct, we should:

* make `owner()` unsafe
* add `assert_owner()`
* make `assign()` unsafe

Do you think that would be OK or too annoying for users?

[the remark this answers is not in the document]
febo: Another option would be to just:
* make `assign()` unsafe

No other changes. This would make it explicit that if you are using `assign` you should be aware of the potential borrows – the problem is there in a way, since that is the function that mutates the value.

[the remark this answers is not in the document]
joncinque: The unsafety doesn't only exist in `assign()` though, that's the whole point. It invalidates all existing references returned by `owner()`.

As an alternative, we could make `owner()` return a copy of the pubkey, so that people are encouraged to use `assert_owner()` instead.

No matter what, `assign()` should definitely be marked unsafe.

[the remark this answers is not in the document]
febo: Yeah, the idea of making `assign` unsafe is to signal that if you are using it you need to guarantee you don't have active references to `owner`. It probably won't be used as much as `owner`.

Between having the `owner` return a copy or having it unsafe, I am leaning towards making `owner()` and `assign()` unsafe and having a `is_owned_by(&Pubkey) -> bool` helper. That gives the option of anyone that needs a reference for another context to have the option of getting the reference.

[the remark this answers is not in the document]
joncinque: Works for me!

PR #78, opened 2025-02-28
Address review comments

febo:
This PR addresses the comments from the #75 review.

Closes #75


sdk/pinocchio/src/cpi.rs
```diff
@@ -139,7 +140,13 @@ pub fn slice_invoke_signed(
...
     }
 
-    let mut accounts = std::vec::Vec::with_capacity(account_infos.len());
+    if account_infos.len() > MAX_CPI_ACCOUNTS {
+        return Err(ProgramError::InvalidArgument);
+    }
+
+    const UNINIT: MaybeUninit<Account> = MaybeUninit::<Account>::uninit();
+    let mut accounts = [UNINIT; MAX_CPI_ACCOUNTS];
+    let mut index = 0;
```
joncinque, line 149: nit: maybe call this `len` or `current_len` to be clearer?

joncinque (commented):
Looks really good! Just some small things

sdk/log/crate/src/lib.rs
```diff
@@ -262,4 +262,186 @@ mod tests {
...
+    macro_rules! generate_signed_test_case {
+        ( $ty:ident, $max_len:literal, $($size:expr),+ $(,)? ) => {
+            $(
+                let mut logger = Logger::<$size>::default();
+                logger.append($ty::MIN);
+                // Signed types have one extra character for the `-` sign.
+                assert!((*logger).len() <= ($max_len + 1));
+            )*
+        };
+    }
```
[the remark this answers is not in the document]
febo: Good point!

sdk/log/crate/src/lib.rs
```diff
@@ -262,4 +262,186 @@ mod tests {
...
+                assert_eq!((*logger).len(), core::cmp::min($str.len(), $size));
+            )*
+        };
+    }
+
+    /// Generate a test function for unsigned (numeric) types.
+    ///
+    /// The main objective is to ensure that the logger will not panic when appending
+    /// a value of the type to the buffer.
+    macro_rules! fn_test_unsigned_logger_buffer {
```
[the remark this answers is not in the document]
febo: Good point! I think this refactoring made the test code easier to read.

sdk/log/crate/src/lib.rs
```diff
@@ -263,62 +293,13 @@ mod tests {
...
-            fn $test_name() {
-                $(
-                    generate_unsigned_test_case!($ty, $max_len, 1,
+    #[test]
+    fn test_logger_buffer_size_unsigned() {
+        // Test case for an unsigned numeric type.
+        macro_rules! unsigned_test_case {
+            ( $( ($ty:ident, $max_len:literal) ),+ $(,)? ) => {
+                    $(
+                        generate_numeric_test_case!($ty::MIN, ($max_len + 1), 1,
```
joncinque, line 302: Since this is unsigned, let's use MAX and remove the + 1, like it was before. Otherwise they're very similar to the signed test cases
```suggestion
                        generate_numeric_test_case!($ty::MAX, $max_len, 1,
```

joncinque (commented):
One last little thing, then this can go in!

joncinque (approved):
Great, thanks!

joncinque (approved):
Looks even better!

sdk/pinocchio/src/cpi.rs
```diff
@@ -10,6 +10,9 @@ use crate::{
     ProgramResult,
 };
 
+/// Maximum number of accounts that can be passed to a cross-program invocation.
+const MAX_CPI_ACCOUNTS: usize = 64;
```
publicqi, line 14: shouldn't it be same as `MAX_TX_ACCOUNTS` which is 128? @febo

febo: The current limit is still `64`, so for now this is ok. We can't fit an array of `Account` of 128 in the stack.

https://github.com/anza-xyz/pinocchio/blob/main/sdk/pinocchio/src/cpi.rs#L148

PR #80, opened 2025-02-28
simplify program error

publicqi:
current `ProgramError` is copied from solana-sdk which is super redundant.

this PR refactors `ProgramError`. it uses one u8 as the discriminator, and an optional 4 bytes for custom error code.

For `u64` to `ProgramError` conversion, previously the code generated was huge:

image

Now it simplifies to

image

And for `ProgramError` to `u64` conversion (i.e. return of entrypoint), the code reduced from

image

to

image

### DIFF
--- a/sdk/pinocchio/src/program_error.rs
+++ b/sdk/pinocchio/src/program_error.rs
@@ -121,11 +121,11 @@ impl From<ProgramError> for u64 {
 
 impl From<u64> for ProgramError {
     fn from(value: u64) -> Self {
-        let discriminant = (value >> BUILTIN_BIT_SHIFT) as u32;
+        let discriminant = (value >> BUILTIN_BIT_SHIFT) as u8;
         match discriminant {
             1 => Self::Custom(0),
             // SAFETY: the discriminant is always the first field for non Custom
-            2..=26 => unsafe { (*(&discriminant as *const u32 as *const ProgramError)).clone() },
+            2..=26 => unsafe { (*(&discriminant as *const u8 as *const ProgramError)).clone() },
             _ => Self::Custom(value as u32),
         }
     }
--- a/sdk/pinocchio/src/program_error.rs
+++ b/sdk/pinocchio/src/program_error.rs
@@ -11,228 +11,145 @@
 
 /// Reasons the program may fail.
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[repr(u8)]
 pub enum ProgramError {
     /// Allows on-chain programs to implement program-specific error types and see them returned
     /// by the Solana runtime. A program-specific error may be any type that is represented as
     /// or serialized to a u32 integer.
     ///
     /// Custom program error: `{0:#x}`
-    Custom(u32),
+    Custom(u32) = 1,
 
     /// The arguments provided to a program instruction were invalid
-    InvalidArgument,
+    InvalidArgument = 2,
 
     /// An instruction's data contents was invalid
-    InvalidInstructionData,
+    InvalidInstructionData = 3,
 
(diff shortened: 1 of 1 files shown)


febo:
@publicqi Did you check the impact on the program size and CU? I did a quick test and it seems that the "simplified" version increases the program size by `~1kb` and CU by `2`. Not a lot, but I was expecting the opposite.

publicqi:
do you have an example program? i checked and the size actually went down. maybe we should add some base examples in the repo to reflect improvements

also i’m digging into some rustc issues. it seems that impl display for u64 occupied a lot of space

publicqi:
i am comparing ping program between master(01c1c7a3) with this branch (4160b7e4).

number of ebpf instruction reduced from 717 to 652. size of the elf reduced from 8424 bytes to 7720 bytes.

no other changes to the ping program except for some dependency updates

```diff
diff --git a/program/Cargo.toml b/program/Cargo.toml
index df1c7bf..0cb6ce9 100644
--- a/program/Cargo.toml
+++ b/program/Cargo.toml
@@ -15,4 +15,5 @@ account-dependencies = []
 crate-type = ["cdylib", "lib"]
 
 [dependencies]
-pinocchio = { version="^0", features=["macro"], git="https://github.com/febo/pinocchio.git" }
+pinocchio = { path = "../../pinocchio/sdk/pinocchio" }
+pinocchio-pubkey = { path = "../../pinocchio/sdk/pubkey" }
\ No newline at end of file
diff --git a/program/src/lib.rs b/program/src/lib.rs
index 1b679f7..db2c49a 100644
--- a/program/src/lib.rs
+++ b/program/src/lib.rs
@@ -1,5 +1,7 @@
+use pinocchio_pubkey::declare_id;
+
 pub mod entrypoint;
 pub mod instruction;
 pub mod processor;
 
-pinocchio::declare_id!("Ping111111111111111111111111111111111111111");
+declare_id!("Ping111111111111111111111111111111111111111");
```

toolchain version:
```
cargo build-sbf -V
solana-cargo-build-sbf 2.1.14
platform-tools v1.43
rustc 1.79.0
```

publicqi:
so the program size both in number of ebpf instruction emitted and the elf size reduced.

how did you compare the CU?

febo:
> so the program size both in number of ebpf instruction emitted and the elf size reduced.
> 
> how did you compare the CU?

I executed an instruction that generates an error, so it triggers the conversion to `u64`.

publicqi:
emm that's probably because in the old code, the `ProgramError` -> `u64` conversion is just branch matching (see screenshot above). in the new code it's now doing shifting. can you try for all types of program errors what are the diffs? i would expect some conversion will take a bit more, but the size should reduce by a lot.

febo:
> @febo what more is needed to push this one?

I will do some more testing.

publicqi:
ok let me know if you need help here.

febo:
We should integrate this change into the SDK, since soon we will be sharing the same `ProgramError` type (#112). I will still do some more testing with this PR to evaluate the impact.

publicqi:
is adding the `#[repr(u8)]` and using the `unsafe` a breaking change for downstream? how will SDK handling the versioning?

febo:
> isn't adding the `#[repr(u8)]` and using the `unsafe` a breaking change for downstream? how will SDK handling the versioning?

We are already introducing a breaking change, so it won't be a problem to have one more. 😅

https://github.com/anza-xyz/solana-sdk/issues/84

febo:
`ProgramError` does not belong to pinocchio anymore, so this needs to be created on the SDK repository.

PR #81, opened 2025-03-01
sdk: Avoid memory initialization on create program address

### LINKED ISSUE #79: unnecessary memory initialization
```rust
// Note: since the array will be returned, it is "cheaper" to have
// it initialized instead of having to do a transmute later.
let mut bytes = [0u8; 32];

let result = unsafe {
    crate::syscalls::sol_create_program_address(
        seeds as *const _ as *const u8,
        seeds.len() as u64,
        program_id as *const _ as *const u8,
        &mut bytes as *mut _ as *mut u8,
    )
};

match result {
    crate::SUCCESS => Ok(bytes),
    _ => Err(result.into()),
}
```

i doubt that initializing the `bytes` to 0 will be "cheaper". we only need to preserve 32 bytes. is there something i'm missing?

also i tried analyze an example ebpf artifact between `MaybeUninit` and the current impl. 

Image
Image

febo:
### Problem

As described on #79, `create_program_address` and `try_find_program_address` are currently initializing the memory for the derived pubkey. This is unnecessary since the memory is initialized by the syscall.

### Solution

Avoid memory initialization by using a `MaybeUninit` array.

Fixes #79

### DIFF
--- a/sdk/pinocchio/src/pubkey.rs
+++ b/sdk/pinocchio/src/pubkey.rs
@@ -120,21 +120,21 @@ pub fn find_program_address(seeds: &[&[u8]], program_id: &Pubkey) -> (Pubkey, u8
 pub fn try_find_program_address(seeds: &[&[u8]], program_id: &Pubkey) -> Option<(Pubkey, u8)> {
     #[cfg(target_os = "solana")]
     {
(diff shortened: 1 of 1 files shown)


publicqi:
`try_find_program_address` too

PR #84, opened 2025-03-03
fix panic_handler

publicqi:
```rust
#![no_std]
use pinocchio::{
    account_info::AccountInfo, default_allocator, default_panic_handler, program_entrypoint,
    pubkey::Pubkey, ProgramResult,
};

program_entrypoint!(process_instruction);
default_allocator!();
default_panic_handler!();

fn process_instruction(_: &Pubkey, _: &[AccountInfo], _: &[u8]) -> ProgramResult {
    Ok(())
}
```

```toml
[package]
name = "pinocchio-test"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib", "lib"]

[dependencies]
pinocchio = { version = "0.7.1" }

[profile.release]
# strip = "symbols"
lto = true
opt-level = 3
```

This code doesn't compile now. Error message is
```
error: `#[panic_handler]` function required, but not found
```

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -220,10 +220,11 @@ pub unsafe fn deserialize<'a, const MAX_ACCOUNTS: usize>(
     (program_id, processed, instruction_data)
 }
 
-/// Default panic handler.
+/// Default panic hook (std).
 ///
-/// This macro sets up a default panic handler that logs the panic message and the file where the
-/// panic occurred.
+/// This macro sets up a default panic hook that logs the panic message and the file where the
+/// panic occurred. Syscall "abort()" will be called after it returns. It acts as a hook after
+/// rust runtime panics.
 ///
 /// Note that this requires the `"std"` feature to be enabled.
 #[cfg(feature = "std")]
@@ -240,11 +241,11 @@ macro_rules! default_panic_handler {
     };
 }
 
-/// Default panic handler.
+/// Default panic hook (no std).
 ///
-/// This macro sets up a default panic handler that logs the file where the panic occurred.
+/// This macro sets up a default panic hook that logs the file where the panic occurred.
 ///
-/// This is used when the `"std"` feature is disabled.
+/// This is used when the `"std"` feature is disabled and program is `std`.
 #[cfg(not(feature = "std"))]
 #[macro_export]
 macro_rules! default_panic_handler {
@@ -258,6 +259,39 @@ macro_rules! default_panic_handler {
             }
             // Panic reporting.
             $crate::log::sol_log("** PANICKED **");
+        }
+    };
+}
+
+/// A rust panic handler for `no_std`.
+///
+/// When all crates are `no_std`, we need to define a global `#[panic_handler]`.
+/// It takes over the default rust panic handler.
+///
+/// This macro is used when the `"std"` feature is disabled.
+#[cfg(not(feature = "std"))]
+#[macro_export]
+macro_rules! nostd_panic_handler {
+    () => {
+        /// A panic handler for `no_std`.
+        #[cfg(target_os = "solana")]
+        #[no_mangle]
+        #[panic_handler]
+        fn handler(info: &core::panic::PanicInfo<'_>) -> ! {
+            if let Some(location) = info.location() {
+                unsafe {
+                    $crate::syscalls::sol_panic_(
+                        location.file().as_ptr(),
+                        location.file().len() as u64,
+                        location.line() as u64,
+                        location.column() as u64,
+                    )
+                }
+            } else {
+                // Panic reporting.
+                $crate::log::sol_log("** PANICKED **");
+                unsafe { $crate::syscalls::abort() }
+            }
         }
     };
 }
--- a/sdk/pinocchio/src/syscalls.rs
+++ b/sdk/pinocchio/src/syscalls.rs
@@ -75,6 +75,8 @@ define_syscall!(fn sol_get_epoch_rewards_sysvar(addr: *mut u8) -> u64);
 define_syscall!(fn sol_poseidon(parameters: u64, endianness: u64, vals: *const u8, val_len: u64, hash_result: *mut u8) -> u64);
 define_syscall!(fn sol_remaining_compute_units() -> u64);
 define_syscall!(fn sol_alt_bn128_compression(op: u64, input: *const u8, input_size: u64, result: *mut u8) -> u64);
+define_syscall!(fn abort() -> !);
+define_syscall!(fn sol_panic_(filename: *const u8, filename_len: u64, line: u64, column: u64) -> !);
 
 #[cfg(target_feature = "static-syscalls")]
 pub const fn sys_hash(name: &str) -> usize {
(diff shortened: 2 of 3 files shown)


febo:
Did you try compiling without being `#![no_std]`? I think when you link the `std`, this change might conflict.

febo:
Perhaps instead of modifying the existing macro, we need to add one specific for `no_std`:
```rust
#[macro_export]
macro_rules! nostd_panic_handler {
    () => {
        /// A panic handler for `no_std`.
        #[cfg(not(test))]
        #[no_mangle]
        #[panic_handler]
        fn custom_panic(info: &core::panic::PanicInfo<'_>) -> ! {
            if let Some(location) = info.location() {
                $crate::log::sol_log(location.file());
            }
            // Panic reporting.
            $crate::log::sol_log("** PANICKED **");
            // Never returns.
            loop {}
        }
    };
}
```

publicqi:
yeah that sounds good.

publicqi:
i had a wrong understanding about how panic works in sbf and now it's corrected:

std:
when a panic happens, rust runtime will call
```
rust_begin_unwind (std generted)
    std::sys::pal::sbf::panic (sbf compiler generated)
        custom_panic (pinocchio)
        abort syscall (sbf compiler generated)
```

no_std:
```
rust_begin_unwind (pinocchio)
```

so when using std, we define `custom_panic` as a hook and abort will be called eventually. but in no_std, we will need to manually implement `rust_begin_unwind`.

sdk/pinocchio/src/syscalls.rs
```diff
@@ -75,6 +75,7 @@ define_syscall!(fn sol_get_epoch_rewards_sysvar(addr: *mut u8) -> u64);
 define_syscall!(fn sol_poseidon(parameters: u64, endianness: u64, vals: *const u8, val_len: u64, hash_result: *mut u8) -> u64);
 define_syscall!(fn sol_remaining_compute_units() -> u64);
 define_syscall!(fn sol_alt_bn128_compression(op: u64, input: *const u8, input_size: u64, result: *mut u8) -> u64);
+define_syscall!(fn abort() -> !);
```
febo, line 78: Let's add the `sol_panic_` syscall as well:
```rust
define_syscall!(fn sol_panic_(filename: *const u8, filename_len: u64, line: u64, column: u64) -> !);
```

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -252,12 +254,24 @@ macro_rules! default_panic_handler {
         /// Default panic handler.
         #[cfg(all(not(feature = "custom-panic"), target_os = "solana"))]
         #[no_mangle]
-        fn custom_panic(info: &core::panic::PanicInfo<'_>) {
+        #[panic_handler]
```
febo, line 257: Let's remove the `#[panic_handler]` attribute from here. The `default_panic_handler` will be used when you import `pinocchio` as a `no_std` but your program links against the `std`.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -252,12 +254,24 @@ macro_rules! default_panic_handler {
...
+
+        /// A dummy handler for clippy
+        ///
+        /// To make clippy work, add panic = "abort" under [profile.dev] in Cargo.toml
+        #[cfg(all(not(feature = "custom-panic"), not(target_os = "solana")))]
+        #[no_mangle]
+        #[panic_handler]
+        fn dummy_handler(info: &core::panic::PanicInfo<'_>) -> ! {
+            unsafe { core::hint::unreachable_unchecked() }
         }
```
febo, line 275: This might not be needed with the proposed change of having an specific `nostd` panic handler.

febo:
When your program is `no_std`:
```rust
program_entrypoint!(...);
no_allocator!();
nostd_panic_handler!();
```

When your program links against the `std`, then you can use either of the `default_panic_handler` - it will depende whether you are importing `pinocchio` as `no_std` or not.

sdk/pinocchio/src/entrypoint/mod.rs
publicqi: yeah let me move it to another pr.

sdk/pinocchio/src/entrypoint/mod.rs
publicqi: so four cases here:
+ if program is std, they'll need to manually specify `features = ["std"]` in their Cargo.toml. in that case `custom-panic` (using `msg`) will be used. 
+ if a program is no_std, rustc compiler will ask them to specify a `[panic_handler]`, which is what `handler` doing.
+ if a program is no_std but specifying `features = ["std"]`, this should be undefined behavior (e.g. `format` in pinocchio won't work anyways)
+ if a program is std while not specifying `features = ["std"]`, this is kind of shady.

to conclude,
+ with this `[panic_handler]`, we discourage the fourth case as std will generate a builtin `[panic_handler]` and they will conflict.
+ without this `[panic_handler]`, people cannot use `entrypoint!`. they need to use `program_entrypoint!` and add a `[panic_handler]` themself.

publicqi:
> When your program links against the std, then you can use either of the default_panic_handler - it will depend on whether you are importing pinocchio as no_std or not:

the functionality difference between the two is only CU diff and format? can we just deprecate the std `custom-panic`?

sdk/pinocchio/src/entrypoint/mod.rs
febo: Few comments:
* in the first case, you could still have `pinocchio` as "no_std" even though your program is `std`. That would mean `pinocchio` won't use anything from the `std` but your program might.
* third case probably will fail compilation, since `std` won't be available.
* fourth case is not an ideal scenario, but it could happen quite often. The benefit of having `pinocchio` as a "no_std" is that it won't bring all the `format!` stuff automatically (reduces the binary size), but you might still have a library in your program that requires `std`.

febo:
> > When your program links against the std, then you can use either of the default_panic_handler - it will depend on whether you are importing pinocchio as no_std or not:
> 
> the functionality difference between the two is only CU diff and format? can we just deprecate the std `custom-panic`?

Soon we can deprecate it. At the moment the one with the `format!` is able to give the dynamic panic message, while the "no_std" variation of the `default_panic_handler` does not.

In rust 1.84, `fmt::Arguments::as_str` will be stabilized (see https://github.com/rust-lang/rust/pull/132511), then we can have a single one that does not require `std`.

febo:
@publicqi Wondering if you had a chance to look into my last comment. I am preparing a release and ideally would like to include this fix.

publicqi:
yes. i'll fix it

publicqi:
@febo ready for review. updated some docs too so hopefully it's clear enough.

publicqi:
force pushed to clean the messy commits

febo:
> force pushed to clean the messy commits

Looks great – just a nit: let's remove the `feature = "custom-panic` references, they might have sneaked in when you rebased.

publicqi:
good catch. fixed.

febo:
One more thing, I think the changes to the `syscalls` module are also missing.

publicqi:
ahh yes...

publicqi:
now should be good. i rebased the wrong `main` branch. 

https://github.com/anza-xyz/pinocchio/compare/main..98245d1c44979e03a8924943bcd8916c6822ca4a

febo (approved):
Looks great – thanks!

PR #89, opened 2025-03-10
remove `custom-heap` and `custom-panic`

publicqi:
`custom-panic` and `custom-panic` were not in Cargo.toml

```rust
#[cfg(feature = "std")]
#[macro_export]
macro_rules! default_panic_handler {
    () => {
        /// Default panic handler.
        #[cfg(all(not(feature = "custom-panic"), target_os = "solana"))]
// ...
```

```rust
macro_rules! default_allocator {
    () => {
        #[cfg(all(not(feature = "custom-heap"), target_os = "solana"))]
// ...
```

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -231,7 +231,7 @@ pub unsafe fn deserialize<'a, const MAX_ACCOUNTS: usize>(
 macro_rules! default_panic_handler {
     () => {
         /// Default panic handler.
-        #[cfg(all(not(feature = "custom-panic"), target_os = "solana"))]
+        #[cfg(target_os = "solana")]
         #[no_mangle]
         fn custom_panic(info: &core::panic::PanicInfo<'_>) {
             // Panic reporting.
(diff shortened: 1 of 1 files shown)


febo:
What if we get rid of those features? They are clunky to use and not strictly needed now that we have individual macros for the program entrypoint. So anyone that wants to use custom allocator and/or panic handler should use the macro that adds only the program entrypoint.

publicqi:
i agree on that

publicqi:
i'm also thinking adding `custom_panic` and `custom_heap` doc somewhere. where do you think i should add it?

febo:
> i'm also thinking adding `custom_panic` and `custom_heap` doc somewhere. where do you think i should add it?

I don't think I understand this – we removed those features, would you like to document the removal? Alternatively, we could add a section on the readme to explain how to define custom ones by not using the `entrypoint!` macro and instead the `program_entrypoint!` one. There is a bit of that already, but we can definitely expand it.

febo:
> yeah adding doc to expand how to use custom panic/allocator

The `README.md` is probably the best place for that.

febo (approved):
This looks good. I will merge this change and we can update the `README.md` in a separate PR.

PR #91, opened 2025-03-13
Add `ToStr` trait

febo:
This PR adds a trait to convert a `ProgramError` to a `&str` value.

An example of how this could be used to log the error message in a program:
```rust
match result {
    Err(error) => {
        pinocchio::log::sol_log(error.to_str::<TokenError>());

        Err(error)
    }
    Ok(_) => Ok(()),
}
```

### DIFF
--- a/sdk/pinocchio/src/program_error.rs
+++ b/sdk/pinocchio/src/program_error.rs
@@ -4,10 +4,6 @@
 //! the Solana SDK:
 //!
 //! <https://github.com/anza-xyz/solana-sdk/blob/master/program-error/src/lib.rs>
-//!
-//! Considerations:
-//!
-//! - Not deriving `thiserror::Error` for now, as it's not clear if it's needed.
 
 /// Reasons the program may fail.
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/program_error.rs
```diff
@@ -236,3 +232,54 @@ impl From<ProgramError> for u64 {
         }
     }
 }
+
+/// A trait for converting a program error to a `&str`.
+pub trait ToStr {
+    fn to_str<E>(&self) -> &'static str
+    where
+        E: 'static + ToStr + TryFrom<u32>;
+}
```
joncinque, line 241: Does this need the `'static` lifetime? It might be more generally useful if it has a more limited lifetime

joncinque (commented):
Looks good! Confirming that these prints are the same as `PrintProgramError` in the solana-program-error

Just a question on the lifetimes

sdk/pinocchio/src/program_error.rs
febo: The issue is that it is returning a reference to a value created on the function `to_str`, so any other lifetime won't be accepted.

joncinque: Ah right, makes sense -- oh well!

joncinque (approved):

PR #94, opened 2025-03-17
feat: add support of address-lookup-table program

sonicfromnewyoke:


programs/address-lookup-table/src/instructions/extend.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn invoke(&self) -> ProgramResult {
+        self.invoke_signed(&[])
+    }
+
+    pub fn invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
+        // Instruction data:
+        // - [0]: Instruction discriminator (1 byte, u8) (2 for Extend)
+
+        // LOOKUP_TABLE_MAX_ADDRESSES == u8::MAX + 1.
+        let mut instruction_data = [0; 8196]; // TODO: huge for stack, maybe forget about no_std and use vector
```
sonicfromnewyoke, line 44: @febo What do you think about such a huge array initialization?
Max addresses to extend is `256 x 32` + `4` for discriminator.
Otherwise I can use `Vec<Pubkey>`, as it's done in the original program, but we lose `no_std` for this.

Maybe you have other, better approaches to tackle it? Maybe use `features` and make 2 realizations for `std` and `no_std` ways?

publicqi: `Vec` is just a wrapper for the allocation. both `alloc::boxed::Box` or `alloc::alloc::alloc` would allocate on the heap.

programs/address-lookup-table/src/instructions/close.rs
```diff
@@ -0,0 +1,58 @@
...
+        let account_metas: [AccountMeta; 3] = [
+            AccountMeta::writable(self.lookup_table.key()),
+            AccountMeta::readonly_signer(self.authority.key()),
+            AccountMeta::writable(self.recipient.key()),
+        ];
+
+        // Instruction data:
+        // - [0]: Instruction discriminator (1 byte, u8) (4 for Close)
+
+        let instruction_data = [4u8];
```
publicqi, line 40: the size is 4 bytes

programs/address-lookup-table/src/instructions/create.rs
```diff
@@ -0,0 +1,80 @@
...
+            AccountMeta::readonly_signer(self.account.key()),
+            AccountMeta::writable_signer(self.funding_account.key()),
+            AccountMeta::readonly(self.system_program.key()),
+        ];
+
+        // Instruction data:
+        // - [0..4 ]: Instruction discriminator (1 byte, u8) (0 for Create)
+        // - [4..12]: Recent Slot
+        // - [12   ]: bump seed
+        let mut instruction_data = [0; 13];
```
publicqi, line 58: use `MaybeUninit` to reduce unnecessary zeroed-out

programs/address-lookup-table/src/instructions/deactivate.rs
```diff
@@ -0,0 +1,51 @@
...
+        // account metadata
+        let account_metas: [AccountMeta; 2] = [
+            AccountMeta::writable(self.lookup_table.key()),
+            AccountMeta::readonly_signer(self.authority.key()),
+        ];
+
+        // Instruction data:
+        // - [0]: Instruction discriminator (1 byte, u8) (3 for Deactivate)
+
+        let instruction_data = [3u8];
```
publicqi, line 37: size is 4 bytes

programs/address-lookup-table/src/instructions/freeze.rs
```diff
@@ -0,0 +1,50 @@
...
+        // account metadata
+        let account_metas: [AccountMeta; 2] = [
+            AccountMeta::writable(self.lookup_table.key()),
+            AccountMeta::readonly_signer(self.authority.key()),
+        ];
+
+        // Instruction data:
+        // - [0]: Instruction discriminator (1 byte, u8) (1 for Freeze)
+
+        let instruction_data = [1u8];
```
publicqi, line 36: size is 4 bytes

programs/address-lookup-table/src/instructions/extend.rs
febo: `8196` won't fit on the stack, so it needs to be less than `4096` (a bit less than that since there are other things that will go on the stack). Perhaps it would be easier if we define a trait `Addresses` that represents the instruction data in this case. Then we can have `no_std` and `std` implementations for it.

We can use a `alloc::vec::Vec` for the `std` variant.

febo (commented):
This is a native program, so instruction discriminators have `4` bytes.

programs/address-lookup-table/src/instructions/extend.rs
illia-bobyr: Why is it important to support `no_std`?
I assume these programs are compiled in two variants:
1. An SBF target, which supports `std`, it just output an SBF bytecode.
2. Unit tests, that might be running in the host configuration.  This should also support `std`.  But I can also see that a lot of the basic platform API will either panic or fail on a non-`solana` platform.  So this use case might actually be limited.

sonicfromnewyoke: it requires you to use things that are "closer to the bare metal" which leads to the resource-wise solution and easier backward compatibility.

and this is easier for the performance optimizations when you have no work with a lot of overhead

mubarizkyc:
when expected  ? ,i have to copy complete deseralizer from ```solana-program``` crate

javi0guerrero4-cloud (commented):
add and review

PR #96, opened 2025-03-17
ci: Update publish workflow

febo:


.github/actions/setup/action.yml
```diff
@@ -67,35 +67,40 @@ runs:
...
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ runner.os }}-${{ inputs.cargo-cache-key }}-${{ hashFiles('**/Cargo.lock') }}
+
     - name: Install 'cargo-audit'
-      if: ${{ contains(inputs.components, 'audit') }}
+      if: ${{ steps.cache-cargo.outputs.cache-hit != 'true' && contains(inputs.components, 'audit') }}
```
joncinque, line 84: Rather than adding all of these checks, which might could have a false positive since the extra tools are not part of the lockfile, it's also possible to use https://github.com/taiki-e/cache-cargo-install-action to install tools

joncinque (approved):
Looks good to me! Just a suggestion for the caching

.github/actions/setup/action.yml
```diff
@@ -69,33 +69,30 @@ runs:
...
+      uses: taiki-e/cache-cargo-install-action@v2
       with:
-        path: |
-          ~/.cargo/bin/
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-        key: ${{ runner.os }}-${{ inputs.cargo-cache-key }}-${{ hashFiles('**/Cargo.lock') }}
-        restore-keys: ${{ runner.os }}-${{ inputs.cargo-cache-key }}
```
joncinque, line 101: Oh sorry it wasn't clear, you should definitely still keep the cache for everything else!

febo: Will this cache be used? The cargo install action includes a cache.

joncinque: It should be used for everything you build in the workflow, right?

febo: Yeah, you are right – added the cache back.

.github/actions/setup/action.yml
```diff
@@ -69,26 +69,36 @@ runs:
...
+
+    - name: Install 'cargo-edit'
+      if: ${{ contains(inputs.components, 'edit') }}
+      uses: taiki-e/cache-cargo-install-action@v2
+      with:
+        tool: cargo-edit@0.13.1
 
     - name: Cache Cargo Dependencies
-      if: ${{ inputs.cargo-cache-key && !inputs.cargo-cache-fallback-key }}
+      if: ${{ inputs.cargo-cache-key }}
```
joncinque, line 101: Was this change intentional?

.github/actions/setup/action.yml
```diff
@@ -98,4 +108,3 @@ runs:
           ~/.cargo/git/db/
           target/
         key: ${{ runner.os }}-${{ inputs.cargo-cache-key }}-${{ hashFiles('**/Cargo.lock') }}
-        restore-keys: ${{ runner.os }}-${{ inputs.cargo-cache-key }}
```
joncinque, line 101: Was this change intentional?

.github/workflows/publish.yml
```diff
@@ -40,9 +49,39 @@ env:
...
+
+      - name: Set Version
+        run: |
+          if [ "${{ inputs.level }}" == "version" ]; then
+            LEVEL=${{ inputs.version }}
+          else
+            LEVEL="--bump ${{ inputs.level }}"
+          fi
+
+          cargo set-version --manifest-path ${{ inputs.crate }}/Cargo.toml $LEVEL
```
joncinque, line 74: Actually, sorry I only just thought about this, but `cargo release` can actually do this too if you run:

```
cargo release <LEVEL> --no-tag --no-publish --no-push --no-confirm --execute
```

And that way we don't need to use a new tool. What do you think?

.github/actions/setup/action.yml
febo: Yes, the fallback key was not an input.

.github/actions/setup/action.yml
febo: Yes, this was to restore stale cache when there is no cache hit, which I don't think we want to do.

.github/workflows/publish.yml
febo: Ah, I see, yeah that could a be a good idea.

.github/actions/setup/action.yml
```diff
@@ -69,26 +69,36 @@ runs:
...
-      run: cargo install cargo-semver-checks
+      uses: taiki-e/cache-cargo-install-action@v2
+      with:
+        tool: cargo-semver-checks@0.39.0
+
+    - name: Install 'cargo-edit'
+      if: ${{ contains(inputs.components, 'edit') }}
+      uses: taiki-e/cache-cargo-install-action@v2
+      with:
+        tool: cargo-edit@0.13.1
```
joncinque, line 98: This shouldn't be needed anymore

joncinque (approved):
One more tiny thing, then this can go in!

joncinque (approved):

PR #97, opened 2025-03-18
ci: Add cargo registry token check

febo:
### Problem

Currently the publish workflow does not check if the cargo registry token is set.

### Solution

Add a step to check whether the cargo registry token is set or not.


febo:
The check was already done at the start of the workflow.

PR #102, opened 2025-03-19
pubkey: Add `no_std` attribute

febo:
### Problem

Currently the `pinocchio-pubkey` crate is no_std "friendly" but does not include the no_std attribute, same as its current `five8_const` dependency. This brings the `std` during the link stage of the build.

### Solution

Add the no_std attribute and update `five8_const` dependency, which is also no_std now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,18 +13,18 @@ dependencies = [
 
 [[package]]
 name = "five8_const"
-version = "0.1.3"
+version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "72b4f62f0f8ca357f93ae90c8c2dd1041a1f665fde2f889ea9b1787903829015"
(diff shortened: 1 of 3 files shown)


joncinque (approved):
Very nice!

sonicfromnewyoke (commented):
the only thing that i guess should be added is `categories` field like follows
```toml
# Cargo.toml
# ...
categories = ["no-std"]

# ...
```
more info about `categories` can be found [here](https://crates.io/category_slugs)

febo:
> the only thing that i guess should be added is `categories` field like follows
> 
> ```toml
> # Cargo.toml
> # ...
> categories = ["no-std"]
> 
> # ...
> ```
> 
> more info about `categories` can be found [here](https://crates.io/category_slugs)

Nice! Let's do that in a separate PR and we can go over all crates.

PR #104, opened 2025-03-21
add dummy handler for non-solana target

publicqi:
it's usual for rust devs to use `cargo clippy` or `cargo check`. when it's no_std, we need to specify the `panic_handler`

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -293,6 +293,16 @@ macro_rules! nostd_panic_handler {
                 unsafe { $crate::syscalls::abort() }
             }
         }
+
+        /// A placeholder for `cargo clippy`
+        ///
+        /// Add `panic = "abort"` to `[profile.dev]` in `Cargo.toml` for clippy
+        #[cfg(not(target_os = "solana"))]
+        #[no_mangle]
+        #[panic_handler]
+        fn handler(info: &core::panic::PanicInfo<'_>) -> ! {
+            unsafe { core::hint::unreachable_unchecked() }
+        }
     };
 }
 


sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -293,6 +293,16 @@ macro_rules! nostd_panic_handler {
...
         }
+
+        /// A placeholder for `cargo clippy`
+        ///
+        /// Add `panic = "abort"` to `[profile.dev]` in `Cargo.toml` for clippy
+        #[cfg(not(target_os = "solana"))]
+        #[no_mangle]
+        #[panic_handler]
+        fn handler(info: &core::panic::PanicInfo<'_>) -> ! {
+            unsafe { core::hint::unreachable_unchecked() }
```
febo, line 304: What is the difference between using just `loop {}` and `unsafe { core::hint::unreachable_unchecked() }`? This probably won't make a difference since this handler should not be used, but in the odd case that it does, the `unreachable_unchecked` function has undefined behaviour.

febo (commented):
We should add this, just have one question about the implementation.

sdk/pinocchio/src/entrypoint/mod.rs
publicqi: i believe they are equivalent on the llvm ir level. let me do some research about this.

publicqi:
so `loop {}` does compile to something like
```asm
.L1:
    jmp .L1
```

and `unreachable_unchecked!` will tell the compiler this is unreachable. both works here since it's truly "unreachable", and there's no side effect from the llvm.

the self jumping pattern, according to [this post](https://users.rust-lang.org/t/why-panic-handler-returns-never-type/99941/3), is quite useful on bare metal programming which makes sense.

i would prefer `unreachable` since semantically it means unreachable instead of halting.

febo (approved):

PR #105, opened 2025-03-22
add memo program

publicqi:


sonicfromnewyoke:
hasn't it  already done in the scope of [p-memo](https://github.com/febo/p-memo)

publicqi:
oh i don't really know its existence. this is a wrapper to call spl-memo.

programs/memo/src/instructions/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+        let instruction = Instruction {
+            program_id: &crate::ID,
+            accounts: unsafe {
+                core::slice::from_raw_parts(account_metas.as_ptr() as _, num_accounts)
+            },
+            data: self.memo,
+        };
+
+        // SAFETY: accounts are validated
+        unsafe {
```
febo, line 62: You can replace this by a call to `slice_invoke_signed`. This also means that you don't need to create accounts array:
```rust
let mut accounts = [UNINIT_ACCOUNT; MAX_CPI_ACCOUNTS];
```

programs/memo/src/instructions/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    ProgramResult,
+};
+
+/// Memo instruction.
+///
+/// ### Accounts:
+///   0-N. `[SIGNER]` Signers
+pub struct Memo<'a> {
+    /// Signing accounts
+    pub account_infos: &'a [&'a AccountInfo],
```
febo, line 16: tiny-nit: should we call this `signers` perhaps?

febo, line 16: tiny-nit: should we call this `signers` or `signer_accounts`?

programs/memo/src/instructions/mod.rs
publicqi: this is to avoid the checks inside existing `invoke_` methods.

programs/memo/src/instructions/mod.rs
publicqi: sg

programs/memo/src/instructions/mod.rs
febo: In that case the function needs to be made `unsafe` since it is not doing the borrow check and have the `unchecked` suffix. We should also have a "safe" variant.

publicqi: makes sense. after reading some code i found the check for borrowing is actually necessary. i'll use `slice_invoke_signed`

programs/memo/src/instructions/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+        const UNINIT_META: MaybeUninit<AccountMeta> = MaybeUninit::<AccountMeta>::uninit();
+
+        // We don't know num_accounts at compile time, so we use MAX_CPI_ACCOUNTS
+        let mut account_metas = [UNINIT_META; MAX_CPI_ACCOUNTS];
+
+        let num_accounts = self.signers.len();
+
+        for i in 0..num_accounts {
+            unsafe {
+                // SAFETY: num_accounts is less than MAX_CPI_ACCOUNTS
```
febo, line 37: We should validate that `num_accounts` is less than `MAX_CPI_ACCOUNTS`.

programs/memo/src/instructions/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// Memo instruction.
+///
+/// ### Accounts:
+///   0. `..+N` `[SIGNER]` N signing accounts
+pub struct Memo<'a> {
+    /// Signing accounts
+    pub signers: &'a [&'a AccountInfo],
+    /// Memo
+    pub memo: &'a [u8],
```
febo, line 18: Let's make this a `&'a str` since we are expecting a "string" slice.

programs/memo/src/instructions/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+        if num_accounts > MAX_CPI_ACCOUNTS {
+            return Err(ProgramError::InvalidArgument);
+        }
+
+        for i in 0..num_accounts {
+            unsafe {
+                // SAFETY: num_accounts is less than MAX_CPI_ACCOUNTS
+                account_metas
+                    .get_unchecked_mut(i)
+                    .write(AccountMeta::readonly_signer(self.signers[i].key()));
```
febo, line 44: We could replace `self.signers[i]` for `self.signers.get_unchecked(i)` to same a bit of CU.

programs/memo/README.md
```diff
@@ -0,0 +1,46 @@
...
+This will add the `pinocchio-memo` dependency to your `Cargo.toml` file.
+
+## Examples
+
+Creating a memo:
+```rust
+// Both accounts should be signers
+Memo {
+    account_infos: &[&account_infos[0], &account_infos[1]],
+    memo: b"hello",
```
febo, line 39: nit: `b"hello"` -> `"hello"`

programs/memo/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This will add the `pinocchio-memo` dependency to your `Cargo.toml` file.
+
+## Examples
+
+Creating a memo:
+```rust
+// Both accounts should be signers
+Memo {
+    account_infos: &[&account_infos[0], &account_infos[1]],
```
febo, line 38: Sorry, just noticed one more: `account_infos` -> `signers`

programs/memo/src/instructions/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    program_error::ProgramError,
+    ProgramResult,
+};
+
+/// Memo instruction.
+///
+/// ### Accounts:
+///   0. `..+N` `[SIGNER]` N signing accounts
+pub struct Memo<'a> {
+    /// Signing accounts
```
febo, line 16: This might be the last of the last one 😅, but I think it will be better to separate the lifetimes here. Otherwise might get tricky to use it.
```rust
pub struct Memo<'a, 'b, 'c> {
    /// Signing accounts
    pub signers: &'b [&'a AccountInfo],
    /// Memo
    pub memo: &'c str,
}

publicqi: agreed

PR #106, opened 2025-03-22
Feature: Instructions Sysvar Support

0xRigel:
### Motivation
There is currently no support for the Instructions Sysvar in Pinocchio, so @L0STE and I decided to implement the SDK under the supervision of @deanmlittle. Our main goals were to make it available for anyone to use, and also to improve performance and modularity—particularly by leveraging and following Pinocchio’s existing zero-copy and `no-std` design philosophy.

### Implementation Details
In the original implementation, each instruction was fully deserialized and allocated as a complete structure in memory—even though the instruction data already resided in the Sysvar account itself and isn't necessarily always fully needed by programs. This approach created unnecessary memory and CU overhead.

To optimize this process, the new Instructions Sysvar SDK creates a "view" into the Sysvar data at the exact offset for the specific instruction being introspected. Instead of deserializing the entire structure, it provides direct access to `accounts`, `instruction_data`, and `program_id` through calculated memory offsets. This approach significantly reduces memory allocations and improves execution efficiency by only accessing the required data when needed.

This is implemented in the [IntrospectedInstruction](https://github.com/0xRigel/pinocchio/blob/92d36d538043fc26e4dcd557c4ffa5f9e0829d5a/sdk/pinocchio/src/sysvars/instructions.rs#L120C1-L124C2) struct and all it's helpers:

```rust
#[repr(C)]
#[derive(Clone, PartialEq, Eq)]
pub struct IntrospectedInstruction {
    pub raw: *const u8,
}

impl IntrospectedInstruction {
    pub unsafe fn get_account_meta_at_unchecked(&self, index: usize) -> IntrospectedAccountMeta {
        IntrospectedAccountMeta {
            raw: self
                .raw
                .add(size_of::<u16>() + index * IntrospectedAccountMeta::SPACE)
                    as *const u8,
        }
    }

    pub fn get_account_meta_at(
        &self,
        index: usize,
    ) -> Result<IntrospectedAccountMeta, SanitizeError> {
        if index >= unsafe { *(self.raw as *const u16) } as usize {
            return Err(SanitizeError::IndexOutOfBounds);
        }

        Ok(unsafe { self.get_account_meta_at_unchecked(index) })
    }

    pub fn get_program_id(&self) -> &Pubkey {
        unsafe {
            let num_accounts = *(self.raw as *const u16);
            &*(self
                .raw
                .add(size_of::<u16>() + num_accounts as usize * IntrospectedAccountMeta::SPACE)
                    as *const Pubkey)
        }
    }

    pub fn get_instruction_data(&self) -> &[u8] {
        unsafe {
            let num_accounts = u16::from_le(*(self.raw as *const u16));
            let data_len = u16::from_le(
                *(self.raw.add(
                    size_of::<u16>()
                        + num_accounts as usize * IntrospectedAccountMeta::SPACE
                        + size_of::<Pubkey>(),
                ) as *const u16),
            );

            core::slice::from_raw_parts(
                self.raw.add(
                    size_of::<u16>()
                        + num_accounts as usize * IntrospectedAccountMeta::SPACE
                        + size_of::<Pubkey>()
                        + size_of::<u16>(),
                ),
                data_len as usize,
            )
        }
    }
}
```

While this approach currently lacks a method to fetch all account_metas in a single call, we believe that this is fine. The user should ideally specify the index of the particular account meta they need, and only retrieve the portion of data they require. For that reason, we introduced another new struct, [IntrospectedAccountMeta](https://github.com/0xRigel/pinocchio/blob/92d36d538043fc26e4dcd557c4ffa5f9e0829d5a/sdk/pinocchio/src/sysvars/instructions.rs#L183), which also uses a raw pointer pattern and provides convenient helpers:

```rust
#[repr(C)]
#[derive(Clone, PartialEq, Eq)]
pub struct IntrospectedAccountMeta {
    pub raw: *const u8,
}

const IS_SIGNER_BIT: u8 = 0; // Assuming bit 0 for signer flag
const IS_WRITABLE_BIT: u8 = 1; // Assuming bit 1 for writable flag

impl IntrospectedAccountMeta {
    const SPACE: usize = 33; 

    pub fn is_writable(&self) -> bool {
        unsafe { (*self.raw & (1 << IS_WRITABLE_BIT)) != 0 }
    }

    pub fn is_signer(&self) -> bool {
        unsafe { (*self.raw & (1 << IS_SIGNER_BIT)) != 0 }
    }

    pub fn key(&self) -> &Pubkey {
        unsafe { &*(self.raw.add(size_of::<u8>()) as *const Pubkey) }
    }

    pub fn to_account_meta(&self) -> AccountMeta {
        AccountMeta::new(self.key(), self.is_signer(), self.is_writable())
    }
}
```

### Result
With this new pattern, performance improvements are significant. Introspecting a single instruction costs only 20 CUs, and introspecting all parts of an instruction with some data and two accounts (including pubkey deserialization) comes in at 100 CUs, as confirmed by our tests in this [repo](https://github.com/L0STE/pinocchio-instructions-sysvar-examples).

Finally, to maintain parity with the original implementation, we introduced the `SanitizeErrors` enum, which can be found [here](https://github.com/0xRigel/pinocchio/blob/feature/instruction-introspection/sdk/pinocchio/src/sanitize_error.rs) 

### Testing
As part of the sysvar-support PR into `mollusk`, @L0STE tested all helper functions by serializing and then fetching a set of mock instructions from the sysvar account.


sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// Load an `Instruction` in the currently executing `Transaction` at the
+/// specified index.
+///
+/// `data` is the instructions sysvar account data.
+///
+/// Unsafe because the sysvar accounts address is not checked; only used
+/// internally after such a check.
+#[inline(always)]
+pub unsafe fn load_instruction_at(index: usize, data: &[u8]) -> IntrospectedInstruction {
```
samuelvanderwaal, line 67: Same thing here: this was private originally and has lots of UB possibilities. Should probably be private or pub(crate) at most and I recommend you document the unsafe behavior better. E.g. there's no bounds checking for the offset table access.

samuelvanderwaal (commented):
This will be a great value add, thanks for taking the time to make the PR!

I am not an expert on unsafe Rust so feel to push back if I'm wrong on any accounts but there appears to be a lot of potential UB here and insufficient explanation about the assumptions that ensure the unsafe parts are actually safe to use.

sdk/pinocchio/src/sysvars/instructions.rs
samuelvanderwaal: Or for example, you're returning a pointer that outlives the function. I believe it's ok because it's pointing to Solana account data which has a static lifetime so is guaranteed to live for the entire instruction call, but maybe that should be explicitly documented?

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// Load the current `Instruction`'s index in the currently executing
+/// `Transaction`.
+///
+/// `data` is the instructions sysvar account data.
+///
+/// Unsafe because the sysvar accounts address is not checked; only used
+/// internally after such a check.
+#[inline(always)]
+pub unsafe fn load_current_index(data: &[u8]) -> u16 {
```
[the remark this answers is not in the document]
deanmlittle: We definitely do not want to make it private. We want to give people the choice to save CUs by not checking in situations where number of instructions is known, or only checking the instruction counter once externally.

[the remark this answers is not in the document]
samuelvanderwaal: Nit: The original code also explicitly uses little endian formatting, I know Solana VM is LE so this is technically correct but seems inconsistent to specify LE in `store_current_index` (and in original code) but not here.

sdk/pinocchio/src/sysvars/instructions.rs
deanmlittle: I think it is also fine. You can't modify the InstructionsSysvar account via CPI, so no need for borrow checking to avoid dangling pointers, and the pointer itself is taken from a reference to a slice with a lifetime, so I don't really see how this could go bad (willing to be proven wrong though!)

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,208 @@
...
+#[inline(always)]
+pub fn load_current_index_checked(
+    instruction_sysvar_account_info: &AccountInfo,
+) -> Result<u16, ProgramError> {
+    if instruction_sysvar_account_info.key() != &INSTRUCTIONS_ID {
+        return Err(ProgramError::UnsupportedSysvar);
+    }
+
+    let instruction_sysvar = instruction_sysvar_account_info.try_borrow_data()?;
+    let index = unsafe { load_current_index(&instruction_sysvar) };
```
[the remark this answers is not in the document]
L0STE: Yep! Afaik the runtime handles the sysvar account this way: 
- Serialize all the instruction in the Transaction using [the following function](https://github.com/anza-xyz/solana-sdk/blob/69edb584ac17df2f5d8b5e817d7afede7877eded/instructions-sysvar/src/lib.rs#L99)
- Then [add](https://github.com/anza-xyz/solana-sdk/blob/69edb584ac17df2f5d8b5e817d7afede7877eded/instructions-sysvar/src/lib.rs#L68) a `u16` to represent the current index
- And run `load_current_index` for the instruction that actually uses the account so it's possible to use the `get_instruction_relative` without any problems

I'll add the `AccountBorrowFailed` as error. You're totally right on that and in general will focus on making some better documentation around the functions

L0STE:
Thanks for your feedback, @samuelvanderwaal!

As you pointed out the Solana VM is Little Endian by default, so there’s no strict need to specify it. However, since using `u16::from_le` doesn’t add any overhead, so I’ve explicitly included it wherever a `u16` is deserialized for added clarity and safety.

I also agree with your point on documentation. I’ve added it wherever possible, especially around the unsafe functions. Our decision to “expose” those private functions via `unsafe fn` stems from the fact that, in most cases, the necessary checks have already been performed (for instance, if multiple instructions are being introspected). This approach lets developers optimize further when they already know the data is valid.

Lastly, I updated the `load_instruction_at` function to include a missing wrapper. Now it checks the Sysvar ID, calls `load_instruction_at_unchecked`, which verifies the index is in range and calls `deserialize_instruction` to create the pointer.

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+    let current_index = load_current_index_checked(&instruction_sysvar_account_info)? as i64;
+    let index = current_index.saturating_add(index_relative_to_current);
+    if index < 0 {
+        return Err(ProgramError::InvalidArgument);
+    }
+
+    let instruction_sysvar_data = instruction_sysvar_account_info.try_borrow_data()?;
+    unsafe {
+        load_instruction_at(index as usize, &instruction_sysvar_data)
+            .map(|instr| instr.clone())
```
samuelvanderwaal, line 168: You don't need to clone here because `load_instruction_at` returns an owned `IntrospectedInstruction` struct, yeah? Small performance improvement.

sdk/pinocchio/src/sysvars/instructions.rs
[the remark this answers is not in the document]
samuelvanderwaal: > We definitely do not want to make it private. We want to give people the choice to save CUs by not checking in situations where number of instructions is known, or only checking the instruction counter once externally.

What actually happens here if the data slice is less than 2 bytes, the usize underflows to have a huge value and points to some area outside the Solana VM's allocated memory and we get a panic? Or would we actually get UB? Seems like there should at least be a line in the docs that warns about always providing a valid data slice here.

That being said I'm not a maintainer on the repo and am aware I'm just drive-by reviewing here, so feel free to do whatever you guys think is best. I just plan to use Pinocchio for stuff in the future so have a vested interest in the API being super solid.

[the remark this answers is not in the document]
deanmlittle: In a hypothetical situation where that's even possible (ie: we borrowed data from the SysvarInstructions account and it somehow isn't populated by the validator), it should panic. The function is unsafe for a reason. We are not in the business of trying to force devs to use training wheels at the expense of CUs. If they know how to handle an unsafe function and read documentation, we've done our part.

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+                ),
+                data_len as usize,
+            )
+        }
+    }
+}
+
+#[repr(C)]
+#[derive(Clone, PartialEq, Eq)]
+pub struct IntrospectedAccountMeta {
```
febo, line 255: Since this type is sized, we can transmute (cast) a pointer to it. This would simplify its logic:
```rust
/// The bit positions for the signer flags in the `AccountMeta`.
const IS_SIGNER: u8 = 0b00000001;

/// The bit positions for the writable flags in the `AccountMeta`.

#[repr(C)]
#[derive(Clone, PartialEq, Eq)]
pub struct IntrospectedAccountMeta {
    /// Account flags:
    ///   * bit `0`: signer
    ///   * bit `1`: writable
    flags: u8,

    /// The account key.
    pub key: Pubkey,
}

impl IntrospectedAccountMeta {
    const LEN: usize = core::mem::size_of::<Self>();

    /// Indicate whether the account is writable or not.
    #[inline(always)]
    pub fn is_writable(&self) -> bool {
        (self.flags & IS_WRITABLE) != 0
    }

    /// Indicate whether the account is a signer or not.
    #[inline(always)]
    pub fn is_signer(&self) -> bool {
        (self.flags & IS_SIGNER) != 0
    }

    /// Convert the `IntrospectedAccountMeta` to an `AccountMeta`.
    pub fn to_account_meta(&self) -> AccountMeta {
        AccountMeta::new(&self.key, self.is_signer(), self.is_writable())
    }
}
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Get the account meta at the specified index.
+    /// 
+    /// # Safety
+    /// 
+    /// This function is unsafe because it does not verify if the index is out of bounds.
+    /// 
+    /// It is typically used internally within the `get_account_meta_at` function, which
+    /// performs the necessary index verification. However, to optimize performance for users 
+    /// who are sure that the index is in bounds, we have exposed it as an unsafe function.
+    pub unsafe fn get_account_meta_at_unchecked(&self, index: usize) -> IntrospectedAccountMeta {
```
febo, line 192: Making the change to `IntrospectedAccountMeta`, we can return a reference to the account directly:
```
 pub unsafe fn get_account_meta_at_unchecked(&self, index: usize) -> &IntrospectedAccountMeta {
    let offset = core::mem::size_of::<u16>() + (index * IntrospectedAccountMeta::LEN);
    &*(self.raw.add(offset) as *const IntrospectedAccountMeta)
}
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    let instruction_sysvar = instruction_sysvar_account_info.try_borrow_data()?;
+    let index = unsafe { load_current_index(&instruction_sysvar) };
+    Ok(index)
+}
+
+/// Store the current `Instruction`'s index in the instructions sysvar data.
+#[inline(always)]
+pub fn store_current_index(data: &mut [u8], instruction_index: u16) {
```
febo, line 59: We don't need this function. It is only used by the validator when setting up the SVM, but this is not the implementation that will be used.

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,287 @@
...
+/// This function is unsafe because it does not verify the address of the sysvar account.
+/// 
+/// It is typically used internally within the `load_current_index_checked` function, which
+/// performs the necessary address verification. However, to optimize performance for users
+/// who have already conducted this check elsewhere in their code, we have exposed it as an
+/// unsafe function rather than making it private, as was done in the original implementation.
+#[inline(always)]
+pub unsafe fn load_current_index(data: &[u8]) -> u16 {
+    let len = data.len();
+    *(data.as_ptr().add(len - 2) as *const u16)
```
febo, line 34: We will need to use `u16::from_le_bytes` here since we cannot guarantee that the alignment will be correct.

febo (commented):
Thanks for the PR, it is looking good. I have a few comments /suggestion– let me know what you think.

I can see that you followed the same API as the existing code, but perhaps we could make some tweaks to bring it more inline to Pinocchio API. In general, the `unsafe` method has a `_unchecked` suffix and the safe one does not have a suffix. That would mean:
- `load_current_index` -> `load_current_index_unchecked`
- `load_current_index_checked` -> `load_current_index`
- `deserialize_instruction` -> `deserialize_instruction_unchecked`
- `load_instruction_at` -> `load_instruction_at_unchecked`
- `load_instruction_at_checked` -> `load_instruction_at`

We could also add other variations:
- `get_instruction_relative_unchcked`: without the "key" validation and the "safe" `try_borrow_data` – so it will be an "unsafe" variant.
- `deserialize_instruction`: with the "key" validation and not "unsafe".

sdk/pinocchio/src/sysvars/instructions.rs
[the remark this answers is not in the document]
samuelvanderwaal: Fair enough; thanks for the explanation!

L0STE:
Solved most of the problem with the implementations and added all the suggestion made, the only thing I will "push-back" on is the safe `deserialize_instruction` function you suggested.  Since that is an internal instruction the "safe" variant is just `load_instruction_at` or `get_instruction_relative`, to make it more "efficient" I moved the index check up one wrapper, so the `deserialize_instruction_unchecked` just create an `IntrospectedInstruction` type.

Additionally I gave a shot to the Phantom Data implementation, don't think it's the prettiest (and maybe even the most efficient way to do so) but initially we were facing the problem that we couldn't specify the lifetime on the `try_borrow_data` of the Sysvar Account. To solve that we implemented a new type `SysvarAccountInfo` that let us specify the lifetimes without much overhead since it uses raw pointer (+ PhantomData to count the borrow)

```rust
#[repr(C)]
#[derive(Clone, PartialEq, Eq)]
pub struct SysvarAccountInfo<'a> {
    pub raw: *const u8,
    pub marker: PhantomData<&'a [u8]>,
}

impl<'a> SysvarAccountInfo<'a> {
    pub fn from_account_info(instruction_sysvar_account_info: &AccountInfo) -> Result<Self, ProgramError> {
        let data = instruction_sysvar_account_info.try_borrow_data()?;
        Ok(Self { raw: data.as_ptr(), marker: PhantomData })
    }

    pub fn try_borrow_data(&self) -> Result<&'a [u8], ProgramError> {
        let data_len = unsafe { *self.raw.sub(8) as *mut u64 };
        let data = unsafe { core::slice::from_raw_parts(self.raw, data_len as usize) };
        Ok(data)
    }
}

#[repr(C)]
#[derive(Clone, PartialEq, Eq)]
pub struct IntrospectedInstruction<'a> {
    pub raw: *const u8,
    pub marker: PhantomData<&'a [u8]>,
}
```

febo:
>     pub fn from_account_info(instruction_sysvar_account_info: &AccountInfo) -> Result<Self, ProgramError> {
>         let data = instruction_sysvar_account_info.try_borrow_data()?;
>         Ok(Self { raw: data.as_ptr(), marker: PhantomData })
>     }

Unfortunately this does not work, since you are keeping the raw pointer after `data` goes out of scope and "drops" the borrow.

Perhaps the way to do it (and maybe improve devex) is to change the `Instructions` struct to hold the reference to the `data` slice. Then we can move all functions to the `Instructions` type:
```rust
pub struct Instructions<T>
where
    T: Deref<Target = [u8]>,
{
    data: T,
}

impl<T> Instructions<T>
where
    T: Deref<Target = [u8]>,
{
    pub unsafe fn new(data: T) -> Self {
        Instructions { data }
    }

    pub unsafe fn load_current_index_unchecked(&self) -> u16 {
        let len = self.data.len();
        u16::from_le_bytes(*(self.data.as_ptr().add(len - 2) as *const [u8; 2]))
    }

    // Other methods that require `data: &[u8]` to be passed in as an argument
}

impl<'a> TryFrom<&'a AccountInfo> for Instructions<Ref<'a, [u8]>> {
    type Error = ProgramError;

    fn try_from(account_info: &'a AccountInfo) -> Result<Self, Self::Error> {
        if account_info.key() != &INSTRUCTIONS_ID {
            return Err(ProgramError::UnsupportedSysvar);
        }

        Ok(Instructions {
            data: account_info.try_borrow_data()?,
        })
    }
}
```

L0STE:
Thanks @febo, this is a much cleaner implementation all around, not just to solve the Phantom Data. 

With this new implementation we don't have the `..._unchecked` functions anymore because the difference between "checked" and "unchecked" functions comes from the way we deserialize the `Instructions` struct (`new` is "unchecked" and therefore unsafe, `try_from` is "checked" and therefore safe). 

For the same reason now all the functions minus `deserialize_instruction` are publicly accessible. 

Note: CUs haven't change really much across the board so this feels like a very clean solution for it

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+    T: Deref<Target = [u8]>,
+{
+    /// Creates a new `Instructions` struct.
+    ///
+    /// `data` is the instructions sysvar account data.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it does not check if the provided data is from the Sysvar Account.
+    pub unsafe fn new(data: T) -> Self {
```
febo, line 35: Let's call this `new_unchecked` to maintain the same pattern for unsafe methods.

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+        Instructions { data }
+    }
+
+    /// Load the current `Instruction`'s index in the currently executing
+    /// `Transaction`.
+    pub fn load_current_index(&self) -> u16 {
+        unsafe {
```
febo, line 42: Let's use the "unsafe" only around the code that is unsafe.
```rust
let len = self.data.len();
/// SAFETY: The last 2 bytes of the Instructions sysvar data represents the current
/// instruction index.
u16::from_le_bytes(unsafe { *(self.data.as_ptr().add(len - 2) as *const [u8; 2]) })
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+where
+    T: Deref<Target = [u8]>,
+{
+    /// Creates a new `Instructions` struct.
+    ///
+    /// `data` is the instructions sysvar account data.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it does not check if the provided data is from the Sysvar Account.
```
febo, line 34: We can annotate this function to be `#[inline(always)]`.

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    /// Creates and returns an `IntrospectedInstruction` for the instruction at the specified index.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it does not check if the provided index is out of bounds. It is
+    /// typically used internally with the `load_instruction_at` or `get_instruction_relative` functions,
+    /// which perform the necessary index verification.
+    pub unsafe fn deserialize_instruction(&self, index: usize) -> IntrospectedInstruction {
+        let offset = u16::from_le(
```
febo, line 56: We can remove the `u16::from_le` here:
```rust
let offset = *(self
    .data
    .as_ptr()
    .add(size_of::<u16>() + index * size_of::<u16>()) as *const u16);
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+
+    /// Creates and returns an `IntrospectedInstruction` for the instruction at the specified index.
+    #[inline(always)]
+    pub fn load_instruction_at(
+        &self,
+        index: usize,
+    ) -> Result<IntrospectedInstruction, ProgramError> {
+        unsafe {
```
febo, line 75: Same here, we can remove the `u16::from_le` and wrap only the "unsafe" code:
```rust
// SAFETY: The first 2 bytes of the Instructions sysvar data represents the
// number of instructions.
let num_instructions = unsafe { *(self.data.as_ptr() as *const u16) };

if index >= num_instructions as usize {
}
// SAFETY: The index was checked to be in bounds.
Ok(unsafe { self.deserialize_instruction(index) })
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /// Creates and returns an `IntrospectedInstruction` relative to the current `Instruction` in the
+    /// currently executing `Transaction.
+    #[inline(always)]
+    pub fn get_instruction_relative(
+        &self,
+        index_relative_to_current: i64,
+    ) -> Result<IntrospectedInstruction, ProgramError> {
+        unsafe {
```
febo, line 93: Same comments for the "unsafe", `u16::to_le` and `ProgramError` things. Also it contains duplicated logic with `load_instruction_at` so we can just call that method instead.
```rust
let current_index = self.load_current_index() as i64;
let index = current_index.saturating_add(index_relative_to_current);

if index < 0 {
}

self.load_instruction_at(index as usize)
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+#[repr(C)]
+#[derive(Clone, PartialEq, Eq)]
+pub struct IntrospectedInstruction<'a> {
+    pub raw: *const u8,
+    pub marker: PhantomData<&'a [u8]>,
+}
+
+impl<'a> IntrospectedInstruction<'a> {
```
febo, line 133: Clippy is complaining that the lifetime can be elided, so we can just write:
```rust
impl IntrospectedInstruction<'_> {
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// Get the account meta at the specified index.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SanitizeError::IndexOutOfBounds`] if the index is out of bounds.
+    pub fn get_account_meta_at(
+        &self,
+        index: usize,
+    ) -> Result<&IntrospectedAccountMeta, SanitizeError> {
+        if index >= unsafe { u16::from_le(*(self.raw as *const u16)) } as usize {
```
febo, line 157: We can't guarantee the alignment here, so we need to use `u16::from_le_bytes`:
```rust
// SAFETY: The first 2 bytes represent the number of accounts in the instruction.
let num_accounts = u16::from_le_bytes(unsafe { *(self.raw as *const [u8; 2]) });

if index >= num_accounts as usize {
}

// SAFETY: The index was checked to be in bounds.
Ok(unsafe { self.get_account_meta_at_unchecked(index) })
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+            return Err(SanitizeError::IndexOutOfBounds);
+        }
+
+        Ok(unsafe { self.get_account_meta_at_unchecked(index) })
+    }
+
+    /// Get the program ID of the `Instruction`.
+    pub fn get_program_id(&self) -> &Pubkey {
+        unsafe {
```
febo, line 166: Same here, we can't guarantee the alignment. Also, let's wrap only the unsafe code in "unsafe":
```rust
// SAFETY: The first 2 bytes represent the number of accounts in the instruction.
let num_accounts = u16::from_le_bytes(unsafe { *(self.raw as *const [u8; 2]) });

unsafe {
        &*(self.raw.add(
                size_of::<u16>() + num_accounts as usize * size_of::<IntrospectedAccountMeta>(),
        ) as *const Pubkey)
}
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,236 @@
...
+            let num_accounts = *(self.raw as *const u16);
+            &*(self.raw.add(
+                size_of::<u16>() + num_accounts as usize * size_of::<IntrospectedAccountMeta>(),
+            ) as *const Pubkey)
+        }
+    }
+
+    /// Get the instruction data of the `Instruction`.
+    pub fn get_instruction_data(&self) -> &[u8] {
+        unsafe {
```
febo, line 176: Let's wrap only the unsafe code with "unsafe". Also we can't guarantee the alignment, so we need to use `u16::from_le_bytes`:
```rust
// SAFETY: The first 2 bytes represent the number of accounts in the instruction.
let offset = u16::from_le_bytes(unsafe { *(self.raw as *const [u8; 2]) }) as usize
    * size_of::<IntrospectedAccountMeta>()
    + PUBKEY_BYTES;

// SAFETY: The instruction data length is located after the program ID.
let data_len = u16::from_le_bytes(unsafe {
    *(self.raw.add(size_of::<u16>() + offset) as *const [u8; 2])
});

// SAFETY: The instruction data is located after the data length.
unsafe {
    core::slice::from_raw_parts(
        self.raw.add(size_of::<u16>() + offset + size_of::<u16>()),
        data_len as usize,
    )
}
```

sdk/pinocchio/src/lib.rs
```diff
@@ -235,6 +235,7 @@ pub mod program {
 }
 pub mod program_error;
 pub mod pubkey;
+pub mod sanitize_error;
```
febo, line 238: There was only one place using this error type, so I think we can just drop it and use `ProgramError` instead.

febo (commented):
Looks great – I have a few more nits to simplify the code, but overall the approach is good.

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    /// Creates and returns an `IntrospectedInstruction` for the instruction at the specified index.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it does not check if the provided index is out of bounds. It is
+    /// typically used internally with the `load_instruction_at` or `get_instruction_relative` functions,
+    /// which perform the necessary index verification.
+    #[inline(always)]
+    pub unsafe fn deserialize_instruction(&self, index: usize) -> IntrospectedInstruction {
```
febo, line 57: nit: could we add the `_unchecked` suffix here since this method is unsafe?

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub unsafe fn get_account_meta_at_unchecked(&self, index: usize) -> &IntrospectedAccountMeta {
+        let offset = core::mem::size_of::<u16>() + (index * IntrospectedAccountMeta::LEN);
+        &*(self.raw.add(offset) as *const IntrospectedAccountMeta)
+    }
+
+    /// Get the account meta at the specified index.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SanitizeError::IndexOutOfBounds`] if the index is out of bounds.
```
febo, line 147: nit: `SanitizeError::IndexOutOfBounds` -> `ProgramError::InvalidArgument`

febo (commented):
Last really small nits and this is ready to go.

febo (approved):
Looks great – thanks!

PR #107, opened 2025-03-23
fix: use u64 prefix string instead of u32 prefix string

sonicfromnewyoke:
the original `system-program` uses `u64` prefix string [idl](https://github.com/solana-program/system/blob/main/program/idl.json#L284)


febo (approved):
Great – thanks!

PR #112, opened 2025-03-27
Use `ProgramError` from Solana SDK

febo:
### Problem

Pinocchio currently re-implements the `ProgramError` type, which is essentially a copy of the same type from Solana SDK. This will make it awkward to write code that interact with both pinocchio and SDK types.

### Solution

Bring the `ProgramError` type from `solana-program-error` crate. This adds a dependency to pinocchio – an "internal" dependency to a SDK crate.


rustopian (dismissed):
Straightforward, looks good

joncinque (dismissed):
Looks great!

joncinque (approved):

PR #115, opened 2025-03-28
publish: Add fetch depth

febo:
### Problem

Currently the changelog generation is failing since it cannot find previous tags.

### Solution

Use the `fetch-depth` parameter on the Git checkout action to retrieve the whole history.


joncinque (approved):
Ah darn, sorry that I missed this

PR #119, opened 2025-04-02
Use `AccountView` from Solana SDK

febo:
### Problem

Currently, pinocchio defines its own (zero-copy) `AccountInfo` struct. The name clash makes it confusing to write code that interacts with both `AccountInfo` types from the SDK and pinocchio.

### Solution

Use the new [`AccountView`](https://github.com/febo/solana-sdk/pull/2) type from Solana SDK instead. While this is a breaking change since the type has a different name, it has the same characteristics of the current `AccountInfo` type, so there is no loss of efficiency.


grod220 (approved):
LGTM

Side note: What's the deal with all of the duplicate program code in this repo? Is it to ensure there aren't breaking changes during core pinocchio development? How do they keep in sync with the other repos (e.g. p-token in solana-program/token)?

febo:
> LGTM
> 
> Side note: What's the deal with all of the duplicate program code in this repo? Is it to ensure there aren't breaking changes during core pinocchio development? How do they keep in sync with the other repos (e.g. p-token in solana-program/token)?

They are CPI clients. For now they are specific to pinocchio, but soon we should move them to their respective program repository. It will be easier once we finish the implementation of https://github.com/codama-idl/renderers-rust-cpi.

programs/system/src/instructions/transfer_with_seed.rs
```diff
@@ -13,22 +13,22 @@ use pinocchio::{
...
 
     /// Recipient account.
-    pub to: &'a AccountInfo,
+    pub to: &'a AccountView,
 
     /// Amount of lamports to transfer.
     pub lamports: u64,
 
-    /// String of ASCII chars, no longer than [`MAX_SEED_LEN`](https://docs.rs/solana-address/latest/solana_address/constant.MAX_SEED_LEN.html).
+    /// String of ASCII chars, no longer than [`solana_address::MAX_SEED_LEN`].
```
joncinque: nit: Was this change intended? I don't think this link will work

sdk/pinocchio/src/lib.rs
```diff
@@ -239,6 +238,10 @@ pub mod sysvars;
 #[deprecated(since = "0.7.0", note = "Use the `entrypoint` module instead")]
 pub use entrypoint::lazy as lazy_entrypoint;
 
+// Re-export the `solana_account_view` for downstream use.
+pub use solana_account_view as account;
```
joncinque, line 242: nit: how about naming the re-export `account_view`?
```suggestion
pub use solana_account_view as account_view;
```

joncinque (commented):
Just a couple of tiny things, looks great!

programs/system/src/instructions/transfer_with_seed.rs
febo: You are right. It works on an IDE since it links to the code of the crate, but not doc.rs – reverted the change.

sdk/pinocchio/src/lib.rs
febo: I was meant to ask about this. At the moment we have:
* `program_error` -> `error`
* `solana_address` -> `address`

Then we could have:
* `account_view` -> `account`
* `instruction_view` -> `instruction`

It is a bit shorter if we "drop" the `_view`, but happy either way.

joncinque: Ah ok, that works! Just wanted to make sure there was a vision

joncinque (approved):
Looks great to me!

joncinque (approved):

PR #120, opened 2025-04-03
log: Use alloc crate

febo:
### Problem

Currently the macro in the `log` package is dependent on `std`, but it does not use anything specific from it.

### Solution

Use `alloc` instead and mark the macro as `no_std`.


joncinque (approved):
Looks great!

PR #121, opened 2025-04-07
add slot hashes sysvar

Ozodimgba:
# Add SlotHashes sysvar support to Pinocchio

This PR adds support for the SlotHashes sysvar to the Pinocchio library, providing on-chain programs with efficient access to historical slot hash data.

## Implementation details

- added a new `slot_hashes.rs` module in the `sysvars` directory
- implemented a `SlotHashesSysvar` struct with methods to efficiently access specific entries using direct syscalls
- used a no_std compatible approach with no dependencies on the standard library
- Added constants for the data layout and sizes to match the actual Solana implementation

## sysvar on chain methods

- `get_entry(index)`: Access a specific slot-hash entry by index
- `get_hash(slot)`: Find the hash for a specific slot
- `position(slot)`: Find the position of a slot in the collection
- `len()`: Get the number of entries available

This implementation allows Pinocchio-based programs to access the SlotHashes sysvar in a way that's compute-efficient and memory-efficient, avoiding the need to deserialize the entire data structure.

question for @febo:

Can I add a std-featured off-chain SlotHashes struct to the implementation? I'd like to include a convenience wrapper that uses Vec for off-chain tooling when the "std" feature is enabled. This would make the API more familiar for off-chain use cases while keeping the core implementation no_std compatible. Would this approach be acceptable for the Pinocchio library's design philosophy?

### DIFF
--- a/sdk/pinocchio/src/sysvars/slot_hashes.rs
+++ b/sdk/pinocchio/src/sysvars/slot_hashes.rs
@@ -2,14 +2,14 @@
 
 use super::Sysvar;
 use crate::{
-    program_error::ProgramError,
-    sysvars::clock::Slot,
+    program_error::ProgramError, 
+    pubkey::Pubkey, 
+    sysvars::clock::Slot
 };
 
 /// The ID of the slot hashes sysvar.
-pub const SLOT_HASHES_ID: [u8; 43] = [83, 121, 115, 118, 97, 114, 83, 49, 111, 116, 72, 97, 115, 
-    104, 101, 115, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 49, 
-    49, 49, 49, 49, 49, 49, 49];
+pub const SLOT_HASHES_ID: Pubkey = [ 6, 167, 213, 23, 25, 47, 10, 175, 200, 117, 226, 225, 132, 
+    87, 124, 80, 105, 207, 200, 70, 73, 227, 235, 146, 120, 47, 149, 141, 72, 0, 0, 0 ];
 
 /// Hash type used to track slot hashes
 pub type Hash = [u8; 32];
--- a/sdk/pinocchio/src/sysvars/slot_hashes.rs
+++ b/sdk/pinocchio/src/sysvars/slot_hashes.rs
@@ -1,0 +1,174 @@
+//! SlotHashes sysvar implementation
+
+use super::Sysvar;
+use crate::{
+    program_error::ProgramError,
+    sysvars::clock::Slot,
+};
+
+/// The ID of the slot hashes sysvar.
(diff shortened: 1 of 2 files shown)


sdk/pinocchio/src/sysvars/slot_hashes.rs
```diff
@@ -0,0 +1,173 @@
...
+
+impl SlotHashesSysvar {
+    /// Returns the number of entries in the slot hashes sysvar.
+    ///
+    /// Returns `None` if the sysvar syscall fails.
+    pub fn len() -> Option<usize> {
+      let mut len_buf = [0u8; VEC_LENGTH_SIZE];
+
+      let result = unsafe {
+          crate::syscalls::sol_get_sysvar(
```
febo, line 58: It is a bit expensive to use a syscall just to get the number of entries. Better to use the syscall to get all the data.

sdk/pinocchio/src/sysvars/slot_hashes.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    /// Gets the hash for the specified slot.
+    ///
+    /// This performs a linear search through the entries.
+    /// Returns `None` if the slot is not found or the sysvar syscall fails.
+    pub fn get_hash(target_slot: Slot) -> Option<Hash> {
+        let len = Self::len()?;
+
+        for i in 0..len {
+            if let Some(entry) = Self::get_entry(i) {
```
febo, line 127: This approach works, but it will be expensive to use multiple syscalls to iterate over them. It will be better to follow a similar approach as [PodSlotHashes](https://github.com/anza-xyz/solana-sdk/blob/0666fa5999750153070e5c43d64813467bfdc38e/sysvar/src/slot_hashes.rs#L106-L145) and [Instructions](https://github.com/anza-xyz/pinocchio/blob/main/sdk/pinocchio/src/sysvars/instructions.rs) sysvar.

sdk/pinocchio/src/sysvars/slot_hashes.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    /// Gets the hash for the specified slot.
+    ///
+    /// This performs a linear search through the entries.
+    /// Returns `None` if the slot is not found or the sysvar syscall fails.
+    pub fn position(target_slot: Slot) -> Option<usize> {
+        let len = Self::len()?;
+
+        for i in 0..len {
+            if let Some(entry) = Self::get_entry(i) {
```
febo, line 147: Same comment here, it will be expensive to iterate using the syscall.

febo:
Thanks for the PR. While this approach will work, it will be expensive to use it – a syscall takes `100` CUs so ideally you will have only one syscall to get all the data. Left some links for you to follow if you decide to refactor this one.

> Can I add a std-featured off-chain SlotHashes struct to the implementation? I'd like to include a convenience wrapper that uses Vec for off-chain tooling when the "std" feature is enabled. 

Off-chain clients can use the SDK crate right? I am not sure what would be the benefit of using this one.

rustopian:
Replaced by #152 - thanks for the initial request & contribution @Ozodimgba

PR #122, opened 2025-04-07
Use `InstructionView` from Solana SDK

febo:
### Problem

Currently, pinocchio defines its own (zero-copy) `Instruction` and `AccountMeta` structs. The name clash makes it confusing to write code that interacts with both Solana SDK and pinocchio.

### Solution

Use the types from the new [`solana-instruction-view`](https://github.com/febo/solana-sdk/pull/3) crate in Solana SDK instead. While this is a breaking change since the type has a different name, it has the same characteristics of the current types, so there is no loss of efficiency.

Note that this change removes the need of having the instruction related code in the main pinocchio crate – instead, client code can import `solana-instruction-view` directly to use the types.


programs/system/src/instructions/assign.rs
joncinque, line 1: Is the idea to refactor the system program helper to not use pinocchio later?

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -206,16 +214,16 @@ impl IntrospectedInstruction<'_> {
...
 
-/// The bit positions for the writable flags in the `AccountMeta`.
+/// The bit positions for the writable flags in the `InstructionAccount`.
 const IS_WRITABLE: u8 = 0b00000010;
 
 #[repr(C)]
 #[cfg_attr(feature = "copy", derive(Copy))]
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub struct IntrospectedAccountMeta {
+pub struct IntrospectedInstructionAccount {
```
joncinque, line 226: Just checking, this change was intended, correct?

joncinque (approved):
Looks good to me! Just two questions that don't stop this from going in

programs/system/src/instructions/assign.rs
febo: Yes, we should bring the improvements from Pinocchio's `Rent` to the SDK and then switch to that implementation.

sdk/pinocchio/src/sysvars/instructions.rs
febo: Yes, but now I wonder whether it is necessary or not. It was mainly to match the naming of `InstructionAccount`, but perhaps it is ok/better to keep it as `IntrospectedAccountMeta`?

joncinque: You're right, it's probably better to match the naming everywhere

PR #125, opened 2025-04-09
add dummy allocator for clippy

publicqi:
similar to #104, if `alloc` is used, we need a dummy `global_allocator` for clippy


febo (approved):
Great, thanks!

PR #127, opened 2025-04-10
[WIP] add Mock Runtime

redmannequin:
Adds a Runtime trait to ensure all runtime implementations expose the same functionality. This is a sample PR to showcase what a possible solution could look like.

The idea here is to provide a mock environment for testing so you don’t need to spin up a local node. Useful for fast iteration, CI, and writing focused unit tests without going full integration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,24 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "autocfg"
-version = "1.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
-
-[[package]]
-name = "bitflags"
-version = "2.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c8214115b7bf84099f1309324e63141d4c5d7cc26862f97a0a857dbefe165bd"
-
-[[package]]
-name = "cfg-if"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


sonicfromnewyoke:
hmmm... interesting feature.

i personally would prefer to test my program against the real node (solana-test-validator) with e2e calls or at least - `solana-program-test` to be sure my program after deployment will behave in the same way I tested it 😅

redmannequin:
Totally agree, e2e testing with a validator is the way to go. This was mostly me just experimenting and thought it’d be a neat idea to try out. I’ll go ahead and close the PR, but I’m glad it got eyes on it. I’ll probably keep the branch around for a bit as a learning tool.

That said, I still think having a trait abstraction layer over the runtime could be useful. It would help draw a clear line between Solana runtime calls and Pinocchio calls, and it’d clean up a lot of the clutter from all the cfg tags(note the `BlackBoxRuntime`)

sonicfromnewyoke:
@redmannequin i like idea itself. This feature can be really handy for someone who is building something irregular. However I still thinking this feature can be a part of [mollusk](https://github.com/anza-xyz/mollusk) like[jest did with mocks](https://jestjs.io/docs/manual-mocks)

redmannequin:
@sonicfromnewyoke Nice shout! I hadn't heard of Mollusk before, just starting to play around with the Solana ecosystem, so I’ll definitely take a deeper dive into it.

Just to add a bit more context on what I was hoping to achieve, a mock runtime that lets me run tests in debug mode and step through the test execution. Hence why I would like that plugable Runtime abstraction hehe, having to go though a VM isn't ideal for this use case.

anyway here's an [example] of a working test using the mock runtime of this branch.

[example]: https://github.com/redmannequin/sol-ez/blob/29134043dbaf5a4014c1556698371e0ed8386b4c/examples/claim/tests/mod.rs

PR #129, opened 2025-04-11
Clean up unnecessary unsafe

febo:
### Problem

There is an unnecessary double "unsafe" in `allocate_unchecked`.

### Solution

Remove the extra "unsafe".


joncinque (approved):
This looks fine to me, but note that we might need to add this back in the future: https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html

febo:
Oh, interesting. Perhaps we should leave it then. What do you think?

joncinque:
How about we split the difference and just put the unsafe block around the last line which requires it?

joncinque (approved):
Looks great and future-proof!

Issue #109, opened 2025-03-26
feat request: add a trait to serialize instruction data


publicqi:
right now all instructions are wrapped behind structs, and instruction data are mostly inside implementation inside `invoke_signed`.

we can add a trait to export/serialize these instructions.

there're two benefits for this:
1. we can easily test again other libraries (e.g. solana-program, anchor, etc.) if our pinocchio instruction has the same results
2. developers may have more control over the instruction data (not common but may be useful for some hacky stuff)

publicqi:
and this could have caught #107

publicqi:
and i just realized for the example i wrote above (`token::transfer`) the discriminator is also wrong (4 bytes). we need to test against the oroginal impl!

febo:
I have been thinking on how to decouple the creation of instruction data and account metas on the clients, so we can add support to both `invoke` and `invoke_unchecked`  without having to duplicate the code.

I think we can avoid using `Box` since it implies having an allocator. We can make a more flexible trait:
```rust
pub trait InstructionData: Deref<Target = [u8]> {}
```
Then from where the "bytes" come from is an implementation detail.

febo:
> and i just realized for the example i wrote above (`token::transfer`) the discriminator is also wrong (4 bytes). we need to test against the original impl!

Not sure what you mean by 4 bytes.

publicqi:
> I have been thinking on how to decouple the creation of instruction data and account metas on the clients, so we can add support to both `invoke` and `invoke_unchecked` without having to duplicate the code.
> 
> I think we can avoid using `Box` since it implies having an allocator. We can make a more flexible trait:
> 
> pub trait InstructionData: Deref<Target = [u8]> {}
> Then where the "bytes" come from is an implementation detail.

this look good

publicqi:
> That one is just 1 byte since it is a BPF program. The native programs have 4 bytes discriminators.

ah right

publicqi:
@febo hey any plan on this one? will this conflict with any of the upcoming updates? if not i can take this one.

> pub trait InstructionData: Deref<Target = [u8]> {}

this is good enough right?

febo:
> [@febo](https://github.com/febo) hey any plan on this one? will this conflict with any of the upcoming updates? if not i can take this one.
> 
> > pub trait InstructionData: Deref<Target = [u8]> {}
> 
> this is good enough right?

I think it looks good. Happy for you to take this one – the changes on the main sdk should not conflict with this.

[...]

PR #131, opened 2025-04-13
Chaz/feat109

Chazzzzzzz:
This implements https://github.com/anza-xyz/pinocchio/issues/109 for Token program

If things look good I will implement this for all other programs

### DIFF
--- a/programs/token/src/instructions/approve.rs
+++ b/programs/token/src/instructions/approve.rs
@@ -8,6 +8,8 @@ use pinocchio::{
 };
 
 use crate::{write_bytes, InstructionData, UNINIT_BYTE};
+extern crate alloc;
+use alloc::boxed::Box;
 
 /// Approves a delegate.
 ///
@@ -55,14 +57,15 @@ impl Approve<'_> {
 }
 
 impl InstructionData for Approve<'_> {
+    #[inline]
     fn get_instruction_data(&self) -> &[u8] {
         // Instruction data layout:
         // -  [0]: instruction discriminator (1 byte, u8)
         // -  [1..9]: amount (8 bytes, u64)
-        let mut instruction_data = [UNINIT_BYTE; 9];
+        let mut instruction_data = Box::new([UNINIT_BYTE; 9]);
 
         // Set discriminator as u8 at offset [0]
-        write_bytes(&mut instruction_data, &[4]);
+        write_bytes(&mut instruction_data.as_mut_slice(), &[4]);
         // Set amount as u64 at offset [1..9]
         write_bytes(&mut instruction_data[1..], &self.amount.to_le_bytes());
 
--- a/programs/token/src/instructions/approve.rs
+++ b/programs/token/src/instructions/approve.rs
@@ -7,7 +7,7 @@ use pinocchio::{
     ProgramResult,
 };
 
-use crate::{write_bytes, UNINIT_BYTE};
+use crate::{write_bytes, InstructionData, UNINIT_BYTE};
 
 /// Approves a delegate.
 ///
@@ -40,7 +40,23 @@ impl Approve<'_> {
             AccountMeta::readonly_signer(self.authority.key()),
         ];
 
-        // Instruction data
+        let instruction = Instruction {
+            program_id: &crate::ID,
+            accounts: &account_metas,
+            data: self.get_instruction_data(),
+        };
+
+        invoke_signed(
+            &instruction,
+            &[self.source, self.delegate, self.authority],
+            signers,
+        )
+    }
+}
+
+impl InstructionData for Approve<'_> {
+    fn get_instruction_data(&self) -> &[u8] {
+        // Instruction data layout:
         // -  [0]: instruction discriminator (1 byte, u8)
         // -  [1..9]: amount (8 bytes, u64)
         let mut instruction_data = [UNINIT_BYTE; 9];
@@ -50,16 +66,6 @@ impl Approve<'_> {
         // Set amount as u64 at offset [1..9]
         write_bytes(&mut instruction_data[1..], &self.amount.to_le_bytes());
 
-        let instruction = Instruction {
-            program_id: &crate::ID,
-            accounts: &account_metas,
-            data: unsafe { from_raw_parts(instruction_data.as_ptr() as _, 9) },
-        };
-
-        invoke_signed(
-            &instruction,
-            &[self.source, self.delegate, self.authority],
-            signers,
-        )
+        unsafe { from_raw_parts(instruction_data.as_ptr() as _, 9) }
     }
 }
--- a/programs/token/src/lib.rs
+++ b/programs/token/src/lib.rs
@@ -15,3 +15,7 @@ fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
         d.write(*s);
     }
 }
+
+pub trait InstructionData {
+    fn get_instruction_data(&self) -> &[u8];
+}
--- a/programs/token/src/instructions/approve_checked.rs
+++ b/programs/token/src/instructions/approve_checked.rs
@@ -8,6 +8,8 @@ use pinocchio::{
 };
 
 use crate::{write_bytes, InstructionData, UNINIT_BYTE};
+extern crate alloc;
+use alloc::boxed::Box;
 
 /// Approves a delegate.
 ///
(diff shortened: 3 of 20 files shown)


programs/token/src/instructions/approve.rs
```diff
@@ -63,3 +55,20 @@ impl Approve<'_> {
...
     }
 }
+
+impl InstructionData for Approve<'_> {
+    #[inline]
+    fn get_instruction_data(&self) -> &[u8] {
+        // Instruction data layout:
+        // -  [0]: instruction discriminator (1 byte, u8)
+        // -  [1..9]: amount (8 bytes, u64)
+        let mut instruction_data = Box::new([UNINIT_BYTE; 9]);
```
febo, line 65: We need to avoid using a `Box` here since it will be expensive. You can use: `core::mem::MaybeUninit<[u8; 9]>`.

programs/token/src/instructions/approve.rs
```diff
@@ -63,3 +55,20 @@ impl Approve<'_> {
...
+
+        // Set discriminator as u8 at offset [0]
+        write_bytes(&mut instruction_data.as_mut_slice(), &[4]);
+        // Set amount as u64 at offset [1..9]
+        write_bytes(&mut instruction_data[1..], &self.amount.to_le_bytes());
+
+        unsafe { from_raw_parts(instruction_data.as_ptr() as _, 9) }
```
febo, line 72: The here you can have `unsafe { instruction_data.assume_init() }`.

programs/token/src/lib.rs
```diff
@@ -15,3 +15,7 @@ fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
         d.write(*s);
     }
 }
+
+pub trait InstructionData {
+    fn get_instruction_data(&self) -> &[u8];
```
febo, line 20: Maybe we can name this `get`.

programs/token/src/instructions/approve.rs
publicqi: there will be a lifetime issue using stack alone. one way to do this is by specifying this function `#[inline(always)]` so that the variable will always be on the same stack with the caller. but this is too hacky wdyt?

publicqi: assuming using the bump allocator, allocating a fixed size may only increment the free pointer, like 2 ebpf instructions?

febo: Based on some quick tests that I did, using a `Box` is a bit more expensive (#126) than just the pointer increment.

febo: We can use a similar approach than the `ReturnData`: https://github.com/anza-xyz/pinocchio/blob/main/sdk/pinocchio/src/cpi.rs#L316-L325

publicqi: > assuming using the bump allocator, allocating a fixed size may only increment the free pointer, like 2 ebpf instructions?

this is interesting. one overhead in the example would be zero initialization. i cannot think of any other operation a simple alloc would do. can verify this later.

Chazzzzzzz: > We can use a similar approach than the `ReturnData`: https://github.com/anza-xyz/pinocchio/blob/main/sdk/pinocchio/src/cpi.rs#L316-L325

In the example, seems only get_return_data generates the data, are you suggesting that we do not parse pointer around, instead just return [u8] back?

programs/token/src/lib.rs
stegaBOB: I think this could be a bit confusing with just `get` alone. A separate idea could be perhaps using the AsRef trait instead?

illia-bobyr, line 20: I might be missing some context.
But this API seems inconsistent between the definition and the implementation/usage.
And as a result, at the usage location, there are `unsafe` blocks that introduce undefined behavior, as they have "use after free" issues.

As written, this function says that the returned slice has the same lifetime as `self`.
Which normally[^1] means that the returned reference points to data inside `self`.
And, the compiler will use this to enforce the lifetime of `self`.
In particular, it would be required for `self` to live at least as long as the returned reference.

[^1]: I'm not sure if it can mean anything else at all.  But I could be missing some cases.

Yet, in the current implementation, all the `get_instruction_data()` methods return reference to freed memory.  For example:

```rust
impl InstructionData for Transfer<'_> {
    #[inline]
    fn get_instruction_data(&self) -> &[u8] {
        // Instruction data layout:
        // -  [0]: instruction discriminator (1 byte, u8)
        // -  [1..9]: amount (8 bytes, u64)
        let mut instruction_data = Box::new([UNINIT_BYTE; 9]);
        // Set discriminator as u8 at offset [0]
        write_bytes(&mut instruction_data.as_mut_slice(), &[3]);
        // Set amount as u64 at offset [1..9]
        write_bytes(&mut instruction_data[1..9], &self.amount.to_le_bytes());
        unsafe { from_raw_parts(instruction_data.as_ptr() as _, 9) }
    }
}
```

`instruction_data` points to a dynamically allocated block of memory, which will be freed at the end of the function execution.
`&[u8]` constructed in the last expression will point to freed memory.

This may work, as Solana programs currently uses bump allocator for the heap, and it never actually allocates new objects on top of the freed memory.  But it is still an undefined behavior.

In general, I would recommend against using `unsafe` if possible.
While in the existing code `unsafe` is used to avoid zero initialization, in the new code there is no explicit justification as to why a new `unsafe` block is necessary.
Ideally, there should be a good reason to disable provided safety mechanisms.

Note that the existing code is correct, because a reference in the `data` field of an `Instruction` instance is alive longer than the lifetime of that `Instruction` instance:

```rust
        // Instruction data layout:
        // -  [0]: instruction discriminator (1 byte, u8)
        // -  [1..9]: amount (8 bytes, u64)
        let mut instruction_data = [UNINIT_BYTE; 9];

        // Set discriminator as u8 at offset [0]
        write_bytes(&mut instruction_data, &[3]);
        // Set amount as u64 at offset [1..9]
        write_bytes(&mut instruction_data[1..9], &self.amount.to_le_bytes());

        // ### Instruction instance is constructed here.
            program_id: &crate::ID,
            accounts: &account_metas,
            data: unsafe { from_raw_parts(instruction_data.as_ptr() as _, 9) },
        };

        invoke_signed(&instruction, &[self.from, self.to, self.authority], signers)
        // ### Instruction instance is destroyed here.
        // ### Array pointed to by `instruction_data` is freed here, when the function stack is unrolled.
    }
```

Depending on what are you trying to do, one possible way to express your intent might be to return a `Vec<u8>` or a `Box<[u8]>`.
This would correctly transfer the ownership of the allocated buffer to the caller.
It might be possible to optimize memory allocation to avoid too much work.
But you would need to do a dynamic allocation, and, I do not know if the compiler is smart enough to avoid it.  As the dynamic allocation is a call to the allocator, which is a side effect.

Another option would be to actually store the buffer that holds the data inside the `self` object.
This way, this interface becomes valid.
And you can even do stack allocation in this case.

Right now, `Transfer` holds "parsed" representation of the call.  One approach is to introduce an intermediate step between the "parsed" representation and the `invoke()`/`invoke_signed()` calls.
You can call it `TransferPacked`, or `TransferSerialized`, or, perhaps, rename the current `Tranfer` to `TransferBuilder`, and use `Transfer` as the name of the serialized representation.

Now, `TransferSerialized` holds an array of bytes that are the encoded instruction.
And `invoke()`/`invoke_signed()` just used these bytes.

The advantage is that it is up to the caller to decide where `TransferSerialized` is allocated.  And as it is a specific type, the compiler knows the size as well, so you can easily allocate it on the stack.

Something like this:

```rust
pub struct Transfer<'a> {
    /// Sender account.
    pub from: &'a AccountInfo,
    /// Recipient account.
    pub to: &'a AccountInfo,
    /// Authority account.
    pub authority: &'a AccountInfo,
    /// Amount of microtokens to transfer.
    pub amount: u64,
}

impl Transfer<'a> {
    pub fn serialize(self) -> TransferSerialized<'a> {
        self.into()
    }
}

pub struct TransferSerialized<'a> {
    from: &'a AccountInfo,
    to: &'a AccountInfo,
    authority: &'a AccountInfo,
    data: [u8; 9];
}

impl<'a> TransferSerialized<'a> {
    pub data(&self) -> &[u8; 9] {
        &self.data
    }

    #[inline(always)]
    pub fn invoke(&self) -> ProgramResult {
        self.invoke_signed(&[])
    }

    pub fn invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
        let Self {
            from,
            to,
            authority,
            data,
        } = self;

        let accounts: [AccountMeta; 3] = [
            AccountMeta::writable(from.key()),
            AccountMeta::writable(to.key()),
            AccountMeta::readonly_signer(authority.key()),
        ];

            program_id: &crate::ID,
            accounts,
            data,
        };

        invoke_signed(&instruction, &[from, to, authority], signers)
    }
}

impl<'a> From<Transfer<'a>> for TransferSerialized<'a> {
    fn from(
        Transfer {
            from,
            to,
            authority,
            amount,
        }: Transfer,
    ) -> Self {
        // Instruction data layout:
        // -  [0]: instruction discriminator (1 byte, u8)
        // -  [1..9]: amount (8 bytes, u64)
        let mut data = [UNINIT_BYTE; 9];

        // Set discriminator as u8 at offset [0]
        write_bytes(&mut data, &[3]);
        // Set amount as u64 at offset [1..9]
        write_bytes(&mut data[1..9], &amount.to_le_bytes());

        // TODO: Replace with `array_assume_init()` when it is stabilized:
        // https://github.com/rust-lang/rust/issues/96097
        //
        // SAFETY: We have initialized all the bytes above.
        let data = unsafe { mem::transmute::<_, [u8; 9]>(data) };

        Self {
            from,
            to,
            authority,
            data,
        }
    }
}
```

You can also compute the `accounts` value and store it in the `TransferSerialiezd` instance.
But it is unclear to me, to be honest, why do you want to expose the `data` field in the first place.
Depending on it, it may or may not make sense to store the `accounts` value.

PR #138, opened 2025-04-19
fix: file extensions in package.json scripts

sonicfromnewyoke:
changed `.mjs` -> `.mts` extensions in the `scripts` section of the root `package.json`.

everything works in the same way as before.
just a small niceness


febo (approved):
Thank you!

PR #139, opened 2025-04-21
added pub to TokenAccount struct fields

dotslashapaar:
error fixed "field is private" for TokenAccount by adding pub to change the visibility of TokenAccount.mint and all the other fields of TokenAccount struct

this is the error
Screenshot 2025-04-21 082718

### DIFF
--- a/programs/token/src/state/token.rs
+++ b/programs/token/src/state/token.rs
@@ -11,41 +11,41 @@ use crate::ID;
 #[repr(C)]
 pub struct TokenAccount {
     /// The mint associated with this account
-    mint: Pubkey,
+    pub mint: Pubkey,
 
     /// The owner of this account.
-    owner: Pubkey,
+    pub owner: Pubkey,
 
     /// The amount of tokens this account holds.
-    amount: [u8; 8],
+    pub amount: [u8; 8],
 
     /// Indicates whether the delegate is present or not.
-    delegate_flag: [u8; 4],
+    pub delegate_flag: [u8; 4],
 
     /// If `delegate` is `Some` then `delegated_amount` represents
     /// the amount authorized by the delegate.
-    delegate: Pubkey,
+    pub delegate: Pubkey,
 
     /// The account's state.
-    state: u8,
+    pub state: u8,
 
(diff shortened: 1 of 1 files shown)


publicqi:
`TokenAccount` is not really for manual construction and using `from_account_info` and `from_bytes` were the expected constructors. what circumstance do you need them to be pub?

dotslashapaar:
Oh, so I was attempting to create a token account for the signer in the program. Are you saying manual construction isn't suitable for this purpose?}

i was trying to do this
```
TokenAccount{       
        mint: *collection_mint.key(),
};
```

publicqi:
if you want to "create" a token account, you should [initialize account](https://github.com/anza-xyz/pinocchio/blob/main/programs/token/src/instructions/initialize_account.rs)

if you want to deserialize an existing token account (i.e. already created/inited), use `from_account_info` with the corresponding account info.

the only scenario making the fields pub i can think of is testing. but keeping them private encourages people not to directly modify them as they should not.

dotslashapaar:
Ohh okay, I can use `InitializeAccount` to create the token account—got it. Thanks a lot for the help!

publicqi:
no problem

sonicfromnewyoke:
@dotslashapaar another one valid solution, IMO, will be using something like `builder pattern` or explicit setters, like it's done in [p-token](https://github.com/solana-program/token/blob/main/interface/src/state/account.rs#L50)

to not blow-up the resulting bin size for everybody, you can add a `feature`, something like `builder`, in `Cargo.toml`

also, if you need packing your struct into bytes, you can add the next packing (not sure this fn will compile, but direction you have to get)
```rust
 #[inline(always)]
 pub unsafe fn into_bytes(&self) -> &[u8] {
    core::slice::from_raw_parts((self as *const Account) as *const u8, Self::LEN)
 }
 ```

dotslashapaar:
@sonicfromnewyoke  could you please share a repo example where p-token is being used in pinocchio? Would help a lot!

dotslashapaar:
lmao, what did I even just ask pretty sure I need some sleep now 😅

Issue #136, opened 2025-04-16
improvements on allocators


publicqi:
## background
as described in #126, svm runtime gives the user heap to use anyways with or without defining a heap allocator. 

default `BumpAllocator` will store two `usize` on bss section, one indicating the free pointer and the other max len.

`NoAllocator` now does not allow `alloc` generated heap allocation, and #126 adds a manual allocation for `NoAllocator`.

## a few thoughts
1. `NoAllocator`'s name is confusing. it is able to let user to claim heap memory, but it cannot let `alloc` crate to get memory.
2. `NoAllocator` is very dev-unfriendly (maybe we can introduce some static check to avoid overlap?)
3. using allocating functions in crate `alloc` like `Box::new` will introduces some rust runtime check, like null ptr check. subsequently it emits panic stuff like in #85 .
4. `allocate_unchecked` needs the annoying `#![cfg_attr(target_os = "solana", feature(const_mut_refs))]`

some easy todo improvements:
- [ ] remove `len` field of `BumpAllocator`
- [ ] remove `const` for `allocate_unchecked` (or we wait for toolchain updating to [1.83.0](https://github.com/anza-xyz/rust/blob/89fb68fd3298a1c6998337b891f975eb60b8a4e4/compiler/rustc_feature/src/accepted.rs#L147))
- [ ] add same interface (`allocate_unchecked`) for `BumpAllocator`
- [ ] introduce [static_assertions](https://docs.rs/static_assertions/latest/static_assertions/) for `allocate_unchecked` to through compile-time error for assertion (e.g. heap oob)

some challenging ones:
- [ ] compile time overlap check (not sure if it's possible without touching the compiler)
- [ ] bypass `alloc` runtime check (i.e. lightweight panic if allocation fails). i tried but failed. compiler change needed for this one

publicqi:
@febo let me know your thoughts on the easy-to-do ones.

febo:
Great stuff!

I am actually working on something that will hopefully improve the combination of static allocations + bump allocator. It will address some of your points. I should have a draft PR very soon.

My idea is to avoid having todo the compile time overlap check, although it will have trade-offs since it means all allocations has to be knows at compile time. E.g., you won't be able to have two different execution paths that allocate all available memory – probably won't be a problem.

Not sure about always returning `&'static mut MaybeUninit<T>`. Currently you have the option to allocated a `MaybeUninit<T>`, which will give you a `&'static mut MaybeUninit<T>` back.

[...]

PR #140, opened 2025-04-22
Improve allocators set up

febo:
### Problem

Currently, the "fallback" panic handler and allocator set up creates different problems when using `cargo test` or `cargo test-sbf`. This happens since tests in general depend on the `std` and therefore create a conflict when the fallback tries to set up a panic handler or global allocator.

### Solution

This PR improves the set up of the fallbacks by simply linking the `std` library instead of trying to set them up, avoiding conflicts if another library links the `std`. Note that the `std` is only linked in targets different than `"solana"`, so `no_std` programs are not affected.

It also addresses a couple of comments from #136:
* deprecate `entrypoint` macro in favour of `entrypoint_with_allocator_and_panic_handler`
* deprecate `no_allocator` macro in favour of `static_allocator`
* remove the need for `feature(const_mut_refs)`

cc: @publicqi

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -33,7 +33,7 @@ pub const SUCCESS: u64 = super::SUCCESS;
 /// Declare the program entrypoint and set up global handlers.
 ///
 /// This macro is deprecated and will be removed in a future version. It is recommended to
-/// use the [`entrypoint_with_allocator_and_panic_handler`] macro instead.
+/// use the [`crate::entrypoint_with_allocator_and_panic_handler`] macro instead.
 #[deprecated(
     since = "0.8.4",
     note = "Use the `entrypoint_with_allocator_and_panic_handler` macro instead"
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -89,6 +89,18 @@ pub const SUCCESS: u64 = super::SUCCESS;
...
+/// linker or not. The `entrypoint!` macro will set up a default panic "hook", that works with the
+/// `#[panic_handler]` set by the `std`. Therefore, this macro should be used when the program or any
+/// of its dependencies are dependent on the `std` library.
+///
+/// When the program and all its dependencies are `no_std`, it is necessary to set a
+/// `#[panic_handler]` to handle panics. This is done by the `nostd_panic_handler!` macro. In this case,
+/// it is not possible to use the `entrypoint` macro. Use the `program_entrypoint!` macro instead and set
+/// up the allocator and panic handler manually.
 #[macro_export]
 macro_rules! entrypoint {
```
joncinque, line 105: nit: given the information, I almost wonder if we should rename this macro to `std_entrypoint` or `entrypoint_with_std` or `entrypoint_with_allocator_and_panic_handler` to be totally clear

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -337,12 +357,30 @@ macro_rules! no_allocator {
...
+    () => {
+        $crate::static_allocator!();
+    };
+}
+
+/// A global allocator that does not dynamically allocate memory.
+///
+/// This macro sets up a global allocator that denies all dynamic allocations, while
+/// allowing static ("manual") allocations. This is useful when the program does not need to
+/// synamically allocate memory and manages their own allocations.
```
joncinque, line 376:
```suggestion
/// dynamically allocate memory and manages their own allocations.
```

joncinque (commented):
Looks great! I'm a big fan of the lengthy and clear comments.

Just a tiny nit

sdk/pinocchio/src/entrypoint/mod.rs
febo: Agree, so we could deprecate the `entrypoint` and add an `entrypoint_with_allocator_and_panic_handler` which would do the same.

joncinque (approved):

PR #141, opened 2025-04-22
Add `const_allocator` macro

febo:
### Problem

PR #126 added the ability to statically allocate memory on the heap reserved space. As noted in #136, there are a few caveats of using it:
* developers need to manually determine the memory offset for the allocation
* there is no compile time check to avoid memory overlaps if a wrong offset is used
* dynamic memory allocation is disabled

### Solution

This PR adds a `const_allocator` macro that combines the static allocation with the existing `BumpAllocator`. In summary, it calculates the offsets for the static allocations and then sets up the `BumpAllocator` with the remaining heap space.
```rust
pinocchio::const_allocator!({
    buffer: [u8; 5000],
    counter: u64,
    some_other_allocation: u32,
});
```

The allocations are available to the program under an `alloc` module generated by the macro:
```rust
let buffer: &mut [u8; 5000] = crate::alloc::buffer();
let counter: &mut u64 = crate::alloc::counter();
let some_other_allocation: &mut u32 = crate::alloc::some_other_allocation();
```

Programs can use the remaining heap space with the "standard" global allocator. The following code will work when using the `const_allocator`:
```rust
let mut dynamic_buffer: Vec<u8> = Vec::with_capacity(5000);
```

If the static allocation exceeds the available heap space, a compile time error is generated:
```
could not evaluate static initializer
the evaluated program panicked at 'Heap allocation overflow: The size of the
allocated memory exceeds the available heap space.'
```

The `const_allocator` is complementary to the `static_allocator` macro. The later gives all the control to the developer, while `const_allocator` provides the flexibility of static and dynamic allocations.

cc: @publicqi

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -569,7 +569,7 @@ macro_rules! const_allocator {
                         // does not exceed the heap size.
                         __len: if len > ($crate::entrypoint::HEAP_LENGTH as usize) {
                             panic!("Heap allocation overflow: \
-                                The size of the struct exceeds the available heap space.");
+                                The size of the allocated memory exceeds the available heap space.");
                             } else {
                                 $crate::entrypoint::HEAP_LENGTH - len
                             },
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -521,3 +521,94 @@ unsafe impl core::alloc::GlobalAlloc for NoAllocator {
         // I deny all allocations, so I don't need to free.
     }
 }
+
+/// A macro to define a constant allocator.
+///
+/// This macro defines a static memory allocator for the program and sets up a bump
+/// allocator with the remaining heap space. It takes a list "pointer" name and type
+/// pairs, allocates them in the heap, and provides a function to access them.
+#[macro_export]
+macro_rules! const_allocator {
+    ( { $( $ptr_name:ident : $ptr_ty:ty ),+ $(,)? } ) => {
+        pub(crate) mod alloc {
+            //! Static memory allocations for the program.
+
+            /// The `Allocator` struct defines the static memory locations
+            /// for the program and the remaining heap size.
+            struct Allocator {
+                __start: usize,
+                __len: usize,
+                $(
+                    $ptr_name : usize
+                ),*
+            }
+
+            impl Allocator {
+                #[allow(clippy::arithmetic_side_effects)]
+                const fn new() -> Self {
+                    let mut start: usize = $crate::entrypoint::HEAP_START_ADDRESS as usize;
+
+                    $(
+                        let align = core::mem::align_of::<$ptr_ty>();
+                        // Align the start address to the next multiple of `align`.
+                        let boundary = (start + align - 1) & !(align - 1);
+                        let $ptr_name = boundary;
+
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -521,3 +521,94 @@ unsafe impl core::alloc::GlobalAlloc for NoAllocator {
...
+                    Self {
+                        __start: start,
+                        // A compile-time check to ensure that the lenght of the allocated memory
+                        // does not exceed the heap size.
+                        __len: if len > ($crate::entrypoint::HEAP_LENGTH as usize) {
+                            panic!("Heap allocation overflow: \
+                                The size of the allocated memory exceeds the available heap space.");
+                            } else {
+                                $crate::entrypoint::HEAP_LENGTH - len
+                            },
```
joncinque, line 575: nit: since it's possible to request a bigger heap size, this isn't necessarily a problem. Maybe this can be tuned in future work though, and we can leave it as is

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -521,3 +521,94 @@ unsafe impl core::alloc::GlobalAlloc for NoAllocator {
...
+                }
+            }
+
+            // A static instance of the allocator.
+            const ALLOCATOR: Allocator = Allocator::new();
+
+            $(
+                // Make this `const` once `const_mut_refs` is stable for the platform-tools
+                // toolchain Rust version.
+                #[doc = concat!(" Returns a mutable reference to `", stringify!($ptr_name), "`.")]
```
joncinque, line 589: nit extra space
```suggestion
                #[doc = concat!("Returns a mutable reference to `", stringify!($ptr_name), "`.")]
```

joncinque (commented):
Looks really good! As discussed, we should probably move this to a different crate, and please add the full example from the PR description into a doc comment so people can understand how to use this more easily

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -521,3 +521,94 @@ unsafe impl core::alloc::GlobalAlloc for NoAllocator {
...
+
+/// A macro to define a constant allocator.
+///
+/// This macro defines a static memory allocator for the program and sets up a bump
+/// allocator with the remaining heap space. It takes a list "pointer" name and type
+/// pairs, allocates them in the heap, and provides a function to access them.
+#[macro_export]
+macro_rules! const_allocator {
+    ( { $( $ptr_name:ident : $ptr_ty:ty ),+ $(,)? } ) => {
+        pub(crate) mod alloc {
```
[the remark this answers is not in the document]
febo: Good point!

sdk/pinocchio/src/entrypoint/mod.rs
febo: Let's make the change to the bump allocator first and then we can have this one without this limit.

sdk/pinocchio/src/entrypoint/mod.rs
febo: I think the extra space is on purpose, so the generated doc string looks like:
```rust
/// Returns a mutable to ...
   ^--- extra space
```
Having said that, I don't think anyone will look at the generated source code so I will remove it. 😅

febo:
The refactoring for the `const_allocator` will be added to a separate crate.

PR #144, opened 2025-04-24
Fix incorrect source pointer type in `sol_memmove`

gtg2619:
This PR corrects the type of the `src` parameter in the `sol_memmove` function.

Changed src from *mut u8 to *const u8 to align with the semantics of the standard memmove function, where the source is a read-only pointer.

### DIFF
--- a/sdk/pinocchio/src/memory.rs
+++ b/sdk/pinocchio/src/memory.rs
@@ -61,7 +61,7 @@ pub unsafe fn sol_memcpy(dst: &mut [u8], src: &[u8], n: usize) {
 ///
 /// [`ptr::copy`]: https://doc.rust-lang.org/std/ptr/fn.copy.html
 #[inline]
-pub unsafe fn sol_memmove(dst: *mut u8, src: *mut u8, n: usize) {
+pub unsafe fn sol_memmove(dst: *mut u8, src: *const u8, n: usize) {
     #[cfg(target_os = "solana")]
     syscalls::sol_memmove_(dst, src, n as u64);
 


sdk/pinocchio/src/memory.rs
```diff
@@ -61,7 +61,7 @@ pub unsafe fn sol_memcpy(dst: &mut [u8], src: &[u8], n: usize) {
 ///
 /// [`ptr::copy`]: https://doc.rust-lang.org/std/ptr/fn.copy.html
 #[inline]
-pub unsafe fn sol_memmove(dst: *mut u8, src: *mut u8, n: usize) {
+pub unsafe fn sol_memmove(dst: *mut u8, src: *const u8, n: usize) {
```
febo, line 64: Should we change the parameters to `dst: &mut [u8], src: &[u8]` so it matches all the functions? This is the only one requiring raw pointers.

gtg2619: Considering the correspondence with the [memmove in cpp](https://en.cppreference.com/w/c/string/byte/memmove), I don't think this is reasonable. And the goal of saving CU, non-slice copy methods should be retained

gtg2619: Oh, I just didn't notice that. It looks like it should be changed as you mentioned to match other functions.

LucasSte: > Should we change the parameters to dst: &mut [u8], src: &[u8] so it matches all the functions? 

Just a reminder that a slice represents a reference to a struct that contains the pointer to the inner data and the length of the slice. When calling the syscall, you still need the pointer to the inner data, not to the slice.

sdk/pinocchio/src/memory.rs
```diff
@@ -61,7 +61,7 @@ pub unsafe fn sol_memcpy(dst: &mut [u8], src: &[u8], n: usize) {
 ///
 /// [`ptr::copy`]: https://doc.rust-lang.org/std/ptr/fn.copy.html
 #[inline]
-pub unsafe fn sol_memmove(dst: *mut u8, src: *mut u8, n: usize) {
+pub unsafe fn sol_memmove(dst: &mut [u8], src: &[u8], n: usize) {
     #[cfg(target_os = "solana")]
     syscalls::sol_memmove_(dst, src, n as u64);
```
LucasSte, line 66: If the slice is the way to go, you need dst.as_mut_ptr() and src.as_ptr() here.

sdk/pinocchio/src/memory.rs
LucasSte: The compiler implementation also utilizes const: https://github.com/anza-xyz/compiler-builtins/blob/759adff89713678bfd9c7630fe0f60088f635085/src/mem/mod.rs#L205

ilya-bobyr:
Moving from pointers to slices changes the semantics.
With pointers, `src` and `dst` can overlap.
But with slices they can not - it would be undefined behavior in Rust.
Also, as slices already contain the length, there seems to be no reason to pass `n` as yet another argument.
There are already two length values.

The [doc-comment for `sol_memmove`](https://github.com/anza-xyz/pinocchio/blob/9741bbf5ded8b91c8c79a7dfd122df3b7219311d/sdk/pinocchio/src/memory.rs#L93-L99) says:

```rust
/// # Safety
///
/// The same safety rules apply as in [`ptr::copy`].
///
/// [`ptr::copy`]: https://doc.rust-lang.org/std/ptr/fn.copy.html
```

And [`std::ptr::copy` docs](https://doc.rust-lang.org/std/ptr/fn.copy.html) explicitly says:

> Copies `count * size_of::<T>()` bytes from `src` to `dst`. The source and destination may overlap.

I do not know if the fact that we should use `std::ptr::copy` safety rules, also means we should be using the overlapping rules.

sdk/pinocchio/src/memory.rs
gtg2619: My mistake. I think it should work now

febo:
> Moving from pointers to slices changes the semantics.

This is a good point. With the change to slice, this effectively becomes the same as `sol_memcpy` since it does not allow overlaps.

@gtg2619 Sorry, my suggestion made was not good. I think we should revert to use raw pointer.

PR #147, opened 2025-04-30
Account borrow state check

febo:
### Problem

The account borrow check done on CPIs is slightly more expensive than it needs to be, since it also checks whether another borrow can be made or not.

### Solution

Create a `BorrowState` enum to represent different borrow masks. This allows to perform checks on the account borrow state. This will enable CPI helpers that already set the accounts in the correct order to use `*_unchecked` variants, but still perform borrow checks.

It also makes 2 other improvements related to CPIs:
* inlines the `invoke_signed_unchecked`
* uses a `MaybeUnit` for the pubkey in the `get_return_data`

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -15,6 +15,26 @@ use crate::{program_error::ProgramError, pubkey::Pubkey, ProgramResult};
 /// single top-level instruction.
 pub const MAX_PERMITTED_DATA_INCREASE: usize = 1_024 * 10;
 
+/// Represents masks for borrow state of an account.
+#[repr(u8)]
+#[derive(Clone, Copy)]
+pub enum BorrowState {
+    /// Mask to check whether an account is already borrowed.
+    ///
+    /// This will test both data and lamports borrow state.
+    Borrowed = 0b_1111_1111,
+
+    /// Mask to check whether an account is already mutably borrowed.
+    ///
+    /// This will test both data and lamports mutable borrow state.
+    MutablyBorrowed = 0b_1000_1000,
+
+    /// Mask to check whether an account is already immutably borrowed.
+    ///
+    /// This will test both data and lamports immutably borrow state.
+    ImmutablyBorrowed = 0b_0111_0111,
+}
+
 /// Raw account data.
 ///
 /// This data is wrapped in an `AccountInfo` struct, which provides safe access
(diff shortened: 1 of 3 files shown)


sdk/pinocchio/src/account_info.rs
```diff
@@ -172,6 +192,15 @@ impl AccountInfo {
         core::ptr::write_volatile(&(*self.raw).owner as *const _ as *mut Pubkey, *new_owner);
     }
 
+    /// Return true if the account borrow state is set to the given state.
+    ///
+    /// This will test both data and lamports borrow state.
+    #[inline(always)]
+    pub fn check_borrowed_state(&self, state: BorrowState) -> bool {
```
joncinque, line 199: nit: up to you, but `is_borrowed_with_state` or `is_borrowed_for_state` or just `is_borrowed` might be a clearer name

sdk/pinocchio/src/account_info.rs
```diff
@@ -452,7 +513,9 @@ impl AccountInfo {
     pub fn close(&self) -> ProgramResult {
         // make sure the account is not borrowed since we are about to
         // resize the data to zero
-        self.check_borrow_mut_data()?;
+        if self.check_borrowed_state(BorrowState::Borrowed) {
+            return Err(ProgramError::AccountBorrowFailed);
+        }
```
joncinque, line 518: I see the bugfix here :wink:

sdk/pinocchio/src/account_info.rs
```diff
@@ -15,6 +15,26 @@ use crate::{program_error::ProgramError, pubkey::Pubkey, ProgramResult};
...
+
+    /// Mask to check whether an account is already mutably borrowed.
+    ///
+    /// This will test both data and lamports mutable borrow state.
+    MutablyBorrowed = 0b_1000_1000,
+
+    /// Mask to check whether an account is already immutably borrowed.
+    ///
+    /// This will test both data and lamports immutably borrow state.
+    ImmutablyBorrowed = 0b_0111_0111,
```
joncinque, line 35: Is this used at all? If not, probably better to just remove it

joncinque (commented):
This looks really great! Just a couple of nits

sdk/pinocchio/src/account_info.rs
febo: Removed it.

sdk/pinocchio/src/account_info.rs
febo: Yep, caught this one on the refactoring.

sdk/pinocchio/src/account_info.rs
febo: Good call, `is_borrowed` sounds good. Now it makes more sense that the method returns a `bool` – the "check" kind of implied that it would return a `Result`.

PR #148, opened 2025-04-30
Relax rent sysvar length check

febo:
### Problem

Currently the `Rent` sysvar  length validation is very strict – it requires the length of the byte buffer to be exactly the size of the type. This creates problems when running fuzzing tests since the sysvar account in some cases is bigger than the type.

### Solution

Relax the length test to require a minimum length instead. This PR also avoids initializing the byte buffer passed to the syscall, which saves CUs.


joncinque (approved):
Looks great! This is the correct behavior. I tested this in solana-sdk, and bincode ignores additional bytes in `Sysvar::from_account_info`. 

This also makes sense because it allows us to add fields to existing sysvars without breaking programs.

joncinque (approved):

PR #151, opened 2025-05-05
added from_acccount_info

Jars1987:
- Added from_account_info implementation to the sysvar Clock struct

This implementation mimics what happens in the Rent struct.

### DIFF
--- a/sdk/pinocchio/src/sysvars/clock.rs
+++ b/sdk/pinocchio/src/sysvars/clock.rs
@@ -1,7 +1,18 @@
 //! Information about the network's clock, ticks, slots, etc.
 
 use super::Sysvar;
-use crate::impl_sysvar_get;
+use crate::{
+    account_info::{AccountInfo, Ref},
+    impl_sysvar_get,
+    program_error::ProgramError,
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/sysvars/clock.rs
```diff
@@ -65,3 +76,67 @@ pub const DEFAULT_MS_PER_SLOT: u64 = 1_000 * DEFAULT_TICKS_PER_SLOT / DEFAULT_TI
...
+        }
+        Ok(Self::from_bytes_unchecked(
+            account_info.borrow_data_unchecked(),
+        ))
+    }
+
+    /// Return a `Clock` from the given bytes.
+    ///
+    /// This method performs a length validation. The caller must ensure that `bytes` contains
+    /// a valid representation of `Clock`.
```
febo, line 121: We are missing the `# Safety` heading here. Let's move the second sentence from the line above to it:
```rust
/// # Safety
///
/// The caller must ensure that `bytes` contains a valid representation of `Clock`.
```

febo (approved):
Looks great, thanks!

PR #153, opened 2025-05-06
Remove input offset counter

febo:
### Problem

Currently entrypoint implementations maintain an offset counter when parsing the input buffer. This is unnecessary since the input pointer can be increment directly.

### Solution

Remove the offset counter and increment the input pointer directly.

### DIFF
--- a/sdk/pinocchio/src/entrypoint/lazy.rs
+++ b/sdk/pinocchio/src/entrypoint/lazy.rs
@@ -245,8 +245,6 @@ impl InstructionContext {
     #[inline(always)]
     unsafe fn read_account(&mut self) -> MaybeAccount {
         let account: *mut Account = self.input as *mut Account;
-        // rent epoch (8 bytes)
-        self.input = self.input.add(core::mem::size_of::<u64>());
 
         if (*account).borrow_state == NON_DUP_MARKER {
             // repurpose the borrow state to track borrows
@@ -256,9 +254,11 @@ impl InstructionContext {
             self.input = self.input.add((*account).data_len as usize);
             self.input = self.input.add(MAX_PERMITTED_DATA_INCREASE);
             self.input = self.input.add(self.input.align_offset(BPF_ALIGN_OF_U128));
+            self.input = self.input.add(core::mem::size_of::<u64>());
 
             MaybeAccount::Account(AccountInfo { raw: account })
         } else {
+            self.input = self.input.add(core::mem::size_of::<u64>());
             //the caller will handle the mapping to the original account
             MaybeAccount::Duplicated((*account).borrow_state)
         }
--- a/sdk/pinocchio/src/entrypoint/lazy.rs
+++ b/sdk/pinocchio/src/entrypoint/lazy.rs
@@ -2,7 +2,8 @@
 //! input buffer.
 
 use crate::{
-    account_info::{Account, AccountInfo, MAX_PERMITTED_DATA_INCREASE},
+    account_info::{Account, AccountInfo},
+    entrypoint::STATIC_ACCOUNT_DATA,
     program_error::ProgramError,
     pubkey::Pubkey,
     BPF_ALIGN_OF_U128, NON_DUP_MARKER,
@@ -100,8 +101,10 @@ macro_rules! lazy_program_entrypoint {
 /// This is a wrapper around the input buffer that provides methods to read the accounts
 /// and instruction data. It is used by the lazy entrypoint to access the input data on demand.
 pub struct InstructionContext {
-    /// Pointer to the runtime input buffer for the instruction.
-    input: *mut u8,
+    /// Pointer to the runtime input buffer to read from.
+    ///
+    /// This pointer is moved forward as accounts are read from the buffer.
+    buffer: *mut u8,
 
     /// Number of remaining accounts.
     ///
@@ -134,8 +137,9 @@ impl InstructionContext {
     pub unsafe fn new_unchecked(input: *mut u8) -> Self {
         Self {
             // SAFETY: The first 8 bytes of the input buffer represent the
-            // number of accounts when serialized by the SVM loader.
-            input: unsafe { input.add(core::mem::size_of::<u64>()) },
+            // number of accounts when serialized by the SVM loader, which is read
+            // when the context is created.
+            buffer: unsafe { input.add(core::mem::size_of::<u64>()) },
             // SAFETY: Read the number of accounts from the input buffer serialized
             // by the SVM loader.
             remaining: unsafe { *(input as *const u64) },
@@ -206,10 +210,10 @@ impl InstructionContext {
     /// before reading all accounts will result in undefined behavior.
     #[inline(always)]
     pub unsafe fn instruction_data_unchecked(&self) -> &[u8] {
-        let data_len = *(self.input as *const usize);
+        let data_len = *(self.buffer as *const usize);
         // shadowing the input to avoid leaving it in an inconsistent position
-        let input = self.input.add(core::mem::size_of::<u64>());
-        core::slice::from_raw_parts(input, data_len)
+        let data = self.buffer.add(core::mem::size_of::<u64>());
+        core::slice::from_raw_parts(data, data_len)
     }
 
     /// Returns the program id for the instruction.
@@ -233,8 +237,8 @@ impl InstructionContext {
     /// before reading all accounts will result in undefined behavior.
     #[inline(always)]
     pub unsafe fn program_id_unchecked(&self) -> &Pubkey {
-        let data_len = *(self.input as *const usize);
-        &*(self.input.add(core::mem::size_of::<u64>() + data_len) as *const Pubkey)
+        let data_len = *(self.buffer as *const usize);
+        &*(self.buffer.add(core::mem::size_of::<u64>() + data_len) as *const Pubkey)
     }
 
     /// Read an account from the input buffer.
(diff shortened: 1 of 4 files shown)


sdk/pinocchio/src/entrypoint/lazy.rs
```diff
@@ -241,10 +233,35 @@ impl InstructionContext {
...
+    /// Read an account from the input buffer.
+    ///
+    /// This can only be called with a buffer that was serialized by the runtime as
+    /// it assumes a specific memory layout.
+    #[allow(clippy::cast_ptr_alignment, clippy::missing_safety_doc)]
+    #[inline(always)]
+    unsafe fn read_account(&mut self) -> MaybeAccount {
+        let account: *mut Account = self.input as *mut Account;
+        // rent epoch (8 bytes)
+        self.input = self.input.add(core::mem::size_of::<u64>());
```
joncinque, line 249: This confused me a bit, since it's adding the rent epoch for a non-dupe account, and adding the dupe-marker + padding for a dupe account. And it's adding the rent-epoch out of order, which made me think it was handling the rent-epoch of the *previous* account.

I know it adds more code, but just for the sake of legibility, can you move this line to where it was before?

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -160,37 +160,35 @@ macro_rules! program_entrypoint {
...
+        let total_accounts = processed;
         // number of accounts to process (limited to MAX_ACCOUNTS)
-        let processed = core::cmp::min(total_accounts, MAX_ACCOUNTS);
+        processed = core::cmp::min(total_accounts, MAX_ACCOUNTS);
 
         for i in 0..processed {
-            let account_info: *mut Account = input.add(offset) as *mut _;
+            let account_info: *mut Account = input as *mut Account;
+            // rent epoch (8 bytes)
+            input = input.add(core::mem::size_of::<u64>());
```
joncinque, line 178: Same with this one, I got very confused by it and had to re-read it many times, so for the sake of future legibility, can you put it back to how it was before?

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -200,38 +198,31 @@ pub unsafe fn deserialize<'a, const MAX_ACCOUNTS: usize>(
...
-        // process any remaining accounts to move the offset to the instruction
+        // Process any remaining accounts to move the offset to the instruction
         // data (there is a duplication of logic but we avoid testing whether we
-        // have space for the account or not)
+        // have space for the account or not).
         for _ in processed..total_accounts {
-            let account_info: *mut Account = input.add(offset) as *mut _;
+            let account_info: *mut Account = input as *mut Account;
+            // rent epoch (8 bytes)
+            input = input.add(core::mem::size_of::<u64>());
```
joncinque, line 207: Maybe we can let this one slide, but I'd again prefer to have the line in both branches

joncinque (commented):
Looks great! I just got a bit confused with one of the changes, so maybe we can improve that.

Does this improve performance of the entrypoint?

sdk/pinocchio/src/entrypoint/mod.rs
febo: I checked with @LucasSte and in this case it seems better to not have the `else` block since we "save" jumps when an account do not satisfy the `if`-path. I will update the comment to make it more clear.

febo:
> Does this improve performance of the entrypoint?

It saves between 1 or 2 CUs. I think it would save more in the case of many duplicated accounts. 😊

joncinque (approved):
Great job!

sdk/pinocchio/src/entrypoint/lazy.rs
febo: It turns out that avoiding duplicating the line, it saves `1` CU per account. Added a comment to make it more clear what the offset it representing.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Same here, it is slightly more efficient to avoid the duplication. Updated the comment to make it more clear what it represents.

joncinque (approved):
Even better!

sdk/pinocchio/src/entrypoint/lazy.rs
```diff
@@ -136,11 +133,12 @@ impl InstructionContext {
     #[inline(always)]
     pub unsafe fn new_unchecked(input: *mut u8) -> Self {
         Self {
-            input,
             // SAFETY: The first 8 bytes of the input buffer represent the
             // number of accounts when serialized by the SVM loader.
+            input: unsafe { input.add(core::mem::size_of::<u64>()) },
```
ilya-bobyr, line 138: The doc comment for `input` needs to be updated now, I think.

Would it also make sense to rename `input` to, maybe, `unprocessed_input` to make it more obvious that it is not the original data passed into `new()`?

sdk/pinocchio/src/entrypoint/lazy.rs
```diff
@@ -214,10 +206,10 @@ impl InstructionContext {
...
     #[inline(always)]
     pub unsafe fn instruction_data_unchecked(&self) -> &[u8] {
-        let data_len = *(self.input.add(self.offset) as *const usize);
-        // shadowing the offset to avoid leaving it in an inconsistent state
-        let offset = self.offset + core::mem::size_of::<u64>();
-        core::slice::from_raw_parts(self.input.add(offset), data_len)
+        let data_len = *(self.input as *const usize);
+        // shadowing the input to avoid leaving it in an inconsistent position
+        let input = self.input.add(core::mem::size_of::<u64>());
+        core::slice::from_raw_parts(input, data_len)
```
ilya-bobyr, line 212: *minor*

When I was reading this code at first, it seemed to me that the function updates `self.input`, due to the `input` name being reused.
Considering this code is constructing a slice that covers the instruction data, perhaps it would be better not to use the `input` name again, and call it `data`, for example:

```suggestion
        let data = self.input.add(core::mem::size_of::<u64>());
        core::slice::from_raw_parts(data, data_len)
```
---
P.S. Probably out of scope for this PR, but why not import `size_of` as it seems to be used quite a lot in this module?

```rust
    use {
        /* ... */
        core::mem::size_of,
        /* ... */
    };

    /* ... */

        let data = self.input.add(size_of::<u64>());
        core::slice::from_raw_parts(data, data_len)
```

sdk/pinocchio/src/entrypoint/lazy.rs
```diff
@@ -241,10 +233,37 @@ impl InstructionContext {
...
-            .add(self.offset + core::mem::size_of::<u64>() + data_len) as *const Pubkey)
+        let data_len = *(self.input as *const usize);
+        &*(self.input.add(core::mem::size_of::<u64>() + data_len) as *const Pubkey)
+    }
+
+    /// Read an account from the input buffer.
+    ///
+    /// This can only be called with a buffer that was serialized by the runtime as
+    /// it assumes a specific memory layout.
+    #[allow(clippy::cast_ptr_alignment, clippy::missing_safety_doc)]
```
ilya-bobyr, line 244: Why disable `clippy::missing_safety_doc` when the comment already has the necessary documentation?

```suggestion
    /// Read an account from the input buffer.
    ///
    /// # Safety
    ///
    /// This can only be called with a buffer that was serialized by the runtime as
    /// it assumes a specific memory layout.
    #[allow(clippy::cast_ptr_alignment)]
```

sdk/pinocchio/src/entrypoint/lazy.rs
```diff
@@ -241,10 +233,37 @@ impl InstructionContext {
...
+        self.input = self.input.add(core::mem::size_of::<u64>());
+
+        if (*account).borrow_state == NON_DUP_MARKER {
+            // Unique account: repurpose the borrow state to track borrows.
+            (*account).borrow_state = 0b_0000_0000;
+
+            self.input = self.input.add(core::mem::size_of::<Account>());
+            self.input = self.input.add((*account).data_len as usize);
+            self.input = self.input.add(MAX_PERMITTED_DATA_INCREASE);
+            self.input = self.input.add(self.input.align_offset(BPF_ALIGN_OF_U128));
```
ilya-bobyr, line 260: *minor*

In two other locations where the same code is repeated, you added `core::mem::size_of::<Account>()` and `MAX_PERMITTED_DATA_INCREASE`, but here these additions are separate.

Maybe extract this increment into a common function used in all 3 locations?

sdk/pinocchio/src/entrypoint/lazy.rs
febo: Good call on both!

sdk/pinocchio/src/entrypoint/lazy.rs
febo: Good call, I will address this and the imports in a separate PR.

sdk/pinocchio/src/entrypoint/lazy.rs
```diff
@@ -100,16 +101,15 @@ macro_rules! lazy_program_entrypoint {
...
+    buffer: *mut u8,
 
     /// Number of remaining accounts.
     ///
     /// This value is decremented each time [`next_account`] is called.
     remaining: u64,
-
-    /// Current memory offset on the input buffer.
-    offset: usize,
 }
```
ilya-bobyr, line 113: I wonder if the `buffer`/`remaining` pair is really different from a slice with a `'static` lifetime.
If we replace this pair with a slice, we could probably reduce the number of `unsafe` blocks by a lot.
Construction of the context will still be an unsafe operation, as it can only be built based from a pointer that the compiler can not fully verify.

This is not directly related to this PR.
Just a possible improvement to consider in the future, perhaps?

PR #154, opened 2025-05-07
Simplify invoke instruction

### LINKED ISSUE #149: `slice_invoke_signed` inline wasted too much stack
#147 made `slice_invoke_signed` inline. the function allocates 64 `Account`s on the stack and each `Account` takes ~49 bytes, which in total will take up 3000 bytes. One stack frame is 4096 bytes, and it will easily overflow.

@febo

febo:
### Problem

Currently `invoke` and `invoke_signed` include an additional validation on the number of accounts that consumes CUs and, at the same time, does not cover all cases:
```rust
if instruction.accounts.len() < ACCOUNTS {
}
```
For example, this does not cover the case where `instruction.accounts.len() > ACCOUNTS`.

Additionally, iteration over the account infos array uses indices, which then require index access to the accounts array being built.

Finally, both `invoke` and `slice_invoke` are inlined function. This create problems when a large number of accounts is used, e.g., likely the code will run out of stack space.

### Solution

This PR refactors the invoke helpers into 3 different pair of helpers:
* `invoke` and `invoke_signed`: expect an array of `AccountInfo`s and it is inlined
* `invoke_with_bounds` and `invoke_signed_with_bounds`: expect a slice of `AccountInfo`s, a `MAX_ACCOUNTS` constant representing the upper limit for the static allocation and it is inlined
* `slice_invoke` and `slice_invoke_signed`: expect a slice of `AccountInfo`s and it is not inlined

The `*_with_bonds` variants allows code that uses slices to still be inlined, provided that there is enough stack space. The `MAX_ACCOUNTS` parameter adjust the amount of stack memory used by the helpers.

In all cases, the number of accounts passed to helpers must be at least the same as the number of accounts expected by the instruction – e.g.,:
```rust
if account_infos.len() < instruction.accounts.len() {
}
```

Fixes #149

### DIFF
--- a/sdk/pinocchio/src/cpi.rs
+++ b/sdk/pinocchio/src/cpi.rs
@@ -47,7 +47,9 @@ pub fn invoke<const ACCOUNTS: usize>(
 /// the number of accounts that need to be statically allocated for cases where the
 /// number of instruction accounts is not known at compile time. The final number of
 /// accounts passed to the cross-program invocation will be the number of accounts
-/// required by the `instruction`, even if `MAX_ACCOUNTS` is greater than that.
+/// required by the `instruction`, even if `MAX_ACCOUNTS` is greater than that. When
+/// `MAX_ACCOUNTS` is lower than the number of accounts expected by the instruction,
+/// this function will return a [`ProgramError::InvalidArgument`] error.
 ///
 /// Note that this function is inlined to avoid the overhead of a function call,
 /// but uses stack memory allocation. When a large number of accounts is needed,
@@ -74,9 +76,11 @@ pub fn invoke_with_bounds<const MAX_ACCOUNTS: usize>(
 /// function with the signers' seeds set to an empty slice.
 ///
 /// Note that the maximum number of accounts that can be passed to a cross-program
-/// invocation is defined by the `MAX_CPI_ACCOUNTS` constant. Even if the slice
+/// invocation is defined by the [`MAX_CPI_ACCOUNTS`] constant. Even if the slice
 /// of `AccountInfo`s has more accounts, only the number of accounts required by
-/// the `instruction` will be used.
+/// the `instruction` will be used. If the number of accounts required by the
+/// instruction is greater than [`MAX_CPI_ACCOUNTS`], this function will return a
+/// [`ProgramError::InvalidArgument`] error.
 ///
 /// # Important
 ///
@@ -144,7 +148,7 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
 ///      accounts in the instruction.
 ///
 /// This validation is done to ensure that the borrow checker rules are followed,
-/// consuming CUs in the process. The `invoke_signed_unchecked` is an alternative
+/// consuming CUs in the process. The [`invoke_signed_unchecked`] is an alternative
 /// to this function that has lower CU consumption since it does not perform
 /// any validation. This should only be used when the caller is sure that the borrow
 /// checker rules are followed.
@@ -154,7 +158,9 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
 /// the number of accounts that need to be statically allocated for cases where the
 /// number of instruction accounts is not known at compile time. The final number of
 /// accounts passed to the cross-program invocation will be the number of accounts
-/// required by the `instruction`, even if `MAX_ACCOUNTS` is greater than that.
+/// required by the `instruction`, even if `MAX_ACCOUNTS` is greater than that. When
+/// `MAX_ACCOUNTS` is lower than the number of accounts expected by the instruction,
+/// this function will return a [`ProgramError::InvalidArgument`] error.
 ///
 /// Note that this function is inlined to avoid the overhead of a function call,
 /// but uses stack memory allocation. When a large number of accounts is needed,
@@ -180,7 +186,7 @@ pub fn invoke_signed_with_bounds<const MAX_ACCOUNTS: usize>(
     // number of accounts expected by the instruction is done in
     // `invoke_signed_with_bounds`.
     if MAX_ACCOUNTS < instruction.accounts.len() {
-        return Err(ProgramError::NotEnoughAccountKeys);
+        return Err(ProgramError::InvalidArgument);
     }
 
     // SAFETY: The stack allocated account storage `MAX_ACCOUNTS` was validated.
@@ -201,7 +207,7 @@ pub fn invoke_signed_with_bounds<const MAX_ACCOUNTS: usize>(
 ///      accounts in the instruction.
 ///
 /// This validation is done to ensure that the borrow checker rules are followed,
-/// consuming CUs in the process. The `invoke_signed_unchecked` is an alternative
+/// consuming CUs in the process. The [`invoke_signed_unchecked`] is an alternative
 /// to this function that have lower CU consumption since it does not perform
 /// any validation. This should only be used when the caller is sure that the borrow
 /// checker rules are followed.
@@ -209,7 +215,9 @@ pub fn invoke_signed_with_bounds<const MAX_ACCOUNTS: usize>(
 /// Note that the maximum number of accounts that can be passed to a cross-program
 /// invocation is defined by the `MAX_CPI_ACCOUNTS` constant. Even if the slice
 /// of `AccountInfo`s has more accounts, only the number of accounts required by
-/// the `instruction` will be used.
+/// the `instruction` will be used. If the number of accounts required by the
+/// instruction is greater than [`MAX_CPI_ACCOUNTS`], this function will return a
+/// [`ProgramError::InvalidArgument`] error.
 ///
 /// # Important
 ///
@@ -222,14 +230,14 @@ pub fn slice_invoke_signed(
     account_infos: &[&AccountInfo],
     signers_seeds: &[Signer],
 ) -> ProgramResult {
-    // Check that the stack allocated account storage `MAX_ACCOUNTS` is sufficient
-    // for the number of accounts expected by the instruction.
+    // Check that the stack allocated account storage `MAX_CPI_ACCOUNTS` is
+    // sufficient for the number of accounts expected by the instruction.
     //
     // The check for the slice of `AccountInfo`s not being less than the
     // number of accounts expected by the instruction is done in
     // `invoke_signed_with_bounds`.
     if MAX_CPI_ACCOUNTS < instruction.accounts.len() {
-        return Err(ProgramError::NotEnoughAccountKeys);
+        return Err(ProgramError::InvalidArgument);
     }
 
     // SAFETY: The stack allocated account storage `MAX_CPI_ACCOUNTS` was validated.


sdk/pinocchio/src/cpi.rs
```diff
@@ -50,34 +87,45 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
     account_infos: &[&AccountInfo; ACCOUNTS],
     signers_seeds: &[Signer],
 ) -> ProgramResult {
-    if instruction.accounts.len() < ACCOUNTS {
-        return Err(ProgramError::NotEnoughAccountKeys);
-    }
```
joncinque, line 55: Isn't there still a chance that `instruction.accounts.len()` is greater than `ACCOUNTS`? Or are you relying on the syscall failing because we don't provide enough account infos for the instruction?

sdk/pinocchio/src/cpi.rs
```diff
@@ -102,48 +162,55 @@ pub fn slice_invoke_signed(
     account_infos: &[&AccountInfo],
     signers_seeds: &[Signer],
 ) -> ProgramResult {
-    if instruction.accounts.len() < account_infos.len() {
-        return Err(ProgramError::NotEnoughAccountKeys);
-    }
```
joncinque, line 107: Same here, are you expecting the syscall to fail if this condition isn't met?

sdk/pinocchio/src/instruction.rs
```diff
@@ -78,6 +81,26 @@ const fn offset<T, U>(ptr: *const T, offset: usize) -> *const U {
...
+    /// Return true if the account borrow state is set to the given state.
+    ///
+    /// This will test both data and lamports borrow state, similarly to
+    /// [`AccountInfo::is_borrowed`].
+    #[inline(always)]
+    pub fn is_borrowed(&self, state: BorrowState) -> bool {
+        // SAFETY: The `borrow_state` of an account is stored in the first byte of the
+        // memory location of the account on the input buffer, which starts at 8 bytes
+        // before the account `key`.
+        unsafe { *(self.key as *const u8).sub(8) & (state as u8) != 0 }
```
joncinque, line 94: This is extremely unsafe, and only works in a VM context, where `Account` was created from an `AccountInfo` passed by the runtime. If we can avoid having this function at all and instead perform the checks on the `AccountInfo`s, it would definitely be safer.

We might just want to have the `invoke_signed` for those who want some validation checks, and `invoke_instruction_signed_unchecked` who want full control. Anything in between might not be necessary and more confusing than helpful

sdk/pinocchio/src/cpi.rs
```diff
@@ -184,11 +251,103 @@ pub unsafe fn invoke_signed_unchecked(
...
+
+/// Invoke a cross-program instruction with signatures.
+#[inline]
+pub fn invoke_instruction_signed<const ACCOUNTS: usize>(
+    program_id: &Pubkey,
+    priviledges: &[Priviledge; ACCOUNTS],
+    instruction_data: &[u8],
+    accounts: &[Account; ACCOUNTS],
+    signers_seeds: &[Signer],
+) -> ProgramResult {
```
joncinque, line 283: I see what you're trying to do here, but the API is a bit strange, since it's requiring the `privileges` line up perfectly with `accounts`. Are there a lot of gains from it? Because it seems a bit delicate to use.

joncinque:
Also added @rustopian who will be using these helpers in the Pinocchio ATA program, so let's get him familiar with them!

sdk/pinocchio/src/cpi.rs
```diff
@@ -184,11 +251,103 @@ pub unsafe fn invoke_signed_unchecked(
...
+                    account.key(),
+                    priviledge.0,
+                    priviledge.1,
+                ));
+
+                account_metas_ptr = account_metas_ptr.add(1);
+            }
+
+            Ok(())
+        })?;
```
ilya-bobyr, line 312: *bikeshedding*

You can also use iterators to avoid both an `unsafe` block and explicit pointer.
`MaybeUninit` has an example of how to [initialize an array element by element](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element).

```suggestion
    /*
     * use itertools::izip;
     */

    let mut account_metas = MaybeUninit::<[AccountMeta; ACCOUNTS]>::uninit();

    for (account, (is_writable, is_signer), account_meta) in
        izip!(accounts, priviledges, &mut account_meta)
    {
        // Check whether all account infos can be safely borrowed according
        // to their mutability on the instruction or not.
        if account.is_borrowed(if is_writable {
            BorrowState::Borrowed
        } else {
            BorrowState::MutablyBorrowed
        }) {
            return Err(ProgramError::AccountBorrowFailed);
        }

        account_meta.write(AccountMeta::new(account.key(), is_writable, is_signer));
    }
    
    // SAFETY: Previous loop only finishes when all elements have been initialized.
    let account_metas = unsafe { account_metas.assume_init_ref() };
```

sdk/pinocchio/src/cpi.rs
```diff
@@ -13,8 +13,27 @@ use crate::{
...
+/// an account priviledge.
+pub const WRITABLE: bool = true;
+
+/// A type representing the privileges of an account for
+/// `invoke_instruction` and `invoke_instruction_signed`.
+///
+/// The first element of the tuple indicates whether the account is
+/// writable or not, and the second element indicates whether the
+/// account is a signer or not.
+pub type Priviledge = (bool, bool);
```
ilya-bobyr, line 30: I think it is better to create a full flagged wrapper, rather than a type alias.
The advantages of a wrapper are that you can define methods for it, as well as trait instances.

It also allows you to hide the internal implementation.

With a "transparent" definition as this one, all the users of this type still directly access the two internal fields, which are both just `bool`.
So every location that uses this type will need to know which of the two bools is the writable flag, and which is the signer flag.
It is not impossible to mix them up.  And the compiler can not check for you.

Plus, if you decide to change the meaning, like you want to add a third flag, the compiler will provide less help during the refactoring process.

Considering this type is part of the API, I would be particularly careful when exposing internal structure.

Considering you are trying to optimize things at a lower level, it seems reasonable to use a `u8` field for "flags" and put both writable and the signer flags in it.
[`bitflags`](https://docs.rs/bitflags/latest/bitflags/) is a pretty popular crate for writing the boilerplate code for this kind of cases.

sdk/pinocchio/src/cpi.rs
febo: Yes, since we avoid the `Pubkey` comparison to align with each `AccountInfo`, as the `invoke_signed` requires. But maybe the solution is for clients to optimize this and use `invoke_signed_unchecked`.

sdk/pinocchio/src/cpi.rs
febo: Adding a test `instruction.accounts.len() != ACCOUNTS` at the start of the function increases CUs by `100`, so I am tempted to let the syscall fail instead.

sdk/pinocchio/src/instruction.rs
febo: I have done a bit more testing with the `invoke_signed` and it seems that there is almost no difference compared to the proposed `invoke_instruction_signed`. I think it is ok to remove the new ones and keep the existing API.

sdk/pinocchio/src/cpi.rs
febo: I have done a bit more testing with the invoke_signed and it seems that there is almost no difference compared to the proposed invoke_instruction_signed. I think it is ok to remove the new ones and keep the existing API.

sdk/pinocchio/src/cpi.rs
febo: In the end we removed the need to this "type".

sdk/pinocchio/src/cpi.rs
```diff
@@ -102,48 +146,55 @@ pub fn slice_invoke_signed(
...
+            // to their mutability on the instruction or not.
+            let state = if account_meta.is_writable {
+                BorrowState::Borrowed
+            } else {
+                BorrowState::MutablyBorrowed
+            };
+
+            if account_info.is_borrowed(state) {
+                return Err(ProgramError::AccountBorrowFailed);
+            }
```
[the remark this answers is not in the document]
febo: Good point! Updated the variable to `invalid_state` and changed the comment to explain what it is testing.

sdk/pinocchio/src/cpi.rs
```diff
@@ -234,7 +283,10 @@ pub unsafe fn invoke_signed_unchecked(
     }
 
     #[cfg(not(target_os = "solana"))]
-    core::hint::black_box((instruction, accounts, signers_seeds));
+    {
+        core::hint::black_box((instruction, accounts, signers_seeds));
+        unreachable!();
+    }
```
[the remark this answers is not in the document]
febo: It is unreachable because this package is meant to be used on-chain only (i.e., `target_os = "solana"`). When the target is not `"solana"`, it should either do nothing or panic. Agree that it would be better to include an explicit panic with a message, so I removed the `unreachable!()` for now and will make this change in a separate PR.

sdk/pinocchio/src/cpi.rs
```diff
@@ -202,7 +251,7 @@ pub unsafe fn invoke_signed_unchecked(
             program_id: *const Pubkey,
 
             /// Accounts expected by the program instruction.
-            accounts: *const AccountMeta<'a>,
+            accounts: *const crate::instruction::AccountMeta<'a>,
```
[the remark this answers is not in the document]
febo: Move to the beginning of the `#[cfg(target_os = "solana")]` block, since it would only be used the target is `"solana"`.

sdk/pinocchio/src/cpi.rs
febo: Nice, we can then remove the `enumerate` on the other iterations.

sdk/pinocchio/src/cpi.rs
```diff
@@ -50,34 +68,44 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
...
+                return Err(ProgramError::AccountBorrowFailed);
+            }
+
+            accounts.write(Account::from(*account_info));
+
+            Ok(())
+        })?;
+
+    // SAFETY: At this point it is guaranteed that all account infos are
+    // borrowable according to their mutability on the instruction.
```
ilya-bobyr, line 108: I do not know if this matters, but if `instruction.accounts` is shorter than `ACCOUNTS`, not all `ACCOUNTS` elements in the `accounts` array will be initialized.

sdk/pinocchio/src/cpi.rs
```diff
@@ -102,48 +142,53 @@ pub fn slice_invoke_signed(
...
+        })?;
+
+    // SAFETY: At this point it is guaranteed that all account infos are
+    // borrowable according to their mutability on the instruction, which
+    // in the worst case is more than what is needed.
     unsafe {
         invoke_signed_unchecked(
             instruction,
-            core::slice::from_raw_parts(accounts.as_ptr() as _, len),
+            core::slice::from_raw_parts(accounts.as_ptr() as _, account_infos.len()),
```
ilya-bobyr, line 191: Similarly, here, only the `min(instruction.accounts.len(), account_infos.len())` elements will actually be initialized.

Should we use a `min()` here?

sdk/pinocchio/src/cpi.rs
febo: Yes, we should avoid that – I just pushed this change + a few small tweaks to the imports.

ilya-bobyr: I thought about it a bit more, and I wonder if `min` is really the right behavior here.
`instruction` argument describes the invocation that the caller of the `invoke_signed()` or the `slice_invoke_signed()` is requesting.
When they specify a certain value for the `accounts` field, they probably expect the invocation to contain exactly this many accounts.
So, it seems that we should rather panic or return `Err(ProgramError::InvalidArgument)` when the actual number of accounts turns out to be different from `instruction.accounts.len()`.

All the other errors are reported as `Err(ProgramError::Something)`, even a `account_infos.len() > MAX_CPI_ACCOUNTS` violation.  So, maybe this case should also be an `Err()` result.
At the same time, it seems that the `account_infos.len() > MAX_CPI_ACCOUNTS` and the insufficient number of accounts are API violations, which are better handled by a panic, rather than an error.

febo: The idea was to simplify the invoke and let the syscall fail in cases where there are accounts missing since the syscall will validate the input as well without consuming more CUs of the program – I updated the description of the PR to reflect this. For some reason, adding a simple `if` statement to test whether the length mismatches adds `100` CUs. 🤷‍♂️

But perhaps we won't be able to avoid it since there is an edge case that we won't be able to validate. Assume an example that we require `5` accounts on the instruction (the last one required to be writable) but we pass only `4` account infos. We can only validate the borrow of the  `4` account infos, which corresponds to the first `4` account metas of the instruction. Now imagine that the `5`th one is a duplicated account that matches one of the given account infos. The syscall is not going to fail since all accounts have been passed in, but the problem is that we did not validate that it is safe to borrow account `5` as writable – this is a problem!

cc: @LucasSte to see if he can help us understand why adding the `if` statement is so costly in this case.

febo: Did a quick comparison using the same instruction with 2 accounts:
* `invoke`: `1315` CUs
* `slice_invoke`: `1468` CUs (+153)

Apart from statically allocating a large `MaybeUninit` array and the `if` statement validating the accounts length, they are very similar functions.

LucasSte: > to see if he can help us understand why adding the if statement is so costly in this case.

I read the conversation, but I didn't follow which if statement you'd like to include here. Is it the one in the PR description?

febo: > > to see if he can help us understand why adding the if statement is so costly in this case.
> 
> I read the conversation, but I didn't follow which if statement you'd like to include here. Is it the one in the PR description?

Yes, something like:
```rust
if instruction.accounts.len() != ACCOUNTS {
}
```
At the very start of the `invoke_signed` function.

LucasSte: Even though there is the `#[inline]` attribute on top of the `invoke_signed` function, when you add the if, the function is not inlined anymore. The hint does not guarantee the function will be inlined, and I believe the extra length makes inlining not worth as the penalty for code size would be too large.

sdk/pinocchio/src/cpi.rs
```diff
@@ -92,58 +129,80 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
...
     signers_seeds: &[Signer],
 ) -> ProgramResult {
-    if instruction.accounts.len() < account_infos.len() {
-        return Err(ProgramError::NotEnoughAccountKeys);
-    }
-
-    if account_infos.len() > MAX_CPI_ACCOUNTS {
+    // Check that the number of accounts match the expected accounts by
+    // the instruction and do not exceed the maximum allowed.
+    if instruction.accounts.len() != account_infos.len() || account_infos.len() > MAX_CPI_ACCOUNTS {
```
ilya-bobyr, line 157: Should there be the same check in `invoke_signed()`?
Or are you still trying to figure out the CU increase it is causing?

P.S.  Is a 100 CU increase a substantially bad thing?
Just curious as to the scale of the CU costs.

sdk/pinocchio/src/cpi.rs
```diff
@@ -92,58 +129,80 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
...
+
+    // SAFETY: At this point it is guaranteed that account infos are borrowable
+    // according to their mutability on the instruction.
     unsafe {
         invoke_signed_unchecked(
             instruction,
-            core::slice::from_raw_parts(accounts.as_ptr() as _, len),
+            from_raw_parts(
+                accounts.as_ptr() as _,
+                min(account_infos.len(), instruction.accounts.len()),
```
ilya-bobyr, line 204: `min()` here is no longer necessary.
Not sure if it would account for the CU increase you saw due to the addition of the `if` check.
But maybe if you remove the `min()` call the total CU cost will also decrease a bit?

sdk/pinocchio/src/instruction.rs
```diff
@@ -135,25 +135,25 @@ impl<'a> AccountMeta<'a> {
 
     /// Creates a new readonly `AccountMeta`.
     #[inline(always)]
-    pub fn readonly(pubkey: &'a Pubkey) -> Self {
+    pub const fn readonly(pubkey: &'a Pubkey) -> Self {
```
ilya-bobyr, line 138: Not sure if these `const` qualifications are related to the other work you are doing in this PR.
If they are not, you might be able to get them in a separate PR faster.

rustopian (approved):
Thanks for tag, informative. Looks good to me.

sdk/pinocchio/src/cpi.rs
febo: Oh, you are right, nice! So what would be the best alternative? If we add `#[inline(always)]` to `invoke_signed` then it can save `100` CUs. The drawback would be if you try to call it with a large `ACCOUNTS` value, you might end up with no stack. Would be acceptable a workaround like [this](https://nnethercote.github.io/perf-book/inlining.html#harder-cases)?

LucasSte: > Would be acceptable a workaround like [this](https://nnethercote.github.io/perf-book/inlining.html#harder-cases)?

Yes, I believe that is a reasonable workaround.

sdk/pinocchio/src/cpi.rs
febo: A `100` increase for a CPI call is not good since it represents almost half of the variable CU required for a CPI.

febo:
Refactored the helpers so now there is one that works with slices and it is not inlined. Happy to get suggestions about the naming since I am not sure if the `_with_bounds` suffix is intuitive.

joncinque (dismissed):
Looks good to me, but please get approval from the other reviewers first

sdk/pinocchio/src/cpi.rs
```diff
@@ -13,148 +13,253 @@ use crate::{
...
+            Ok(())
+        })?;
+
+    // SAFETY: At this point it is guaranteed that account infos are borrowable
+    // according to their mutability on the instruction.
     unsafe {
         invoke_signed_unchecked(
             instruction,
-            core::slice::from_raw_parts(accounts.as_ptr() as _, len),
+            from_raw_parts(accounts.as_ptr() as _, instruction.accounts.len()),
```
ilya-bobyr, line 212: I think this function is missing a check, that ensures that `MAX_ACCOUNTS` is greater or equal to the `instruction.accounts.len()`.
If I call
```rust
invoke_signed_with_bounds<1>(
    /* instruction that takes 2 accounts */,
    &[ /* two AccountInfo reference */],
    &[]
)
```
this `invoke_signed_unchecked()` would be accessing past the allocated `accounts` array, no?

sdk/pinocchio/src/cpi.rs
```diff
@@ -13,148 +13,253 @@ use crate::{
...
-        unsafe {
-            accounts
-                .get_unchecked_mut(len)
-                .write(Account::from(*account_info));
-        }
-
-        len += 1;
-    }
-    // SAFETY: The accounts have been validated.
+    let mut accounts = [UNINIT; MAX_ACCOUNTS];
```
ilya-bobyr, line 172: If we know that the `Account` structures are so big, is there a good reason we do not allocate them on the heap?

```suggestion
    let mut accounts = Box::new([UNINIT; MAX_ACCOUNTS]);
```

CUs for the heap allocation are significant?

sdk/pinocchio/src/cpi.rs
febo: Good catch! Added a test for the constant length.

sdk/pinocchio/src/cpi.rs
febo: Significance is relative 😊, but a heap allocation definitely has an overhead (`~70` CUs more).

LucasSte (dismissed):

sdk/pinocchio/src/cpi.rs
```diff
@@ -13,137 +13,269 @@ use crate::{
...
+    signers_seeds: &[Signer],
+) -> ProgramResult {
+    // Check that the potential number of accounts provided is not less than
+    // the number of accounts expected by the instruction;
+    //
+    // The check for the slice of `AccountInfo`s not being less than the
+    // number of accounts expected by the instruction is done in
+    // `invoke_signed_with_bounds`.
+    if MAX_ACCOUNTS < instruction.accounts.len() {
+        return Err(ProgramError::NotEnoughAccountKeys);
```
illia-bobyr: Sorry for being picky, but I find this comment confusing.
The check is not looking at the number of provided accounts.
It only checks the number of accounts an instruction is expecting.

I think the comment should say something like "make sure that the stack allocated account storage is sufficient for the instruction being used".

sdk/pinocchio/src/cpi.rs
```diff
@@ -13,137 +13,269 @@ use crate::{
...
-        // `MAX_CPI_ACCOUNTS`.
-        unsafe {
-            accounts
-                .get_unchecked_mut(len)
-                .write(Account::from(*account_info));
-        }
+/// Internal function to invoke a cross-program instruction with signatures
+/// from a slice of `AccountInfo`s performing borrow checking.
+#[inline(always)]
+fn _invoke_signed_with_bounds<const MAX_ACCOUNTS: usize>(
```
illia-bobyr: In Rust, symbol names that start with an underscore are excluded from the reachability analysis.
So you should not use it for things that are actually used.

I would actually argue this function is `unsafe`, as there is away to call it that would cause undefined behavior.  And this is exactly what `unsafe` is supposed to mark.
So, I think, it should be just an unsafe function, and the safety comment would explain the relationship between `MAX_ACCOUNTS` and both `instruction.accounts.len()` and `accounts_infos.len()`.

sdk/pinocchio/src/cpi.rs
febo: Good point – comment updated to read:
```rust
// Check that the stack allocated account storage `MAX_ACCOUNTS` is sufficient
// for the number of accounts expected by the instruction.
```

sdk/pinocchio/src/cpi.rs
febo: Cool, fixed that. Also, improve the comments on the invoke functions. This is the commit [56b0b54](https://github.com/anza-xyz/pinocchio/pull/154/commits/56b0b54f19379c00758bbf53d26f57aa3c3f42f8).

sdk/pinocchio/src/cpi.rs
```diff
@@ -13,137 +13,315 @@ use crate::{
...
+    signers_seeds: &[Signer],
+) -> ProgramResult {
+    // Check that the stack allocated account storage `MAX_ACCOUNTS` is sufficient
+    // for the number of accounts expected by the instruction.
+    //
+    // The check for the slice of `AccountInfo`s not being less than the
+    // number of accounts expected by the instruction is done in
+    // `invoke_signed_with_bounds`.
+    if MAX_ACCOUNTS < instruction.accounts.len() {
+        return Err(ProgramError::NotEnoughAccountKeys);
```
illia-bobyr: *nitpicking*

I wonder if `InvalidArgument` would be a better error here.
Technically, the number of accounts is checked elsewhere.
And this check is just to make sure that enough stack is allocated for the execution.

The reason is that, when people see the error, they make a guess as to what might be causing it.
And if I would see `NotEnoughAccountKeys` from a call to `invoke_signed_with_bounds()` I would be looking at the `account_infos` argument as the first suspect.
`InvalidArgument`, on the other hands, means something more along the lines of "there is an error somehow related to the accounts, you will need to inspect pretty much everything to find it".

We could probably mention that if `MAX_ACCOUNTS` is too small then `InvalidArgument` is returned, in the method doc comment as well.  Hoping that it would save someone the need to read the method source.

Ideally, we should have a dedicated error code here.
But we have to reuse an existing one, I guess.

sdk/pinocchio/src/cpi.rs
```diff
@@ -13,137 +13,315 @@ use crate::{
...
 ) -> ProgramResult {
-    if instruction.accounts.len() < account_infos.len() {
+    // Check that the stack allocated account storage `MAX_ACCOUNTS` is sufficient
+    // for the number of accounts expected by the instruction.
+    //
+    // The check for the slice of `AccountInfo`s not being less than the
+    // number of accounts expected by the instruction is done in
+    // `invoke_signed_with_bounds`.
+    if MAX_CPI_ACCOUNTS < instruction.accounts.len() {
         return Err(ProgramError::NotEnoughAccountKeys);
```
illia-bobyr: Same here, I think `InvalidArgument` might be a better error here.

sdk/pinocchio/src/cpi.rs
febo: You are spot on! `ProgramError::InvalidArgument` is definitely the one to use. I am not sure we need a dedicated one, since this error will be raised because of an implementation error (i.e., using a wrong value for the static constant), not for a "variable" condition.

ilya-bobyr (approved):

illia-bobyr:
How come approvals are not removed when you make changes?
Is this a repo configuration?
This is a bit unusual.

febo:
> How come approvals are not removed when you make changes? Is this a repo configuration? This is a bit unusual.

@joncinque Should we enable this?

LucasSte:
> Is this a repo configuration?

Yes, it is. The admin must enable that option in the settings.

febo:
> > Is this a repo configuration?
> 
> Yes, it is. The admin must enable that option in the settings.

It is now configured to dismiss approvals when new commits are pushed, together with all the other usual settings.

PR #155, opened 2025-05-08
log: Improve digit logic

febo:
### Problem

Currently the logic to determine the digit to log uses an array, which is unnecessary since a digit can be determine from an offset and a starting value of `b'0'`.

### Solution

Change the logic to use `b'0' + remainder`, where remainder represents the digit to log. This PR also simplifies a reference to `self.len()` to just `self.len` to avoid dereferencing the log buffer.


joncinque (approved):
Looks good, nice and simple!

PR #156, opened 2025-05-08
log: Add support for `bool` type

febo:
### Problem

Currently it is not possible to log `bool` values.

### Solution

Add support for `bool` type.


joncinque (approved):

Nagaprasadvr:
Yo this is cool ty ser

PR #157, opened 2025-05-22
Use static syscalls when available

LucasSte:
**Problem**

Pinocchio logger declares syscalls as extern functions, but this method is not compatible with SBPFv3. If we build a program using the new v3 architecture, we'll encounter an invalid function error.

**Solution**

Declare static syscalls in an internal `syscall` module.


febo (approved):
Thanks!

PR #158, opened 2025-05-24
Add `RefMut::map_split` method and unit test

dvrvsimi:
### Summary

This PR adds a `map_split` method to `RefMut` in the Pinocchio SDK, emulating the behavior of `std::cell::RefMut::map_split` from the Rust standard library. It allows safely splitting a mutable borrow into two non-overlapping mutable borrows, which is useful for working with disjoint parts of account data in Solana programs.

### Details

- **Feature:**  
  - Implements `RefMut::map_split`, which takes a closure that splits a mutable reference into two non-overlapping mutable references, returning two `RefMut`s that share the same borrow state.
  - Adds comprehensive documentation to clarify safety requirements and drop semantics.
- **Testing:**  
  - Adds a thorough unit test for `map_split` that:
    - Splits a `RefMut<[u8]>` into two slices and mutates both.
    - Asserts correct mutation of the underlying data.

### Motivation

Inspired by this [solana stackexchange question](https://solana.stackexchange.com/questions/22154/pinocchiorefmut-is-missing-the-map-split-function), the feature brings Pinocchio’s ergonomics closer to the Rust standard library, making it easier for developers to write safe, efficient Solana programs that need to work with disjoint regions of account data.

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -689,6 +689,53 @@ impl<'a, T: ?Sized> RefMut<'a, T> {
             None => Err(ManuallyDrop::into_inner(orig)),
         }
     }
+
+    /// Splits a `RefMut` into two `RefMut`s for different, non-overlapping components.
+    ///
+    /// # Safety
+    ///
+    /// The closure **must** return two non-overlapping mutable references derived from the original.
+    /// If the returned references overlap, this will cause undefined behavior.
+    ///
+    /// Emulates the behavior of [`std::cell::RefMut::map_split`].
+    ///
+    /// # Example
+    /// ```ignore
+    /// let (left, right) = RefMut::map_split(ref_mut, |slice| slice.split_at_mut(2));
+    /// // left and right are now independent mutable borrows of disjoint parts of the slice
+    /// ```
+    #[inline]
+    pub fn map_split<U: ?Sized, V: ?Sized, F>(
+        orig: RefMut<'a, T>,
+        f: F,
+    ) -> (RefMut<'a, U>, RefMut<'a, V>)
+    where
+        F: FnOnce(&mut T) -> (&mut U, &mut V),
+    {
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/account_info.rs
```diff
@@ -834,4 +881,44 @@ mod tests {
...
+        // Mutate both halves independently
+        left.copy_from_slice(&[10, 20]);
+        right.copy_from_slice(&[30, 40]);
+
+        // assert the changes
+        assert_eq!(&data, &[10, 20, 30, 40]);
+
+        // drop one half and ensure the other still works
+        drop(left);
+        // right is still valid
```
febo, line 913: Unfortunately, after the first drop, the mutable flag gets reset since `left` and `right` chare the same borrow state. You would expect that adding this:
```rust
// borrow state should still indicate mutable borrow
assert_eq!(state & 0b_0000_1000, 0b_0000_1000);
```
to be valid since `right` is still borrowed, but this is not the case.

febo:
Thanks for the PR, nice idea! Unfortunately it does not work since the `RefMut` has a single flag to track mutable borrows, which gets reset as soon as one of the slices is dropped.

A workaround is to rely on the borrow checker in this case – e.g.:
```rust
// Split the RefMut into two non-overlapping slices
let (left, right) = ref_mut.split_at_mut(2);

// Mutate both halves independently
left.copy_from_slice(&[10, 20]);
right.copy_from_slice(&[30, 40]);

// assert the changes on the original data
assert_eq!(&data, &[10, 20, 30, 40]);

right[0] = 99;
assert_eq!(&data, &[10, 20, 99, 40]);

// drop the mutable reference
drop(ref_mut);

// borrow state should be reset after dropping
assert_eq!(state & 0b_0000_1000, 0);
```
The borrow checked will track the lifetime of the slices returned by `split_at_mut`, so only when `ref_mut` goes out of scope the borrow state gets reset. After the `drop`, you can't use `left` and `right` references.

PR #159, opened 2025-05-25
Add const qualifier

febo:
### Problem

Creating an `AccountMeta` can be done at compile time, but the type does not have `const` qualifiers on its "constructors".

### Solution

Add `const` qualifiers.


sonicfromnewyoke (commented):
to me looks like a reasonable change. 
Also you can run `cargo clippy -- -Wclippy::missing-const-for-fn` and apply suggestions

ilya-bobyr (approved):

PR #160, opened 2025-05-26
Add get_sysvar

LStan:
This PR adds `get_sysvar` to call `sol_get_sysvar` syscall. I extracted the main logic to `get_sysvar_unchecked` so it can be called without checks if `dst` buffer is known to be created with the necessary length.

### DIFF
--- a/sdk/pinocchio/src/sysvars/mod.rs
+++ b/sdk/pinocchio/src/sysvars/mod.rs
@@ -1,6 +1,6 @@
 //! Provides access to cluster system accounts.
 
-use crate::program_error::ProgramError;
+use crate::{program_error::ProgramError, pubkey::Pubkey};
 
 pub mod clock;
 pub mod fees;
@@ -43,4 +43,50 @@ macro_rules! impl_sysvar_get {
             }
         }
     };
+}
+
+/// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
+/// syscall.
+///
+/// # Safety
+///
+/// `dst` must be large enough to hold `length` bytes.
+#[inline]
+pub unsafe fn get_sysvar_unchecked(
+    dst: &mut [u8],
+    sysvar_id: &Pubkey,
+    offset: u64,
+    length: u64,
+) -> Result<(), ProgramError> {
+    let sysvar_id = sysvar_id as *const _ as *const u8;
+    let var_addr = dst as *mut _ as *mut u8;
+
+    #[cfg(target_os = "solana")]
+    let result = unsafe { crate::syscalls::sol_get_sysvar(sysvar_id, var_addr, offset, length) };
+
+    #[cfg(not(target_os = "solana"))]
+    let result = core::hint::black_box(sysvar_id as u64 + var_addr as u64 + offset + length);
+
+    match result {
+        crate::SUCCESS => Ok(()),
+        e => Err(e.into()),
+    }
+}
+
+/// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
+/// syscall.
+#[inline]
+pub fn get_sysvar(
+    dst: &mut [u8],
+    sysvar_id: &Pubkey,
+    offset: u64,
+    length: u64,
+) -> Result<(), ProgramError> {
+    // Check that the provided destination buffer is large enough to hold the
+    // requested data.
+    if dst.len() < length as usize {
+        return Err(ProgramError::InvalidArgument);
+    }
+
+    unsafe { get_sysvar_unchecked(dst, sysvar_id, offset, length) }
 }
--- a/sdk/pinocchio/src/sysvars/mod.rs
+++ b/sdk/pinocchio/src/sysvars/mod.rs
@@ -1,6 +1,10 @@
 //! Provides access to cluster system accounts.
 
+#[cfg(target_os = "solana")]
+use crate::syscalls::sol_get_sysvar;
 use crate::{program_error::ProgramError, pubkey::Pubkey};
+#[cfg(not(target_os = "solana"))]
+use core::hint::black_box;
 
 pub mod clock;
 pub mod fees;
@@ -50,42 +54,43 @@ macro_rules! impl_sysvar_get {
 ///
 /// # Safety
 ///
-/// `dst` must point to a buffer that can hold at least `length` bytes.
+/// The caller must ensure that the `dst` pointer is valid and has enough space
+/// to hold the requested `len` bytes of data.
 #[inline]
 pub unsafe fn get_sysvar_unchecked(
     dst: *mut u8,
     sysvar_id: &Pubkey,
-    offset: u64,
-    length: u64,
+    offset: usize,
+    len: usize,
 ) -> Result<(), ProgramError> {
-    let sysvar_id = sysvar_id as *const _ as *const u8;
-
     #[cfg(target_os = "solana")]
-    let result = unsafe { crate::syscalls::sol_get_sysvar(sysvar_id, dst, offset, length) };
+    {
+        let result = unsafe {
+            sol_get_sysvar(
+                sysvar_id as *const _ as *const u8,
+                dst,
+                offset as u64,
+                len as u64,
+            )
+        };
+
+        match result {
+            crate::SUCCESS => Ok(()),
+            e => Err(e.into()),
+        }
+    }
 
     #[cfg(not(target_os = "solana"))]
-    let result = core::hint::black_box(sysvar_id as u64 + dst as u64 + offset + length);
-
-    match result {
-        crate::SUCCESS => Ok(()),
-        e => Err(e.into()),
+    {
+        black_box((dst, sysvar_id, offset, len));
+        Ok(())
     }
 }
 
 /// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
 /// syscall.
-#[inline]
-pub fn get_sysvar(
-    dst: &mut [u8],
-    sysvar_id: &Pubkey,
-    offset: u64,
-    length: u64,
-) -> Result<(), ProgramError> {
-    // Check that the provided destination buffer is large enough to hold the
-    // requested data.
-    if dst.len() < length as usize {
-        return Err(ProgramError::InvalidArgument);
-    }
-
-    unsafe { get_sysvar_unchecked(dst as *mut _ as *mut u8, sysvar_id, offset, length) }
+#[inline(always)]
+pub fn get_sysvar(dst: &mut [u8], sysvar_id: &Pubkey, offset: usize) -> Result<(), ProgramError> {
+    // SAFETY: Use the length of the slice as the length parameter.
+    unsafe { get_sysvar_unchecked(dst.as_mut_ptr(), sysvar_id, offset, dst.len()) }
 }
--- a/sdk/pinocchio/src/sysvars/mod.rs
+++ b/sdk/pinocchio/src/sysvars/mod.rs
@@ -50,22 +50,21 @@ macro_rules! impl_sysvar_get {
 ///
 /// # Safety
 ///
-/// `dst` must be large enough to hold `length` bytes.
+/// `dst` must point to a buffer that can hold at least `length` bytes.
 #[inline]
 pub unsafe fn get_sysvar_unchecked(
-    dst: &mut [u8],
+    dst: *mut u8,
     sysvar_id: &Pubkey,
     offset: u64,
     length: u64,
 ) -> Result<(), ProgramError> {
     let sysvar_id = sysvar_id as *const _ as *const u8;
-    let var_addr = dst as *mut _ as *mut u8;
 
     #[cfg(target_os = "solana")]
-    let result = unsafe { crate::syscalls::sol_get_sysvar(sysvar_id, var_addr, offset, length) };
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/sysvars/mod.rs
```diff
@@ -44,3 +44,48 @@ macro_rules! impl_sysvar_get {
...
+    length: u64,
+) -> Result<(), ProgramError> {
+    // Check that the provided destination buffer is large enough to hold the
+    // requested data.
+    if dst.len() < length as usize {
+        return Err(ProgramError::InvalidArgument);
+    }
+
+    unsafe { get_sysvar_unchecked(dst as *mut _ as *mut u8, sysvar_id, offset, length) }
+}
```
illia-bobyr: `dst` slice already contains the length.
And if someone wants to use a different offset, they can just create a new slice, that points to the desired location.
I think this would be a more canonical use of the slices:

```suggestion
pub fn get_sysvar(
    dst: &mut [u8],
    sysvar_id: &Pubkey,
) -> Result<(), ProgramError> {
    unsafe { get_sysvar_unchecked(dst as *mut _ as *mut u8, sysvar_id, dst.len()) }
}
```

I also do not understand why `offset` is used when `dst` is a pointer.
So I have removed it from the call.  It should just always be `0`.
As one can always create a slice that has an offset like this:

```rust
    let dst_with_offset = &mut dst[offset..];
```

sdk/pinocchio/src/sysvars/mod.rs
```diff
@@ -44,3 +44,48 @@ macro_rules! impl_sysvar_get {
...
+    let result = unsafe { crate::syscalls::sol_get_sysvar(sysvar_id, dst, offset, length) };
+
+    #[cfg(not(target_os = "solana"))]
+    let result = core::hint::black_box(sysvar_id as u64 + dst as u64 + offset + length);
+
+    match result {
+        crate::SUCCESS => Ok(()),
+        e => Err(e.into()),
+    }
+}
```
illia-bobyr: It seems `offset` is redundant here, no?
One can always shift the `dst` pointer to point to whatever location they want to use during the function invocation.

```suggestion
/// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
/// syscall.
///
/// # Safety
///
/// `dst` must point to a buffer that can hold at least `length` bytes.
#[inline]
pub unsafe fn get_sysvar_unchecked(
    dst: *mut u8,
    length: u64,
    sysvar_id: &Pubkey,
) -> Result<(), ProgramError> {
    let sysvar_id = sysvar_id as *const _ as *const u8;

    #[cfg(target_os = "solana")]
    let result = unsafe { crate::syscalls::sol_get_sysvar(sysvar_id, dst, 0, length) };

    #[cfg(not(target_os = "solana"))]
    let result = core::hint::black_box(sysvar_id as u64 + dst as u64 + length);

    match result {
        crate::SUCCESS => Ok(()),
        e => Err(e.into()),
    }
}
```

Also, addition of pointers is a rather strange operation.
And construction of a `ProgramError` instance from a pointer value is also quite unexpected.

```suggestion
/// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
/// syscall.
///
/// # Safety
///
/// `dst` must point to a buffer that can hold at least `length` bytes.
#[inline]
pub unsafe fn get_sysvar_unchecked(
    dst: *mut u8,
    length: u64,
    sysvar_id: &Pubkey,
) -> Result<(), ProgramError> {
    let sysvar_id = sysvar_id as *const _ as *const u8;

    #[cfg(target_os = "solana")]
    {
        let result = crate::syscalls::sol_get_sysvar(sysvar_id, dst, 0, length);
        match result {
            crate::SUCCESS => Ok(()),
            e => Err(e.into()),
        }
    }

    #[cfg(not(target_os = "solana"))]
    {
        let _ = core::hint::black_box(dst);
        let _ = core::hint::black_box(length);
        let _ = core::hint::black_box(sysvar_id);
        Err(ProgramError::Custom(0x656e76 /* "env" */))
    }
}
```

It would be nice to have a dedicated error in the `ProgramError` enum for the case when the code is executed in an invalid environment.

sdk/pinocchio/src/sysvars/mod.rs
```diff
@@ -44,3 +44,48 @@ macro_rules! impl_sysvar_get {
...
+pub unsafe fn get_sysvar_unchecked(
+    dst: *mut u8,
+    sysvar_id: &Pubkey,
+    offset: u64,
+    length: u64,
+) -> Result<(), ProgramError> {
+    let sysvar_id = sysvar_id as *const _ as *const u8;
+
+    #[cfg(target_os = "solana")]
+    let result = unsafe { crate::syscalls::sol_get_sysvar(sysvar_id, dst, offset, length) };
```
illia-bobyr: As `get_sysvar_unchecked()` is marked `unsafe` this inner `unsafe` block has no meaning, I think.

```suggestion
    let result = crate::syscalls::sol_get_sysvar(sysvar_id, dst, offset, length);
```

sdk/pinocchio/src/sysvars/mod.rs
```diff
@@ -44,3 +44,48 @@ macro_rules! impl_sysvar_get {
...
+
+    #[cfg(not(target_os = "solana"))]
+    let result = core::hint::black_box(sysvar_id as u64 + dst as u64 + offset + length);
```
illia-bobyr: As this is a function call, and not a macro, please import all the dependencies at the top of the module, and use reasonably short names in the function body:

```rust
use core::hint::black_box;

use crate::{
    /* ... */,
    crate::syscalls::sol_get_sysvar,
    /* ... */,
};
```

```suggestion
    #[cfg(target_os = "solana")]
    let result = unsafe { sol_get_sysvar(sysvar_id, dst, offset, length) };

    #[cfg(not(target_os = "solana"))]
    let result = black_box(sysvar_id as u64 + dst as u64 + offset + length);
```

sdk/pinocchio/src/sysvars/mod.rs
LStan: It mimics the function from [solana-sdk](https://github.com/anza-xyz/solana-sdk/blob/b0b47b74f1d2b37424598d0cc3f2a7d8bb7fa85b/sysvar/src/lib.rs#L217).
Do you suggest creating a buffer with a size larger than necessary? `offset` can be much larger than the stack size

sdk/pinocchio/src/sysvars/mod.rs
LStan: Maybe it is better to panic like [here](https://github.com/anza-xyz/pinocchio/blob/9ba554088d5c5ccf231d765dc99d452026502503/sdk/pinocchio/src/pubkey.rs#L198) ?

sdk/pinocchio/src/sysvars/mod.rs
LStan: It will be a warning in [rust 2024](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html)

sdk/pinocchio/src/sysvars/mod.rs
LStan: That is how it is done in other pinocchio functions: [here](https://github.com/anza-xyz/pinocchio/blob/9ba554088d5c5ccf231d765dc99d452026502503/sdk/pinocchio/src/pubkey.rs#L180), [here](https://github.com/anza-xyz/pinocchio/blob/9ba554088d5c5ccf231d765dc99d452026502503/sdk/pinocchio/src/log.rs#L99), [here](https://github.com/anza-xyz/pinocchio/blob/9ba554088d5c5ccf231d765dc99d452026502503/sdk/pinocchio/src/cpi.rs#L226), etc.

sdk/pinocchio/src/sysvars/mod.rs
illia-bobyr: Yes, panic is the right choice here, as this is the other functions do.

sdk/pinocchio/src/sysvars/mod.rs
illia-bobyr: I do not know why the SDK does it this way.
I think this is just the basics of how slices work.
Each slice is a pointer and a length.
You can easily create a new slice from an existing one, with the offset you want.

Why would you need to create a larger buffer?
`dst` is a pointer, not a buffer.
You can create slices from existing slices with an offset, as shown above, and it is a cheap operation.
Which makes the `offset` argument redundant.

I do not see a reason to copy an overly complicated API from the SDK.
But I could be missing something.

sdk/pinocchio/src/sysvars/mod.rs
illia-bobyr: I personally do not like it, as it breaks consistency.
There are now two ways to call any function: either directly by name, or via an absolute path.
Plus, your import block at the top is incomplete now as well.
I see it as an unnecessary complication.

In the other PRs that I review, I ask for this style to be avoided.
But it is the case that the code in the solana/agave repo is not super consistent.

So, I would say that the other locations should be changed to avoid absolute references in function calls as well.
But it is quite possible that other reviewers would approve this PR as is.

sdk/pinocchio/src/sysvars/mod.rs
febo: The `offset` argument refers to the offset on the sysvar account, so it is required. Agree that we don't need the `length`, which also means we don't need the `get_sysvar_unchecked` variant – we can use `dst.len()` directly.

sdk/pinocchio/src/sysvars/mod.rs
febo: We tend to return a dummy value currently – this is not meant to be used off-chain and it is there to please clippy.

sdk/pinocchio/src/sysvars/mod.rs
febo: > The `offset` argument refers to the offset on the sysvar account, so it is required. Agree that we don't need the `length`, which also means we don't need the `get_sysvar_unchecked` variant – we can use `dst.len()` directly.

We probably want an `unchecked` variant to support retrieving data with an uninitialized buffer.

sdk/pinocchio/src/sysvars/mod.rs
febo: Agree with @ilya-bobyr, let's start adding imports whenever is possible.

sdk/pinocchio/src/sysvars/mod.rs
febo: This is a suggestion:
```rust
/// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
/// syscall.
#[inline(always)]
pub fn get_sysvar(dst: &mut [u8], sysvar_id: &Pubkey, offset: usize) -> Result<(), ProgramError> {
    // SAFETY: Use the length of the slice as the length parameter.
    unsafe { get_sysvar_unchecked(dst.as_mut_ptr(), sysvar_id, offset, dst.len()) }
}

/// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
/// syscall.
///
/// # Safety
///
/// The caller must ensure that the `dst` pointer is valid and has enough space
/// to hold the requested `len` bytes of data.
#[inline]
pub unsafe fn get_sysvar_unchecked(
    dst: *mut u8,
    sysvar_id: &Pubkey,
    offset: usize,
    len: usize,
) -> Result<(), ProgramError> {
    #[cfg(target_os = "solana")]
    {
        let result = unsafe {
            sol_get_sysvar(
                sysvar_id as *const _ as *const u8,
                dst,
                offset as u64,
                len as u64,
            )
        };

        match result {
            crate::SUCCESS => Ok(()),
            e => Err(e.into()),
        }
    }

    {
        black_box((dst, sysvar_id, offset, len));
        Ok(())
    }
}
```

illia-bobyr: My bad.  I misunderstood what `offset` is for and did not check the API.
Thank you, Fernando, for looking into it.

LStan: @febo [here](https://github.com/anza-xyz/pinocchio/blob/9ba554088d5c5ccf231d765dc99d452026502503/sdk/pinocchio/src/pubkey.rs#L198) there is a `panic` and not `Ok`. Should I panic here, or should that code be changed to return `Ok` for consistency?

febo: Let's return `Ok` – I think that is the only case that we have a panic. 😅

LStan: [here](https://github.com/anza-xyz/pinocchio/blob/9ba554088d5c5ccf231d765dc99d452026502503/sdk/pinocchio/src/pubkey.rs#L198) there is panic and [here](https://github.com/anza-xyz/pinocchio/blob/9ba554088d5c5ccf231d765dc99d452026502503/sdk/pinocchio/src/pubkey.rs#L145) `None` is returned. So there will be panic, None and Ok for the same thing. Should I change them to be the same?

febo: Let's do that in a separate PR – we can then decide to either always panic or return dummy values.

PR #161, opened 2025-05-26
Add create_with_seed

LStan:
This PR adds `create_with_seed` function. It mimics the same function from `solana-pubkey`, except that `seed` is `&[u8]` instead of `&str`.


sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +232,54 @@ pub fn checked_create_program_address(
...
+
+    #[cfg(not(target_os = "solana"))]
+    {
+        core::hint::black_box((base, seed, owner));
+        panic!("create_with_seed is only available on target `solana`")
+    }
+}
+
+/// Derive a Pubkey from another Pubkey, seed, and a program id.
+#[inline]
```
febo, line 274: This one can always be inlined.

sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +232,54 @@ pub fn checked_create_program_address(
 
     create_program_address(seeds, program_id)
 }
+
+/// Derive a Pubkey from another Pubkey, seed, and a program id.
+/// This function does *not* validate whether the given `seed` is within
+/// the valid length or not.
```
febo, line 238: Let's move this comment to a "Safety" section.

```suggestion
/// # Safety
///
/// This function does not validate whether the given `seed` is within the valid length or not.
```

sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +232,54 @@ pub fn checked_create_program_address(
...
+#[inline]
+pub fn create_with_seed_unchecked(
```
febo, line 240: Missing `unsafe` qualifier.

sdk/pinocchio/src/pubkey.rs
LStan: This is analogous to `create_program_address` which also does not validate whether the given `seeds` are within
the valid length or not. The comment there is not in a "Safety" section

sdk/pinocchio/src/pubkey.rs
LStan: Can it produce UB? This is analogous to `create_program_address` which is not `unsafe`

sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +232,54 @@ pub fn checked_create_program_address(
...
+    seed: &[u8],
+    owner: &Pubkey,
+) -> Result<Pubkey, ProgramError> {
+    #[cfg(target_os = "solana")]
+    {
+        let mut bytes = core::mem::MaybeUninit::<[u8; PUBKEY_BYTES]>::uninit();
+
+        let vals = &[base, seed, owner];
+
+        let result = unsafe {
```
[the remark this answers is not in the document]
LStan: I don't know how I missed it. I'll add it.
Should I add `if owner.len() >= PDA_MARKER.len()` part? `owner` len is 32 and `PDA_MARKER` is 21, so this check is always true

sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +234,49 @@ pub fn checked_create_program_address(
...
+            crate::syscalls::sol_sha256(
+                vals as *const _ as *const u8,
+                vals.len() as u64,
+                bytes.as_mut_ptr() as *mut _,
+            )
+        };
```
febo: The syscall always returns `0` when it succeed; otherwise it aborts the vm. There is no need to handle the return.
```suggestion
        unsafe {
            crate::syscalls::sol_sha256(
                vals as *const _ as *const u8,
                vals.len() as u64,
                bytes.as_mut_ptr() as *mut _,
            );
        }
```

sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +234,49 @@ pub fn checked_create_program_address(
...
+
+        match result {
+            // SAFETY: The syscall has initialized the bytes.
+            crate::SUCCESS => Ok(unsafe { bytes.assume_init() }),
+            _ => Err(result.into()),
+        }
```
febo:
```suggestion
        // SAFETY: The syscall has initialized the bytes.
        Ok(unsafe { bytes.assume_init() })
```

sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +234,49 @@ pub fn checked_create_program_address(
...
     create_program_address(seeds, program_id)
 }
+
+/// Derive a Pubkey from another Pubkey, seed, and a program id.
+#[inline]
+pub fn create_with_seed(
+    base: &Pubkey,
+    seed: &[u8],
+    owner: &Pubkey,
+) -> Result<Pubkey, ProgramError> {
```
illia-bobyr: There is a mismatch between the documentation text and the argument names.
The documentation lists "another Pubkey", "seed", and a "program id", but the argument names are `base`, `seed`, and `owner`.

While this matches the argument names in the `Pubkey::create_with_seed` it still seems inconsistent.
Would it make sense to rename `owner` to `program_id`?

sdk/pinocchio/src/pubkey.rs
```diff
@@ -232,3 +234,49 @@ pub fn checked_create_program_address(
...
+    if seed.len() > MAX_SEED_LEN {
+        return Err(ProgramError::MaxSeedLengthExceeded);
+    }
+
+    if owner.len() >= PDA_MARKER.len() {
+        let slice = &owner[owner.len() - PDA_MARKER.len()..];
+        if slice == PDA_MARKER {
+            return Err(ProgramError::IllegalOwner);
+        }
+    }
```
illia-bobyr: Is this the same as `ends_with()`?

```suggestion
    if owner.ends_with(PDA_MARKER) {
        return Err(ProgramError::IllegalOwner);
    }
```

sdk/pinocchio/src/pubkey.rs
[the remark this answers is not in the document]
illia-bobyr: The idea is to prevent you from using `create_with_seed()` to compute a program derived address.
As all the bytes provided to the hash function are concatenated, if you pass in an `onwer` that ends with the `PDA_MARKER` value, you would be able to turn `create_with_seed()` into `create_program_address()`:

https://github.com/anza-xyz/solana-sdk/blob/1276772ee61fbd1f8a60cfec7cd553aa4f6a55f3/pubkey/src/lib.rs#L928-L932

You need to put the first part of the desired `program_id` into the `seed`, and the second part in `owner`.  Something like

```rust
  let pad_address = create_with_seed(
      /* base */  b"<concatenate 32 bytes of the normal seeds>",
      /* seeds */ [b"rest of the seeds", b"<first 21 bytes of the program id>"],
      /* owner */ b"<last 11 bytes of the program id>ProgramDerivedAddress",
  );
```

I think both `owner.len()` and `PDA_MARKER.len()` are known at compilation time.
So the compiler should be able to remove the check automatically.
But not having the check at all, is a bit unsafe.

sdk/pinocchio/src/pubkey.rs
```diff
@@ -237,3 +237,43 @@ pub fn checked_create_program_address(
...
+
+    #[cfg(target_os = "solana")]
+    {
+        let mut bytes = core::mem::MaybeUninit::<[u8; PUBKEY_BYTES]>::uninit();
+
+        let vals = &[base, seed, program_id];
+
+        unsafe {
+            crate::syscalls::sol_sha256(
+                vals as *const _ as *const u8,
```
illia-bobyr, line 264: *minor*

The double cast is probably unnecessary.

```suggestion
                vals as *const u8,
```

LStan:
```
mismatched types
expected `&[u8]`, found `u8`
casting `&[&[u8]; 3]` as `*const u8` is invalid`
```

febo (approved):

PR #162, opened 2025-05-26
feat: add`log_cu_usage` proc_macro_attribute

sonicfromnewyoke:
### Problem
there is hard to investigate CU usage of functions in the program

### Solution
add `proc_macro_attribute` that helps in CU calculation based on this [guide on solana.com](https://solana.com/it/developers/guides/advanced/how-to-optimize-compute)

### DIFF
--- a/sdk/log/macro/src/lib.rs
+++ b/sdk/log/macro/src/lib.rs
@@ -258,11 +258,11 @@ pub fn log(input: TokenStream) -> TokenStream {
 ///     // Function body
 /// }
 /// ```
-/// 
+///
 /// logging output will look like:
-/// 
+///
 /// "Program log: Function my_function consumed 36 compute units"
-/// 
+///
 /// # References
 ///
 /// * [Logging syscall](https://github.com/anza-xyz/agave/blob/d88050cda335f87e872eddbdf8506bc063f039d3/programs/bpf_loader/src/syscalls/logging.rs#L70)
--- a/sdk/log/macro/src/lib.rs
+++ b/sdk/log/macro/src/lib.rs
@@ -294,13 +294,15 @@ pub fn compute_fn(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let block = &input.block;
 
     input.block = syn::parse_quote!({
-        ::pinocchio::msg!(concat!(stringify!(#fn_name), " {{"));
-        ::pinocchio::log::sol_log_compute_units();
+        let cu_before =  ::pinocchio::log::sol_remaining_compute_units();
 
         let __result = (|| #block)();
 
-        ::pinocchio::log::sol_log_compute_units();
-        ::pinocchio::msg!(concat!("}} // ", stringify!(#fn_name)));
+        let cu_after =  ::pinocchio::log::sol_remaining_compute_units();
+        let empty_log_cost = 102;
+        let diff = cu_before - cu_after - empty_log_cost;
+
+        ::pinocchio_log::log!( "Function {} consumed {} compute units", stringify!(#fn_name), diff);
 
         __result
     });
--- a/sdk/log/macro/src/lib.rs
+++ b/sdk/log/macro/src/lib.rs
@@ -253,7 +253,7 @@ pub fn log(input: TokenStream) -> TokenStream {
 ///  # Example
 ///
 /// ```rust,ignore
-/// #[pinocchio_log::compute_fn]
+/// #[pinocchio_log::log_cu_usage]
 /// fn my_function() {
 ///     // Function body
 /// }
@@ -269,17 +269,17 @@ pub fn log(input: TokenStream) -> TokenStream {
 /// * [Compute budget](https://github.com/anza-xyz/agave/blob/d88050cda335f87e872eddbdf8506bc063f039d3/program-runtime/src/compute_budget.rs#L150)
 ///
 #[proc_macro_attribute]
-pub fn compute_fn(_attr: TokenStream, item: TokenStream) -> TokenStream {
+pub fn log_cu_usage(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut input = parse_macro_input!(item as ItemFn);
     let fn_name = &input.sig.ident;
     let block = &input.block;
 
     input.block = syn::parse_quote!({
-        let cu_before =  ::pinocchio::log::sol_remaining_compute_units();
+        let cu_before = unsafe { ::pinocchio_log::logger::remaining_compute_units() };
 
         let __result = (|| #block)();
 
-        let cu_after =  ::pinocchio::log::sol_remaining_compute_units();
+        let cu_after = unsafe { ::pinocchio_log::logger::remaining_compute_units() };
         let introspection_cost = 102; // 100 - compute budget syscall_base_cost,  2 - extra calculations
 
         let consumed = cu_before - cu_after - introspection_cost;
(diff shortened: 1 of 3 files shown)


sonicfromnewyoke:
@febo wdyt about this helper-function in the `pinocchio-log` crate?

sdk/log/macro/src/lib.rs
```diff
@@ -235,3 +235,75 @@ pub fn log(input: TokenStream) -> TokenStream {
...
+///  - https://github.com/anza-xyz/agave/blob/d88050cda335f87e872eddbdf8506bc063f039d3/program-runtime/src/compute_budget.rs#L150
+#[proc_macro_attribute]
+pub fn compute_fn(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut input = parse_macro_input!(item as ItemFn);
+    let fn_name = &input.sig.ident;
+    let block = &input.block;
+
+    input.block = syn::parse_quote!({
+        ::pinocchio::msg!(concat!(stringify!(#fn_name), " {{"));
+        ::pinocchio::log::sol_log_compute_units();
```
febo: Perhaps we could use the `sol_remaining_compute_units` syscall here so we can do all the calculation on the macro and print only how much compute was used by the block.

febo (commented):
This is a nice addition to the log crate, but maybe we can improve the CU consumption calculation – left a comment about this.

sdk/log/macro/src/lib.rs
```diff
@@ -235,3 +235,77 @@ pub fn log(input: TokenStream) -> TokenStream {
...
+///  "Program log: Program log: }} // process_instruction",
+///  "Program EMPTY_WITH_LOG consumed 445 of 200000 compute units",
+///  "Program EMPTY_WITH_LOG success"
+///
+///  EMPTY where nothing happens at all.
+///  TOTAL COMPUTE UNITS USED: 36
+///  
+///  "Program EMPTY invoke [1]",
+///  "Program EMPTY consumed 36 of 200000 compute units",
+///  "Program EMPTY success"
```
illia-bobyr: Would be nice to update this note to match the new code.

sdk/log/macro/src/lib.rs
```diff
@@ -235,3 +235,77 @@ pub fn log(input: TokenStream) -> TokenStream {
...
+    let fn_name = &input.sig.ident;
+    let block = &input.block;
+
+    input.block = syn::parse_quote!({
+        let cu_before =  ::pinocchio::log::sol_remaining_compute_units();
+
+        let __result = (|| #block)();
+
+        let cu_after =  ::pinocchio::log::sol_remaining_compute_units();
+        let empty_log_cost = 102;
```
illia-bobyr: It would be great to have a comment explaining how `102` was calculated.
For example, it could be useful if someone would want to change this code and would need to update the constant.

sdk/log/macro/src/lib.rs
sonicfromnewyoke: yeah, prev commit is marked as WIP. Will ask you for the recheck after actualizing the doc

sonicfromnewyoke:
example of cpi call to pump-fun buy ix
```
Program 6EF8rrecthR5Dkzon8Nwu78hRvfCKubJ14M5uBEwF6P invoke [2]
Program log: Instruction: Buy
....
Program 6EF8rrecthR5Dkzon8Nwu78hRvfCKubJ14M5uBEwF6P success
Program log: Function buy consumed 45518 compute units
```

also would like to ask you, are you happy with `compute_fn` naming, or maybe we can come up with something better?

illia-bobyr:
> also would like to ask you, are you happy with `compute_fn` naming, or maybe we can come up with something better?

I think it makes sense to name it from the standpoint of someone using it.
In other words, as a user of this macro, with the goal of seeing the function CU usage, what is a reasonable way to request this kind of information?
I would suggest, maybe, "log_cu_usage":

```rust
#[pinocchio_log::log_cu_usage]
fn my_function() {
    // Function body
}
```

It is tempting to remove the duplicate `log_` prefix:

```rust
#[pinocchio_log::cu_usage]
fn my_function() {
    // Function body
}
```

But if the macro is imported, then the `pinocchio_log` prefix is gone, and the `log_` prefix is meaningful:

```rust
#[log_cu_usage]
fn my_function() {
    // Function body
}
```

I personally prefer to import things, and avoid absolute paths, so I would prefer a version with the `log_` prefix.

Maybe @febo has an opinion here, as I'm actually only partially involved in the Pinocchio development.

sdk/log/macro/src/lib.rs
```diff
@@ -235,3 +235,59 @@ pub fn log(input: TokenStream) -> TokenStream {
...
+/// * [Compute budget](https://github.com/anza-xyz/agave/blob/d88050cda335f87e872eddbdf8506bc063f039d3/program-runtime/src/compute_budget.rs#L150)
+///
+#[proc_macro_attribute]
+pub fn compute_fn(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut input = parse_macro_input!(item as ItemFn);
+    let fn_name = &input.sig.ident;
+    let block = &input.block;
+
+    input.block = syn::parse_quote!({
+        let cu_before =  ::pinocchio::log::sol_remaining_compute_units();
```
febo: `pinocchio_log` can be used in any program and currently does not depend on pinocchio. We need to add the syscall definition at the top the `logger.rs` file instead. This duplication will eventually go away when we integrate the SDK, but for now it is needed.

I can't comment directly there, but you will need to add:
```rust
pub fn sol_remaining_compute_units() -> u64;

pub fn sol_remaining_compute_units() -> u64 {
    let syscall: extern "C" fn() -> u64 = unsafe { core::mem::transmute(3991886574u64) }; // murmur32 hash of "sol_remaining_compute_units"
    syscall(dest, src, n)
}
```

sdk/pinocchio/src/log.rs
```diff
@@ -165,3 +165,13 @@ pub fn sol_log_compute_units() {
         crate::syscalls::sol_log_compute_units_();
     }
 }
+#[inline]
```
febo: This change can be reverted.

febo:
> > also would like to ask you, are you happy with `compute_fn` naming, or maybe we can come up with something better?
> 
> I think it makes sense to name it from the standpoint of someone using it. In other words, as a user of this macro, with the goal of seeing the function CU usage, what is a reasonable way to request this kind of information? I would suggest, maybe, "log_cu_usage":
> 
> ```rust
> #[pinocchio_log::log_cu_usage]
> fn my_function() {
>     // Function body
> }
> ```
> 
> It is tempting to remove the duplicate `log_` prefix:
> 
> ```rust
> #[pinocchio_log::cu_usage]
> fn my_function() {
>     // Function body
> }
> ```
> 
> But if the macro is imported, then the `pinocchio_log` prefix is gone, and the `log_` prefix is meaningful:
> 
> ```rust
> #[log_cu_usage]
> fn my_function() {
>     // Function body
> }
> ```
> 
> I personally prefer to import things, and avoid absolute paths, so I would prefer a version with the `log_` prefix.
> 
> Maybe @febo has an opinion here, as I'm actually only partially involved in the Pinocchio development.

This suggestion is good, the name `log_cu_usage` is very adequate.

sdk/log/macro/src/lib.rs
sonicfromnewyoke: done

sdk/pinocchio/src/log.rs
sonicfromnewyoke: moved into the logger to give access to the private syscalls of its module from the macro space

ilya-bobyr (approved):
Looks good to me!
But maybe @febo should approve as well.

PR #163, opened 2025-05-26
Add EpochRewards sysvar type

LStan:
This PR adds EpochRewards sysvar type. I copied it from solana-sdk. Not sure about `total_points` because it is `u128`. It will work with `get`, but if `from_account_info` is implemented it can cause UB. Probably it should be changed to [u8; 16].

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,6 +34,7 @@ dependencies = [
  "solana-account-view",
  "solana-address",
  "solana-define-syscall",
+ "solana-hash",
  "solana-instruction-view",
  "solana-program-error",
 ]
@@ -113,6 +114,12 @@ name = "solana-define-syscall"
 version = "4.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "57e5b1c0bc1d4a4d10c88a4100499d954c09d3fecfae4912c1a074dff68b1738"
+
+[[package]]
+name = "solana-hash"
+version = "4.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a5d48a6ee7b91fc7b998944ab026ed7b3e2fc8ee3bc58452644a86c2648152f"
 
 [[package]]
 name = "solana-instruction-view"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ pinocchio = { version = "0.9", default-features = false, path = "sdk" }
 solana-account-view = "1.0"
 solana-address = "2.0"
 solana-define-syscall = "4.0"
+solana-hash = "4.0"
 solana-instruction-view = "1.0"
 solana-program-error = "3.0"
 
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -23,13 +23,14 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 
 [features]
 alloc = ["solana-instruction-view?/slice-cpi"]
-copy = ["solana-account-view/copy", "solana-address/copy"]
+copy = ["solana-account-view/copy", "solana-address/copy", "solana-hash/copy"]
 cpi = ["dep:solana-instruction-view"]
 default = ["alloc"]
 
 [dependencies]
 solana-account-view = { workspace = true }
 solana-address = { workspace = true, features = ["syscalls"] }
+solana-hash = { workspace = true }
 solana-instruction-view = { workspace = true, features = ["cpi"], optional = true }
 solana-program-error = { workspace = true }
 
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -258,6 +258,10 @@ pub use solana_address::Address;
 #[cfg(any(target_os = "solana", target_arch = "bpf"))]
 pub use solana_define_syscall::definitions as syscalls;
 
+// Re-export the `solana_hash` for downstream use.
+pub use solana_hash as hash;
+pub use solana_hash::Hash;
+
 // Re-export the `solana_instruction_view` for downstream use.
 #[cfg(feature = "cpi")]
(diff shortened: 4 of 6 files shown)


sdk/pinocchio/src/sysvars/epoch_rewards.rs
```diff
@@ -0,0 +1,50 @@
...
+pub const HASH_BYTES: usize = 32;
+
+#[derive(Default, Debug, Clone)]
+pub struct Hash([u8; HASH_BYTES]);
+
+impl Hash {
+    pub fn to_bytes(self) -> [u8; HASH_BYTES] {
+        self.0
+    }
+}
```
illia-bobyr: I'm not an expert in how things are organized in Pinocchio, but it seems strange to define the blockhash type inside `sysvars::epoch_rewards`.

sdk/pinocchio/src/sysvars/epoch_rewards.rs
```diff
@@ -0,0 +1,50 @@
+use super::Sysvar;
+use crate::impl_sysvar_get;
+
+pub const HASH_BYTES: usize = 32;
+
+#[derive(Default, Debug, Clone)]
+pub struct Hash([u8; HASH_BYTES]);
```
illia-bobyr: Should this also be `repr(C)` if it is used inside a `repr(C)` struct?

sdk/pinocchio/src/sysvars/epoch_rewards.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl Hash {
+    pub fn to_bytes(self) -> [u8; HASH_BYTES] {
+        self.0
+    }
+}
+
+#[repr(C, align(16))]
```
illia-bobyr: *minor*

This `align` attribute is redundant:
> The alignment of the struct is the alignment of the most-aligned field in it.

https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.c.struct.align

sdk/pinocchio/src/sysvars/epoch_rewards.rs
LStan: Maybe I should just remove it completely and change `parent_blockhash: Hash` to `parent_blockhash: [u8; 32]` ?

sdk/pinocchio/src/sysvars/epoch_rewards.rs
LStan: It is `repr(transparent)` in [solana-sdk](https://github.com/anza-xyz/solana-sdk/blob/b0b47b74f1d2b37424598d0cc3f2a7d8bb7fa85b/hash/src/lib.rs#L52)

sdk/pinocchio/src/sysvars/epoch_rewards.rs
LStan: I took it from [solana-sdk](https://github.com/anza-xyz/solana-sdk/blob/b0b47b74f1d2b37424598d0cc3f2a7d8bb7fa85b/epoch-rewards/src/lib.rs#L22)

sdk/pinocchio/src/sysvars/epoch_rewards.rs
illia-bobyr: Either `repr(transparent)` or `repr(C)` would work here, I think.
`repr(transparent)` is more reasonable if the internal state is also public.
This way, the user can see what layout is being used in the docs, for example.

So, maybe this:

```suggestion
#[derive(Default, Debug, Clone)]
#[repr(transparent)]
pub struct Hash(pub [u8; HASH_BYTES]);
```

I think any 32 bytes are a valid hash.  So it seems fine to expose `[u8; HASH_BYTES]` I guess?

The SDK folks will probably want to look at this change as well.

@joncinque You are the expert here, right?
Or you can probably get someone with the SDK expertise to double-check the approach here.

sdk/pinocchio/src/sysvars/epoch_rewards.rs
illia-bobyr: Having a dedicated type for the hashes seems quite reasonable.
Not sure what is the state of the Pinocchio coverage here.
There should be a `Hash` type at some point, I would think.

sdk/pinocchio/src/sysvars/epoch_rewards.rs
illia-bobyr: Still is redundant.  I do not have a strong opinion either way.
I see no reason to have it, so I would lean towards removing it.
But it is unlikely that the content of the struct changes to make this alignment overly aggressive, right?

sdk/pinocchio/src/sysvars/epoch_rewards.rs
joncinque: We avoided exposing the hash bytes publicly for the base hash type in the sdk, which makes we wary of exposing it here: https://github.com/anza-xyz/solana-sdk/blob/965e2a8a63bcc730ee6c001a0e66e3314bb924d6/hash/src/lib.rs#L53

We should definitely go with `repr(transparent)` though.

@febo do you have any opinions about a generic hash type for pinocchio? Did you want to create newtype, a type alias, or eventually use `solana_hash::Hash`? I'm hoping we can go with the last option, but we might have to remove the `solana_sanitize` dependency first.

febo: @LStan I think it will be better to mimic the SDK type here – so avoid exposing the hash bytes:
```rust
[repr(transparent)]
pub struct Hash(pub(crate) [u8; HASH_BYTES]);
```
We eventually will use the SDK type so it will be easier to "swap" the type if they follow the same implementation.

sdk/pinocchio/src/sysvars/epoch_rewards.rs
febo: Agree, let's remove the unnecessary `align(16)` attribute.

PR #165, opened 2025-05-30
system: Add inline to system instructions

febo:
### Problem

The `invoke_signed` method on system instruction is not inlined, which makes them use an "extra" `~120` CUs.

### Solution

Add `inline(always)` to each instruction to improve CU utilization.

### DIFF
--- a/sdk/pinocchio/src/cpi.rs
+++ b/sdk/pinocchio/src/cpi.rs
@@ -13,22 +13,6 @@ use crate::{
 /// Maximum number of accounts that can be passed to a cross-program invocation.
 pub const MAX_CPI_ACCOUNTS: usize = 64;
 
-/// A constant to indicate that an account is read-only when passed as
-/// an account priviledge.
-pub const READONLY: bool = false;
-
-/// A constant to indicate that an account is writable when passed as
-/// an account priviledge.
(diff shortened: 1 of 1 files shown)


joncinque (dismissed):
It's too bad that this is needed, but :+1: from me!

rustopian (approved):

sonicfromnewyoke (commented):
due to more aggressive inlining  i run into `LLVM ERROR: Branch target out of insn range`.

solved it with remove of `inlinen(always)` in the function that invokes `system-transfer` in my program.

not something crucial, just inform you. I believe everybody who decides to use  overoptimized program-framework can tackle it.

febo:
> due to more aggressive inlining i run into `LLVM ERROR: Branch target out of insn range`.
> 
> solved it with remove of `inlinen(always)` in the function that invokes `system-transfer` in my program.
> 
> not something crucial, just inform you. I believe everybody who decides to use overoptimized program-framework can tackle it.

Yeah, it is a bit unfortunate. If we don't have the inline here, then there is no way to inline it on client code. So it seems a "good" trade-off to be more aggressive on the inlining on the helper, since the client code can always wrap the call in a non-inlined function.

PR #166, opened 2025-06-02
Simplify program entrypoint

febo:
### Problem

Currently the program entrypoint includes logic to support a custom maximum number of accounts. There is no much benefit of doing that apart from reducing the stack memory requirement, which in most cases it is not a problem.

Additionally, in order to track account borrows it zeroes the `NON_DUP_MARKER` value for each account so that borrows are represented by `1` bits – `NON_DUP_MARKER` has a `255` value.

### Solution

This PR simplifies the entrypoint by:
* Statically allocating space to process `MAX_TX_ACCOUNTS` accounts, which avoid the need for the logic to skip unprocessed accounts. Note that it is still possible to indicate a maximum number to control how much stack memory the entrypoint uses.
* Inverting the account borrow logic to represent borrows as `0` bits. This avoids having to set the `NON_DUP_MARKER` to `0` on each account.

Both of these changes contribute to minimize CUs and program binary size.

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -732,7 +732,7 @@ impl<T: ?Sized> Drop for RefMut<'_, T> {
 
 #[cfg(test)]
 mod tests {
-    use core::mem::MaybeUninit;
+    use core::mem::{size_of, MaybeUninit};
 
     use crate::NON_DUP_MARKER;
 
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -193,7 +193,10 @@ impl AccountInfo {
     #[inline(always)]
     pub fn is_borrowed(&self, state: BorrowState) -> bool {
         let borrow_state = unsafe { (*self.raw).borrow_state };
-        borrow_state & (state as u8) != 0
+        let mask = state as u8;
+        // If borrow state has any of the state bits of the mask not set,
+        // then the account is borrowed for that state.
+        (borrow_state & mask) != mask
     }
 
     /// Returns a read-only reference to the lamports in the account.
@@ -250,7 +253,7 @@ impl AccountInfo {
 
         let borrow_state = unsafe { &mut (*self.raw).borrow_state };
         // increment the immutable borrow count
-        *borrow_state += 1 << LAMPORTS_SHIFT;
+        *borrow_state -= 1 << LAMPORTS_SHIFT;
 
         // return the reference to lamports
         Ok(Ref {
@@ -269,7 +272,7 @@ impl AccountInfo {
 
         let borrow_state = unsafe { &mut (*self.raw).borrow_state };
         // set the mutable lamport borrow flag
-        *borrow_state |= 0b_1000_0000;
+        *borrow_state &= 0b_0111_1111;
 
         // return the mutable reference to lamports
         Ok(RefMut {
@@ -295,12 +298,12 @@ impl AccountInfo {
         let borrow_state = unsafe { (*self.raw).borrow_state };
 
         // check if mutable borrow is already taken
-        if borrow_state & 0b_1000_0000 != 0 {
+        if borrow_state & 0b_1000_0000 != 0b_1000_0000 {
             return Err(ProgramError::AccountBorrowFailed);
         }
 
         // check if we have reached the max immutable borrow count
-        if borrow_state & 0b_0111_0000 == 0b_0111_0000 {
+        if borrow_state & 0b_0111_0000 == 0 {
             return Err(ProgramError::AccountBorrowFailed);
         }
 
@@ -322,7 +325,7 @@ impl AccountInfo {
         let borrow_state = unsafe { (*self.raw).borrow_state };
 
         // check if any borrow (mutable or immutable) is already taken for lamports
-        if borrow_state & 0b_1111_0000 != 0 {
+        if borrow_state & 0b_1111_0000 != 0b_1111_0000 {
             return Err(ProgramError::AccountBorrowFailed);
         }
 
@@ -336,8 +339,9 @@ impl AccountInfo {
         self.can_borrow_data()?;
 
         let borrow_state = unsafe { &mut (*self.raw).borrow_state };
-        // increment the immutable data borrow count
-        *borrow_state += 1;
+        // increment the immutable data borrow count (it is guaranteed that
+        // we have at least one immutable borrow available)
+        *borrow_state -= 1;
 
         // return the reference to data
         Ok(Ref {
@@ -356,7 +360,7 @@ impl AccountInfo {
 
         let borrow_state = unsafe { &mut (*self.raw).borrow_state };
         // set the mutable data borrow flag
-        *borrow_state |= 0b_0000_1000;
+        *borrow_state &= 0b_1111_0111;
 
         // return the mutable reference to data
         Ok(RefMut {
@@ -383,12 +387,12 @@ impl AccountInfo {
 
         // check if mutable data borrow is already taken (most significant bit
         // of the data_borrow_state)
-        if borrow_state & 0b_0000_1000 != 0 {
+        if borrow_state & 0b_0000_1000 == 0 {
             return Err(ProgramError::AccountBorrowFailed);
         }
 
         // check if we have reached the max immutable data borrow count (7)
-        if borrow_state & 0b_0111 == 0b0111 {
+        if borrow_state & 0b_0000_0111 == 0 {
             return Err(ProgramError::AccountBorrowFailed);
         }
 
@@ -410,7 +414,7 @@ impl AccountInfo {
         let borrow_state = unsafe { (*self.raw).borrow_state };
 
         // check if any borrow (mutable or immutable) is already taken for data
-        if borrow_state & 0b_0000_1111 != 0 {
+        if borrow_state & 0b_0000_1111 != 0b_0000_1111 {
             return Err(ProgramError::AccountBorrowFailed);
         }
 
@@ -643,15 +647,15 @@ impl<T: ?Sized> core::ops::Deref for Ref<'_, T> {
 impl<T: ?Sized> Drop for Ref<'_, T> {
     // decrement the immutable borrow count
     fn drop(&mut self) {
-        unsafe { *self.state.as_mut() -= 1 << self.borrow_shift };
+        unsafe { *self.state.as_mut() += 1 << self.borrow_shift };
     }
 }
 
 /// Mask representing the mutable borrow flag for lamports.
-const LAMPORTS_MASK: u8 = 0b_0111_1111;
+const LAMPORTS_MASK: u8 = 0b_1000_0000;
 
 /// Mask representing the mutable borrow flag for data.
-const DATA_MASK: u8 = 0b_1111_0111;
+const DATA_MASK: u8 = 0b_0000_1000;
 
 /// Mutable reference to account data or lamports with checked borrow rules.
 pub struct RefMut<'a, T: ?Sized> {
@@ -722,18 +726,22 @@ impl<T: ?Sized> core::ops::DerefMut for RefMut<'_, T> {
 impl<T: ?Sized> Drop for RefMut<'_, T> {
     fn drop(&mut self) {
         // unset the mutable borrow flag
-        unsafe { *self.state.as_mut() &= self.borrow_mask };
+        unsafe { *self.state.as_mut() |= self.borrow_mask };
     }
 }
 
 #[cfg(test)]
 mod tests {
+    use core::mem::MaybeUninit;
+
+    use crate::NON_DUP_MARKER;
+
     use super::*;
 
     #[test]
     fn test_data_ref() {
         let data: [u8; 4] = [0, 1, 2, 3];
-        let mut state = 1 << DATA_SHIFT;
+        let mut state = NON_DUP_MARKER - (1 << DATA_SHIFT);
 
         let ref_data = Ref {
             value: NonNull::from(&data),
@@ -745,30 +753,30 @@ mod tests {
 
         let new_ref = Ref::map(ref_data, |data| &data[1]);
 
-        assert_eq!(state, 1 << DATA_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER - (1 << DATA_SHIFT));
         assert_eq!(*new_ref, 1);
 
         let Ok(new_ref) = Ref::filter_map(new_ref, |_| Some(&3)) else {
             unreachable!()
         };
 
-        assert_eq!(state, 1 << DATA_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER - (1 << DATA_SHIFT));
         assert_eq!(*new_ref, 3);
 
         let new_ref = Ref::filter_map(new_ref, |_| Option::<&u8>::None);
 
-        assert_eq!(state, 1 << DATA_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER - (1 << DATA_SHIFT));
         assert!(new_ref.is_err());
 
         drop(new_ref);
 
-        assert_eq!(state, 0 << DATA_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER);
     }
 
     #[test]
     fn test_lamports_ref() {
         let lamports: u64 = 10000;
-        let mut state = 1 << LAMPORTS_SHIFT;
+        let mut state = NON_DUP_MARKER - (1 << LAMPORTS_SHIFT);
 
         let ref_lamports = Ref {
             value: NonNull::from(&lamports),
@@ -780,30 +788,30 @@ mod tests {
 
         let new_ref = Ref::map(ref_lamports, |_| &1000);
 
-        assert_eq!(state, 1 << LAMPORTS_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER - (1 << LAMPORTS_SHIFT));
         assert_eq!(*new_ref, 1000);
 
         let Ok(new_ref) = Ref::filter_map(new_ref, |_| Some(&2000)) else {
             unreachable!()
         };
 
-        assert_eq!(state, 1 << LAMPORTS_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER - (1 << LAMPORTS_SHIFT));
         assert_eq!(*new_ref, 2000);
 
         let new_ref = Ref::filter_map(new_ref, |_| Option::<&i32>::None);
 
-        assert_eq!(state, 1 << LAMPORTS_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER - (1 << LAMPORTS_SHIFT));
         assert!(new_ref.is_err());
 
         drop(new_ref);
 
-        assert_eq!(state, 0 << LAMPORTS_SHIFT);
+        assert_eq!(state, NON_DUP_MARKER);
     }
 
     #[test]
     fn test_data_ref_mut() {
         let mut data: [u8; 4] = [0, 1, 2, 3];
-        let mut state = 0b_0000_1000;
+        let mut state = 0b_1111_0111;
 
         let ref_data = RefMut {
             value: NonNull::from(&mut data),
@@ -819,19 +827,19 @@ mod tests {
 
         *new_ref = 4;
 
-        assert_eq!(state, 8);
+        assert_eq!(state, 0b_1111_0111);
         assert_eq!(*new_ref, 4);
 
         drop(new_ref);
 
         assert_eq!(data, [4, 1, 2, 3]);
-        assert_eq!(state, 0);
+        assert_eq!(state, NON_DUP_MARKER);
     }
 
     #[test]
     fn test_lamports_ref_mut() {
         let mut lamports: u64 = 10000;
-        let mut state = 0b_1000_0000;
+        let mut state = 0b_0111_1111;
 
         let ref_lamports = RefMut {
             value: NonNull::from(&mut lamports),
@@ -846,12 +854,138 @@ mod tests {
             lamports
         });
 
-        assert_eq!(state, 128);
+        assert_eq!(state, 0b_0111_1111);
         assert_eq!(*new_ref, 200);
 
         drop(new_ref);
 
         assert_eq!(lamports, 200);
-        assert_eq!(state, 0);
+        assert_eq!(state, NON_DUP_MARKER);
+    }
+
+    #[test]
+    fn test_borrow_data() {
+        let mut data = [0u8; size_of::<Account>()];
+        // Set the borrow state.
+        data[0] = NON_DUP_MARKER;
+        let account_info = AccountInfo {
+            raw: data.as_mut_ptr() as *mut Account,
+        };
+
+        // Check that we can borrow data and lamports.
+        assert!(account_info.can_borrow_data().is_ok());
+        assert!(account_info.can_borrow_mut_data().is_ok());
+        assert!(account_info.can_borrow_lamports().is_ok());
+        assert!(account_info.can_borrow_mut_lamports().is_ok());
+
+        // Borrow immutable data (7 immutable borrows available).
+        const ACCOUNT_REF: MaybeUninit<Ref<[u8]>> = MaybeUninit::<Ref<[u8]>>::uninit();
+        let mut refs = [ACCOUNT_REF; 7];
+
+        refs.iter_mut().for_each(|r| {
+            let Ok(data_ref) = account_info.try_borrow_data() else {
+                panic!("Failed to borrow data");
+            };
+            r.write(data_ref);
+        });
+
+        // Check that we cannot borrow the data anymore.
+        assert!(account_info.can_borrow_data().is_err());
+        assert!(account_info.try_borrow_data().is_err());
+        assert!(account_info.can_borrow_mut_data().is_err());
+        assert!(account_info.try_borrow_mut_data().is_err());
+        // Lamports should still be borrowable.
+        assert!(account_info.can_borrow_lamports().is_ok());
+        assert!(account_info.can_borrow_mut_lamports().is_ok());
+
+        // Drop the immutable borrows.
+        refs.iter_mut().for_each(|r| {
+            let r = unsafe { r.assume_init_read() };
+            drop(r);
+        });
+
+        // We should be able to borrow the data again.
+        assert!(account_info.can_borrow_data().is_ok());
+        assert!(account_info.can_borrow_mut_data().is_ok());
+
+        // Borrow mutable data.
+        let ref_mut = account_info.try_borrow_mut_data().unwrap();
+
+        // Check that we cannot borrow the data anymore.
+        assert!(account_info.can_borrow_data().is_err());
+        assert!(account_info.try_borrow_data().is_err());
+        assert!(account_info.can_borrow_mut_data().is_err());
+        assert!(account_info.try_borrow_mut_data().is_err());
+
+        drop(ref_mut);
+
+        // We should be able to borrow the data again.
+        assert!(account_info.can_borrow_data().is_ok());
+        assert!(account_info.can_borrow_mut_data().is_ok());
+
+        assert!(data[0] == NON_DUP_MARKER);
+    }
+
+    #[test]
+    fn test_borrow_lamports() {
+        let mut data = [0u8; size_of::<Account>()];
+        // Set the borrow state.
+        data[0] = NON_DUP_MARKER;
+        let account_info = AccountInfo {
+            raw: data.as_mut_ptr() as *mut Account,
+        };
+
+        // Check that we can borrow lamports and data.
+        assert!(account_info.can_borrow_lamports().is_ok());
+        assert!(account_info.can_borrow_mut_lamports().is_ok());
+        assert!(account_info.can_borrow_data().is_ok());
+        assert!(account_info.can_borrow_mut_data().is_ok());
+
+        // Borrow immutable lamports (7 immutable borrows available).
+        const LAMPORTS_REF: MaybeUninit<Ref<u64>> = MaybeUninit::<Ref<u64>>::uninit();
+        let mut refs = [LAMPORTS_REF; 7];
+
+        refs.iter_mut().for_each(|r| {
+            let Ok(lamports_ref) = account_info.try_borrow_lamports() else {
+                panic!("Failed to borrow lamports");
+            };
+            r.write(lamports_ref);
+        });
+
+        // Check that we cannot borrow the lamports anymore.
+        assert!(account_info.can_borrow_lamports().is_err());
+        assert!(account_info.try_borrow_lamports().is_err());
+        assert!(account_info.can_borrow_mut_lamports().is_err());
+        assert!(account_info.try_borrow_mut_lamports().is_err());
+        // Data should still be borrowable.
+        assert!(account_info.can_borrow_data().is_ok());
+        assert!(account_info.can_borrow_mut_data().is_ok());
+
+        // Drop the immutable borrows.
+        refs.iter_mut().for_each(|r| {
+            let r = unsafe { r.assume_init_read() };
+            drop(r);
+        });
+
+        // We should be able to borrow the lamports again.
+        assert!(account_info.can_borrow_lamports().is_ok());
+        assert!(account_info.can_borrow_mut_lamports().is_ok());
+
+        // Borrow mutable lamports.
+        let ref_mut = account_info.try_borrow_mut_lamports().unwrap();
+
+        // Check that we cannot borrow the lamports anymore.
+        assert!(account_info.can_borrow_lamports().is_err());
+        assert!(account_info.try_borrow_lamports().is_err());
+        assert!(account_info.can_borrow_mut_lamports().is_err());
+        assert!(account_info.try_borrow_mut_lamports().is_err());
+
+        drop(ref_mut);
+
+        // We should be able to borrow the data again.
+        assert!(account_info.can_borrow_lamports().is_ok());
+        assert!(account_info.can_borrow_mut_lamports().is_ok());
+
+        assert!(data[0] == NON_DUP_MARKER);
     }
 }
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -256,10 +256,13 @@ impl AccountInfo {
         // check if the account lamports are already borrowed
         self.can_borrow_lamports()?;
 
-        let borrow_state = unsafe { &mut (*self.raw).borrow_state };
+        let borrow_state = self.raw as *mut u8;
+        // SAFETY: The `borrow_state` is a mutable pointer to the borrow state
+        // of the account, which is guaranteed to be valid.
+        //
         // "consumes" one immutable borrow for lamports; we are guaranteed
-        // that there is at least one immutable borrow available
-        *borrow_state -= 1 << LAMPORTS_SHIFT;
+        // that there is at least one immutable borrow available.
+        unsafe { *borrow_state -= 1 << LAMPORTS_SHIFT };
 
         // return the reference to lamports
         Ok(Ref {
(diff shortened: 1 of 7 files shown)


sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -112,14 +107,16 @@ const STATIC_ACCOUNT_DATA: usize = core::mem::size_of::<Account>() + MAX_PERMITT
 /// handler manually.
 #[macro_export]
 macro_rules! entrypoint {
-    ( $process_instruction:ident ) => {
-        entrypoint!($process_instruction, { $crate::MAX_TX_ACCOUNTS });
-    };
-    ( $process_instruction:ident, $maximum:expr ) => {
-        $crate::program_entrypoint!($process_instruction, $maximum);
+    ( $process_instruction:expr ) => {
```
febo, line 117: Changing `ident` to `expr` (in all entrypoint macros) allows the use of any function pointer not just identifiers in the current scope.

illia-bobyr: *minor*

If you think it is worth mentioning, maybe it should be added as a comment to the code.
Possibly even to the `entrypoint` doc-comment.
GitHub review comment will be effectively gone after the PR is merged.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -112,14 +107,16 @@ const STATIC_ACCOUNT_DATA: usize = core::mem::size_of::<Account>() + MAX_PERMITT
...
+        $crate::program_entrypoint!($process_instruction);
         $crate::default_allocator!();
         $crate::default_panic_handler!();
     };
+    ( $process_instruction:expr, $maximum:expr ) => {
+        // The `maximum` argument is not used in this macro, but it is kept for
+        // backwards compatibility.
+        entrypoint!($process_instruction);
+    };
```
illia-bobyr: Would it make sense to add a deprecation warning here?
Also, maybe rename the unused argument, so that the generated docs show it is no longer used as well.

```rust
#[deprecated(note = "`$maximum` argument is no longer used. MAX_TX_ACCOUNTS is always used now.")]
#[allow(unused)]
#[inline(always)]
fn entrypoint_no_longer_user_account_number_argument() {}
```

```suggestion
    ( $process_instruction:expr, $_unused_maximum:expr ) => {
        entrypoint_no_longer_user_account_number_argument();
        entrypoint!($process_instruction);
    };
```

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -156,71 +150,64 @@ macro_rules! program_entrypoint {
...
-                        .clone(),
-                );
-            }
-        }
+    for i in 0..processed {
+        let account_info: *mut Account = input as *mut Account;
+        // Adds an 8-bytes offset for:
+        //   - rent epoch in case of a non-duplicated account
+        //   - duplicated marker + 7 bytes of padding in case of a duplicated account
+        input = input.add(core::mem::size_of::<u64>());
```
illia-bobyr, line 320: *minor*

Would it make sense to import `core::mem::size_of`?
Not sure why so many symbols use absolute paths in this file.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -156,71 +150,64 @@ macro_rules! program_entrypoint {
...
+
+        let account = if (*account_info).borrow_state == NON_DUP_MARKER {
+            // Unique account: create a new `AccountInfo` to represent the account.
+            input = input.add(STATIC_ACCOUNT_DATA);
+            input = input.add((*account_info).data_len as usize);
+            input = input.add(input.align_offset(BPF_ALIGN_OF_U128));
+
+            AccountInfo { raw: account_info }
```
illia-bobyr, line 328: Unrelated to this PR.

I think we could create a slice with [from_raw_parts_mut](https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html) here.  With a `'static` lifetime.
As we are parsing the accounts provided by the SVM, it seems the right match.

And after that, access to the account data can happen without unsafe code.

sdk/pinocchio/src/account_info.rs
```diff
@@ -250,7 +253,7 @@ impl AccountInfo {
 
         let borrow_state = unsafe { &mut (*self.raw).borrow_state };
         // increment the immutable borrow count
-        *borrow_state += 1 << LAMPORTS_SHIFT;
+        *borrow_state -= 1 << LAMPORTS_SHIFT;
```
illia-bobyr: A number of locations where the borrow state is updates have comments that mostly repeat the code.
You've updated the code but did not update the comments.
One can argue, that the comments are redundant, considering they mostly repeat the code.

So, there are two options: update the comments to match the code or remove the comments.

sdk/pinocchio/src/account_info.rs
```diff
@@ -269,7 +272,7 @@ impl AccountInfo {
 
         let borrow_state = unsafe { &mut (*self.raw).borrow_state };
         // set the mutable lamport borrow flag
-        *borrow_state |= 0b_1000_0000;
+        *borrow_state &= 0b_0111_1111;
```
illia-bobyr: Unrelated to this PR.

There was a separate PR, where there was a need to track more than one mutable borrow.
I wonder if bits used to count the immutable borrows can be used for that.
When the mutable borrow bit is set, the immutable borrow bits have to be in a known state - no outstanding immutable borrows.
Does it mean they can be used to count mutable borrows now, or am I missing some use for them?

sdk/pinocchio/src/account_info.rs
```diff
@@ -336,8 +339,9 @@ impl AccountInfo {
         self.can_borrow_data()?;
 
         let borrow_state = unsafe { &mut (*self.raw).borrow_state };
-        // increment the immutable data borrow count
-        *borrow_state += 1;
+        // increment the immutable data borrow count (it is guaranteed that
+        // we have at least one immutable borrow available)
+        *borrow_state -= 1;
```
illia-bobyr: Perhaps my comment that you need to update the code comments is incorrect.
Perhaps you intentionally left the "increment" semantics in the comments, while updating the code to use the decrements?

I find it a bit confusing if this is the case.
Maybe, an alternative point of view would be to say that the `borrow_state` now tracks the number of borrows that can still be taken, rather than the number of outstanding borrows?
You start with 8 borrows, and count down for the shared ones.
And for the exclusive, the flag tells you if you can create a mutable borrow.
Not 100% sure it would be easier to understand in the end.
But it does seem more straightforward, than when the comment tells a completely opposite thing compared to the code next to it.

sdk/pinocchio/src/account_info.rs
```diff
@@ -706,18 +710,22 @@ impl<T: ?Sized> core::ops::DerefMut for RefMut<'_, T> {
...
-        unsafe { *self.state.as_mut() &= self.borrow_mask };
+        unsafe { *self.state.as_mut() |= self.borrow_mask };
     }
 }
 
 #[cfg(test)]
 mod tests {
+    use core::mem::{size_of, MaybeUninit};
+
+    use crate::NON_DUP_MARKER;
```
illia-bobyr: Would it make sense to give this constant a second name.
Maybe, `NO_OUTSTANDING_BORROWS` or something like that.
Could make the test code a bit less cryptic.

illia-bobyr:
Overall, it looks good.
Just a few notes.

It is a bit hard to accept the idea that we are inverting the reference tracking logic, and making the code harder to understand, just to avoid a single assignment.
How much CU does a single u64 assignment cost? %)

The account allocation part I'm more accepting of.
If we do not care for the memory allocation, it does make it easier to just allocate the maximum in all the cases.
At the same time, it does mean that the caller will have uninitialled elements in the array they hold.
I guess the only intended caller here is the code produced by our own macro.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Good point, I will add some comments there.

sdk/pinocchio/src/entrypoint/mod.rs
febo: That is a good idea! 👍

sdk/pinocchio/src/entrypoint/mod.rs
febo: There are many places it is using the absolute path but we can start changing that.

sdk/pinocchio/src/entrypoint/mod.rs
febo: I am not sure I follow this. We create a slice once we call the program entrypoint, but at this point we still haven't initialized all accounts.

sdk/pinocchio/src/account_info.rs
febo: Yeah, it is a bit redundant. Sometimes looking at the code it is not necessarily clear what it is doing – in this case it looks like decrementing but in fact we are increasing the borrow count. Maybe we need to expand the code to make that more clear.

sdk/pinocchio/src/account_info.rs
febo: We only allow one mutable borrow – this flag is used to mimic the borrow checker where you can have multiple immutable borrows, but only one (exclusive) mutable borrow.

sdk/pinocchio/src/account_info.rs
febo: >Perhaps my comment that you need to update the code comments is incorrect.
Perhaps you intentionally left the "increment" semantics in the comments, while updating the code to use the decrements?

Yes, exactly.

> Maybe, an alternative point of view would be to say that the borrow_state now tracks the number of borrows that can still be taken, rather than the number of outstanding borrows?

This is a very nice way to see it, I think it will make the logic easier to understand.

sdk/pinocchio/src/account_info.rs
febo: Yeah, maybe `NOT_BORROWED` since it represents the case where there are no active borrows.

febo:
> It is a bit hard to accept the idea that we are inverting the reference tracking logic, and making the code harder to understand, just to avoid a single assignment.
> How much CU does a single u64 assignment cost? %)

It is not much about CU savings, which is about 1 CU per account, but more to clean up the logic. It is a bit strange to be setting this value to every account even the ones that you don't borrow the data – e.g., programs. I will make the tweaks that you suggested to view this flag as "remaining borrows" so perhaps this would make the code clearer.

sdk/pinocchio/src/account_info.rs
illia-bobyr: Sorry for not being specific enough.
I got an impression that the functionality suggested in [Add RefMut::map_split method and unit test](https://github.com/anza-xyz/pinocchio/pull/158) is useful.
But can not be implemented with the current model, as we can only track a single outstanding mutable borrow.

sdk/pinocchio/src/entrypoint/mod.rs
illia-bobyr: You are right.  This suggestion can certainly be viewed as out of scope for this PR.
Maybe I should send a PR that just fixed the imports and see if someone would approve it :)

sdk/pinocchio/src/entrypoint/mod.rs
illia-bobyr: All the operations that interact with the `raw` field need to be unsafe:

https://github.com/anza-xyz/pinocchio/blob/936160198fe05e95920db5a8bb0576cbd7fda8f1/sdk/pinocchio/src/account_info.rs#L117-L163

Which is not necessary, I think, if we convert `raw` into `&'static [u8]`.
We do not need to be uncertain about the sharing semantics for this pointer.
We know that it is a global memory that is not going away, and it will not be accessed in parallel.
If we use a raw pointer, this sharing information is not available to the compiler.
If we use a slice - then it is.

One other thing is that the slice holds the length.
Which may or may not be visible in your CU measurements.
On x86 the common knowledge is that slices are as efficient as pointers, but are more convenient to use, so pointer usage should be minimized.
It seems that our target is somewhat different both in terms of the compiler effectiveness and in terms of the overhead we are willing to ignore.

sdk/pinocchio/src/account_info.rs
illia-bobyr: If we look at the borrow state as holding a borrow budget, then this code can be read as "using one more borrow from the budget".
You could probably phrase it better.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Hmmm, I don't think we can use a slice. This pointer is shared among `AccountInfo`s since an account can be duplicated – there are potentially multiple copies of each. That would violate Rust aliasing requirements since the pointer needs to be mutable.

sdk/pinocchio/src/account_info.rs
febo: Yeah, we can't track multiple ones since we are reusing an `u8` that is already present in the account and divided into data and lamports flags.

There is a workaround to do the split if you work with the `&[u8]` and map it. I posted as a reply to the original Stack Exchange question: https://solana.stackexchange.com/questions/22154/pinocchiorefmut-is-missing-the-map-split-function

sdk/pinocchio/src/entrypoint/mod.rs
illia-bobyr: I didn't realize it.
I think the `unsafe` blocks in all these readers might benefit from good comments, explaining why is it safe to read through this pointer at this particular moment.

It seems that `lamports` and `data` are the only fields that can be modified, right?

For account bytes, aka `data`, borrowing is checked by the `Ref`/`RefMut` instances - those seem to enforce the constraints.
Which leaves `lamports()` and `data_len()`.

I think `lamports()` and `try_borrow_mut_lamports()` in combination violate the rule that there could only be one mutable reference to an address.
If I create a mutable reference to the lamports part of the `raw` data with `try_borrow_mut_lamports()` the compiler assumes that there should exist no other pointer to this data.
Yet, I can call `lamports()` to read the same value without any checks.
I think `lampots()` needs to have a `self.can_borrow_lamports()` call at the beginning, before it dereferences the `raw` pointer.
What do you think?
I wonder if there is still an issue of interior mutability (see below). 

A similar problem may exist for the `data_len()`.
It is a bit more subtle, as the only way to change the data length is via the `realloc()` call.
But `realloc()` only requires `&self`, meaning, we have a case of [interior mutability](https://doc.rust-lang.org/reference/interior-mutability.html).
This breaks the compiler invariants, unless explicitly marked with [`core::cell:UnsafeCell`](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html).

sdk/pinocchio/src/account_info.rs
illia-bobyr: But each part separately has more state space that we can use.
For example, for the data references, we use 4 bits.
One bit is for the presence of a mutable borrow, and 3 bits for the immutable borrows.

Currently (after this PR) valid values for these 4 bits are:
 * `1 1 1 1` - no borrows exist, it is possible to take a mutable borrow, or an immutable borrow.
 * `0 1 1 1` - a mutable borrow exists. No immutable borrows are possible.  No more mutable borrows are possible.
 * `1 x y z` - where `xyz` is not `111`.  One or more immutable borrows exist.  No mutable borrows are possible.

State of `0 x y z` where `xyz` is anything but `111` is not used at the moment.
We can use these 3 bits to count outstanding non-overlapping mutable borrows, if we want to.

As it is impossible to have both mutable and immutable borrows at the same time, it is fine to reuse the immutable borrows counter for mutable borrows.  We must have had 0 immutable borrows at the moment when we allocated the first mutable borrow.

---

I also wonder if we could have used other fields in the account to avoid this bit manipulation.
As well as allow more than 7 references.
Right now we take over the state byte, but there are also "is signer", and "is writable" bytes.
Both use only a single bit really.
With the state byte we have 3 bytes.
We could have shifted all the flags into the first byte, and use the other two as two counters, for the data and the lamport references.
Or, there are probably other layouts.

illia-bobyr:
> > It is a bit hard to accept the idea that we are inverting the reference tracking logic, and making the code harder to understand, just to avoid a single assignment.
> > How much CU does a single u64 assignment cost? %)
> 
> It is not much about CU savings, which is about 1 CU per account, but more to clean up the logic.

Thank you for the explanation.

> It is a bit strange to be setting this value to every account even the ones that you don't borrow the data – e.g., programs. [...]

It is probably subjective, but it seems fine to me.
The VM gives us the data.  We shape it in a way we see fit.
We are going to reuse this space for our own memory layout.
It does not have to match the original layout literally, as long as the meaning is the same.
We update the program account to have correct meaning, using our internal representation of the account information.
I agree that it is a bit of a waste if we do the transformation, but then do not use the account in any particular way within the current program.

At the same time, even for program accounts, I think, you can read the program bytes, no?
So, the program account is not different from any other account.
And the modification is an update of a single byte - seems pretty minimal :)

I would actually argue, it might be OK to update a few more bytes if it give a better UX :P

sdk/pinocchio/src/entrypoint/mod.rs
febo: > I didn't realize it. I think the `unsafe` blocks in all these readers might benefit from good comments, explaining why is it safe to read through this pointer at this particular moment.

Good point, might be better to do in a separate PR right?

> It seems that `lamports` and `data` are the only fields that can be modified, right?

And `owner` as well – that one has an unsafe getter.

> For account bytes, aka `data`, borrowing is checked by the `Ref`/`RefMut` instances - those seem to enforce the constraints. Which leaves `lamports()` and `data_len()`.

`data_length()` has no setter (apart from `realloc()`) and it does not overlap with the `data` slice. There is no other way to change the length of the data apart from `realloc()`, so I guess that one is safe.

> I think `lamports()` and `try_borrow_mut_lamports()` in combination violate the rule that there could only be one mutable reference to an address. If I create a mutable reference to the lamports part of the `raw` data with `try_borrow_mut_lamports()` the compiler assumes that there should exist no other pointer to this data. Yet, I can call `lamports()` to read the same value without any checks. I think `lampots()` needs to have a `self.can_borrow_lamports()` call at the beginning, before it dereferences the `raw` pointer. What do you think? I wonder if there is still an issue of interior mutability (see below).

Agreed, this is a very good point. My initial idea is to have `lamports()` do the borrow check for mutability and return a `Result`; add another `lamports_unchcked()` that does not do the check for cases where the program guarantees unique access.

> A similar problem may exist for the `data_len()`. It is a bit more subtle, as the only way to change the data length is via the `realloc()` call. But `realloc()` only requires `&self`, meaning, we have a case of [interior mutability](https://doc.rust-lang.org/reference/interior-mutability.html). This breaks the compiler invariants, unless explicitly marked with [`core::cell:UnsafeCell`](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html).

The difference is that `realloc()` does the borrow check so I think we are safe here.

sdk/pinocchio/src/account_info.rs
febo: Oh, I see. Yeah, we could reuse some other bits. Just wondering if it is worth doing it. I think at this point it would be better to rely on the borrow checker – e.g., you borrow the data slice using `try_borrow_mut_data`, then use `split_at_mut` on the deref'ed slice. The original byte slice is still "controlled" by the `Ref*` returned and now you have non-overlapping mutable slices with borrow checker.

> I also wonder if we could have used other fields in the account to avoid this bit manipulation.
As well as allow more than 7 references.
Right now we take over the state byte, but there are also "is signer", and "is writable" bytes.

I feel it could get confusing if we start re-using multiple fields to track borrows. There are ways to work with the current references that are returned by the `try_borrow_*` (as mentioned above). One of the limitation of having "only" `7` borrows is that if you have an account duplicated more than `7` times, you might end up with issues trying to borrow the at the same time – most likely this will be extremely rare.

febo:
> At the same time, even for program accounts, I think, you can read the program bytes, no?

Depend on the program loader – for loader `v3`, the program data is in a separate account.

> So, the program account is not different from any other account. And the modification is an update of a single byte - seems pretty minimal :)
> 
> I would actually argue, it might be OK to update a few more bytes if it give a better UX :P

I tend to agree, but in this case I think the logic of having a borrow "budget" in a way makes more sense. Also, as the number of accounts increase, CUs for this would also increase – what might seem minimal now, it could eventually add up once the maximum number of accounts increase.

Somewhat related to this, simplifying the entrypoint logic eventually leads to bigger CU gains – [PR #176](https://github.com/anza-xyz/pinocchio/pull/176) improves by `~20%`. 😊

sdk/pinocchio/src/entrypoint/mod.rs
illia-bobyr: > > I didn't realize it. I think the `unsafe` blocks in all these readers might benefit from good comments, explaining why is it safe to read through this pointer at this particular moment.
> 
> Good point, might be better to do in a separate PR right?

Yes, this whole thread is not related to the work done in this PR.
The first sentence in the first comment in this thread is "Unrelated to this PR."
I was wondering if I should repeat it in some of my replies, but figure that as the thread starts like this, it would probably be unnecessary :)

> > It seems that `lamports` and `data` are the only fields that can be modified, right?
> 
> And `owner` as well – that one has an unsafe getter.

I thought only the system program can change the account owner, no?
So, I assumed it is a read-only field.

> > For account bytes, aka `data`, borrowing is checked by the `Ref`/`RefMut` instances - those seem to enforce the constraints. Which leaves `lamports()` and `data_len()`.
> 
> `data_length()` has no setter (apart from `realloc()`) and it does not overlap with the `data` slice. There is no other way to change the length of the data apart from `realloc()`, so I guess that one is safe.

But `realloc()` is a setter.

> > I think `lamports()` and `try_borrow_mut_lamports()` in combination violate the rule that there could only be one mutable reference to an address. If I create a mutable reference to the lamports part of the `raw` data with `try_borrow_mut_lamports()` the compiler assumes that there should exist no other pointer to this data. Yet, I can call `lamports()` to read the same value without any checks. I think `lampots()` needs to have a `self.can_borrow_lamports()` call at the beginning, before it dereferences the `raw` pointer. What do you think? I wonder if there is still an issue of interior mutability (see below).
> 
> Agreed, this is a very good point. My initial idea is to have `lamports()` do the borrow check for mutability and return a `Result`; add another `lamports_unchcked()` that does not do the check for cases where the program guarantees unique access.

Yeah, this is one option.
Another option is to remove the direct length getter, and rely on the `data()` that returns a `Ref` wrapping the data slice, from which one can get the length.
I am not recommending one over the other, just stating that there is a possible alternative to consider.

> > A similar problem may exist for the `data_len()`. It is a bit more subtle, as the only way to change the data length is via the `realloc()` call. But `realloc()` only requires `&self`, meaning, we have a case of [interior mutability](https://doc.rust-lang.org/reference/interior-mutability.html). This breaks the compiler invariants, unless explicitly marked with [`core::cell:UnsafeCell`](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html).
> 
> The difference is that `realloc()` does the borrow check so I think we are safe here.

But the `data_len()` does not check.
So you can read length via `data_len()` then resize, then read via `data_len()` again.
The compiler can reuse the first `data_len()` result, as it assumes that `&self` can not be mutated internally, unless there is an `UnsafeCell` that disables this assumption.

I would imagine that it is unlikely that an average program would trigger the undefined behavior either via `lamports()` or via `data_len()`.
But if I'm being pedantic, this is still a violation of the language rules.

sdk/pinocchio/src/account_info.rs
illia-bobyr: > Oh, I see. Yeah, we could reuse some other bits. Just wondering if it is worth doing it. I think at this point it would be better to rely on the borrow checker – e.g., you borrow the data slice using `try_borrow_mut_data`, then use `split_at_mut` on the deref'ed slice. The original byte slice is still "controlled" by the `Ref*` returned and now you have non-overlapping mutable slices with borrow checker.

I thought these are different things.
The borrow checker only works for code that can be analyzed statically.
And this is my it does not need any runtime support.

References are a run-time equivalent.
So, if you can use the borrow checker, you absolutely should, as it means you are only paying the price at compile time.
But for applications that require run time checking, the borrow checker is of no help.

I'm not sure if the proposed change was actually adding something that can only be done at run time.
I must admit I did not analyze it all the way, after you said it is impossible with the current system.

sdk/pinocchio/src/entrypoint/mod.rs
febo: > > > I didn't realize it. I think the `unsafe` blocks in all these readers might benefit from good comments, explaining why is it safe to read through this pointer at this particular moment.
> > 
> > 
> > Good point, might be better to do in a separate PR right?
> 
> Yes, this whole thread is not related to the work done in this PR. The first sentence in the first comment in this thread is "Unrelated to this PR." I was wondering if I should repeat it in some of my replies, but figure that as the thread starts like this, it would probably be unnecessary :)

Unnecessary, I was just confirming. 😉

> > > It seems that `lamports` and `data` are the only fields that can be modified, right?
> > 
> > 
> > And `owner` as well – that one has an unsafe getter.
> 
> I thought only the system program can change the account owner, no? So, I assumed it is a read-only field.

There is an `assign()` method that a program could call on an `AccountInfo`. They can do this when they own the account and want to "give it" to another program. It can only happen in a very specific way (empty/zeroed account data), but it is possible.

> > > For account bytes, aka `data`, borrowing is checked by the `Ref`/`RefMut` instances - those seem to enforce the constraints. Which leaves `lamports()` and `data_len()`.
> > 
> > 
> > `data_length()` has no setter (apart from `realloc()`) and it does not overlap with the `data` slice. There is no other way to change the length of the data apart from `realloc()`, so I guess that one is safe.
> 
> But `realloc()` is a setter.
>
> > > I think `lamports()` and `try_borrow_mut_lamports()` in combination violate the rule that there could only be one mutable reference to an address. If I create a mutable reference to the lamports part of the `raw` data with `try_borrow_mut_lamports()` the compiler assumes that there should exist no other pointer to this data. Yet, I can call `lamports()` to read the same value without any checks. I think `lampots()` needs to have a `self.can_borrow_lamports()` call at the beginning, before it dereferences the `raw` pointer. What do you think? I wonder if there is still an issue of interior mutability (see below).
> > 
> > 
> > Agreed, this is a very good point. My initial idea is to have `lamports()` do the borrow check for mutability and return a `Result`; add another `lamports_unchcked()` that does not do the check for cases where the program guarantees unique access.
> 
> Yeah, this is one option. Another option is to remove the direct length getter, and rely on the `data()` that returns a `Ref` wrapping the data slice, from which one can get the length. I am not recommending one over the other, just stating that there is a possible alternative to consider.
> 
> > > A similar problem may exist for the `data_len()`. It is a bit more subtle, as the only way to change the data length is via the `realloc()` call. But `realloc()` only requires `&self`, meaning, we have a case of [interior mutability](https://doc.rust-lang.org/reference/interior-mutability.html). This breaks the compiler invariants, unless explicitly marked with [`core::cell:UnsafeCell`](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html).
> > 
> > 
> > The difference is that `realloc()` does the borrow check so I think we are safe here.
> 
> But the `data_len()` does not check. So you can read length via `data_len()` then resize, then read via `data_len()` again. The compiler can reuse the first `data_len()` result, as it assumes that `&self` can not be mutated internally, unless there is an `UnsafeCell` that disables this assumption.
> 
> I would imagine that it is unlikely that an average program would trigger the undefined behavior either via `lamports()` or via `data_len()`. But if I'm being pedantic, this is still a violation of the language rules.

Got it, very good point! We should fix this to avoid issues. For `data_len()`, would using `read_volatile` mitigate the issue? Otherwise we could remove `data_len()` as you suggested.

sdk/pinocchio/src/account_info.rs
febo: > > Oh, I see. Yeah, we could reuse some other bits. Just wondering if it is worth doing it. I think at this point it would be better to rely on the borrow checker – e.g., you borrow the data slice using `try_borrow_mut_data`, then use `split_at_mut` on the deref'ed slice. The original byte slice is still "controlled" by the `Ref*` returned and now you have non-overlapping mutable slices with borrow checker.
> 
> I thought these are different things. The borrow checker only works for code that can be analyzed statically. And this is my it does not need any runtime support.
> 
> References are a run-time equivalent. So, if you can use the borrow checker, you absolutely should, as it means you are only paying the price at compile time. But for applications that require run time checking, the borrow checker is of no help.
> 
> I'm not sure if the proposed change was actually adding something that can only be done at run time. I must admit I did not analyze it all the way, after you said it is impossible with the current system.

We can't rely on the borrow checker at first, since accounts can be duplicated and they share the "raw" pointer. But once you get a reference to the data via `try_borrow_data_*`, then that slice can now be under the borrow checker rules. So this is why I don't think we need to create a more complex way to handle non-overlapping borrows – borrow checked can do this for us.

In other words, instead of supporting non-overlapping borrows that need to return multiple `RefMut`s, we offer a way to get a single `RefMut` that represents the whole data and then use `split_at_mut` on the slice to create non-overlapping borrows but this time controlled by the borrow checker. Does that make sense?

sdk/pinocchio/src/entrypoint/lazy.rs
```diff
@@ -271,6 +268,7 @@ impl InstructionContext {
 }
 
 /// Wrapper type around an [`AccountInfo`] that may be a duplicate.
+#[repr(C)]
```
joncinque: nit: Is this needed?

sdk/pinocchio/src/cpi.rs
joncinque, line 1: I'm guessing a rebase will take out these changes and all the ones under `programs/system`, correct?

febo: Yes, I will rebase the PR.

sdk/pinocchio/src/entrypoint/lazy.rs
febo: No, that should be removed.

sdk/pinocchio/src/entrypoint/mod.rs
febo: > Maybe I should send a PR that just fixed the imports and see if someone would approve it :)

Yes please! 😊

sdk/pinocchio/src/entrypoint/mod.rs
febo: I reverted this change, so we still have 2 versions: one specifying the maximum and skipping any "extra" one; and another using the theoretical maximum number of accounts. We could eventually deprecate the first one when we have dynamic stack frames.

febo:
@ilya-bobyr @joncinque This is ready for another look. Only what is on the PR description is included on this PR. There are very good points that Ilya raised here which we can address in separate PRs once we decide how to approach them.

sdk/pinocchio/src/account_info.rs
```diff
@@ -37,16 +37,21 @@ pub enum BorrowState {
...
-    /// 4 bit state: [1 bit mutable borrow flag | u3 immmutable borrow flag]
-    /// This gives us up to 7 immutable borrows. Note that does not mean 7
-    /// duplicate account infos, but rather 7 calls to borrow lamports or
-    /// borrow data across all duplicate account infos.
+    ///   * `1 1 1 1` (first 4 bits): borrows available for lamports,
+    ///     `1` mutable borrow or `111` (7) immutable borrows.
+    ///
+    ///   * `1 1 1 1` (last 4 bits): borrows available for data,
+    ///     `1` mutable borrow or `111` (7) immutable borrows.
+    ///
```
illia-bobyr: *bikeshedding*

If you are already going into the trouble of explaining bits, I've seen this style that I find very easy to understand:

```suggestion
    /// Bit in the borrow byte are used as follows:
    ///
    ///   * `x . . . . . . .` : lamport mutable borrow flag.
    ///     `7 6 5 4 3 2 1 0`
    ///     `1` - the lamport field can be mutably borrowed.
    ///     `0` - there is an outstanding mutable borrow for the lamports.
    ///
    ///   * `. x x x . . . .` : lamport immutable borrow count.
    ///     `7 6 5 4 3 2 1 0`
    ///     Number of immutable borrows that can still be allocated, for the
    ///     lamports field.
    ///     Ranges from 7 (`111`) to 0 (`000`).
    ///
    ///   * `. . . . x . . .` : data mutable borrow flag.
    ///     `7 6 5 4 3 2 1 0`
    ///     `1` - the account data can be mutably borrowed.
    ///     `0` - there is an outstanding mutable borrow for the account data.
    ///
    ///   * `. . . . . x x x` : data immutable borrow count.
    ///     `7 6 5 4 3 2 1 0`
    ///     Number of immutable borrows that can still be allocated, for the
    ///     account data.
    ///     Ranges from 7 (`111`) to 0 (`000`).
    ///
```

Maybe the text can be rephrased, by the main idea is that we explicitly list each bit group and specify what does it mean.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -236,6 +284,71 @@ pub unsafe fn deserialize<'a, const MAX_ACCOUNTS: usize>(
...
+///
+/// # Safety
+///
+/// The caller must ensure that the input buffer is valid, i.e., it represents the
+/// program input parameters serialized by the SVM loader.
+#[inline(always)]
+pub unsafe fn parse<'a, const ACCOUNTS: usize>(
+    mut input: *mut u8,
+    accounts: &mut [core::mem::MaybeUninit<AccountInfo>; ACCOUNTS],
+) -> (&'a Pubkey, usize, &'a [u8]) {
```
illia-bobyr: *minor*

I personally try to name the lifetime parameters, in order to make it easier to understand what they mean.

Here, I think, it could be particularly useful, as `'a` is absent from the input parameters, so it is very confusing, as it looks like the caller can just come up with any lifetime they want and it would be fine.
While in reality, `'a` here is the lifetime of the `input` pointer.

```suggestion
pub unsafe fn parse<'input, const ACCOUNTS: usize>(
    mut input: *mut u8,
    accounts: &mut [core::mem::MaybeUninit<AccountInfo>; ACCOUNTS],
) -> (&'input Pubkey, usize, &'input [u8]) {
```

While renaming does not allow the compiler to check correctness, it helps the human reader, I think.

At the same time, I wonder if this should be just `'static`?
From the standpoint of the SVM program, `input` is global and exist for the lifetime of the program, right?

I think depending on the answer to the question above, the `Safety` requirement for this function could be updated.
It should either state that `input` needs to be global, or it needs to state that the returned values reference `input` and it is the caller responsibility to ensure they are not used beyond the `input` lifetime.

sdk/pinocchio/src/lib.rs
```diff
@@ -243,14 +243,14 @@ pub use entrypoint::lazy as lazy_entrypoint;
...
+///  reserved to indicate non-duplicated accounts.
 ///
-/// This is based on the current [maximum number of accounts] that a transaction
-/// may lock in a block.
-///
-/// [maximum number of accounts]: https://github.com/anza-xyz/agave/blob/2e6ca8c1f62db62c1db7f19c9962d4db43d0d550/runtime/src/bank.rs#L3209-L3221
-pub const MAX_TX_ACCOUNTS: usize = 128;
+/// The `MAX_TX_ACCOUNTS` is used to statically initialize the array of
+/// `AccountInfo`s when parsing accounts in an instruction.
+pub const MAX_TX_ACCOUNTS: usize = u8::MAX as usize;
```
illia-bobyr: *minor*

If you are mentioning that `255` can not be used as an account index, would it then make sense to reduce the max number by 1?
I think, in practice, `128` is a more realistic estimate.
Though, overestimating is probably free and have no real negative consequences.

```rust
pub const MAX_TX_ACCOUNTS: usize = (u8::MAX - 1) as usize;
```

By the way, do you consider this change related to the program entrypoint simplification or did it get here by accident?

ilya-bobyr (dismissed):
Looks good to me.
A few minor things you may want to consider.

sdk/pinocchio/src/account_info.rs
febo: Nice suggestion, thanks! Tweaked a bit the layout so it renders like intended.

sdk/pinocchio/src/lib.rs
febo: Yes, it is part since the idea is to avoid having the duplicated logic to skip accounts when the array does not have space for them. The reason for `u8::MAX` (now with a `- 1`)  is to use the theoretical maximum value, so there is no chance the number of accounts can go over it.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Updated in [96587ca](https://github.com/anza-xyz/pinocchio/pull/166/commits/96587caa4af70717f56ac2841735d9d90f27a5f8).

sdk/pinocchio/src/account_info.rs
illia-bobyr: I like your version better :)

ilya-bobyr (approved):
Great work!

PR #170, opened 2025-06-04
Remove *const cast

febo:
### Problem

The way the data pointer (`data_ptr`) is calculated to access account data within an `AccountInfo` does an unnecessary cast to `*const _` before casting to `*mut u8`. Since `self.raw` is already a `*mut` pointer, it can be casted to `*mut u8` directly.

### Solution

Remove the unnecessary `*const _`.


joncinque (approved):
Makes sense to me!

PR #171, opened 2025-06-04
WIP: Changes required for star-frame integration

stegaBOB:
Adds some fixes/updates to enable compatibility/ease of use with our work-in-progress solana framework.


sdk/pinocchio/src/account_info.rs
```diff
@@ -104,7 +105,7 @@ const GET_LEN_MASK: u32 = !SET_LEN_MASK;
 /// used to track borrows of the account data and lamports, given that an
 /// account can be "shared" across multiple `AccountInfo` instances.
 #[repr(C)]
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
```
bonedaddy, line 108: Would be nice to have `Debug` traits feature gated with `cfg_attr`.

These debug traits only seem useful in specific circumstances (ie: not when deploying a program to production), and always deriving the debug traits will increase the size of the compiled binary.

stegaBOB: Every public type should impl debug IMO. I'm a bit skeptical this would actually increase binary size at all, but if it does I'd argue that's worth it. Not having debug is very painful to work with.

stegaBOB:
Splitting into separate PRs and rebasing!

PR #175, opened 2025-06-07
Simplify realloc logic

febo:
### Problem

The current way that `realloc` tracks the original data length is unnecessarily complex.

### Solution

This PR simplifies the logic on `realloc` to store difference between the original data length and the current length.

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -1,4 +1,5 @@
 //! Data structures to represent account information.
+
 use core::{
     marker::PhantomData,
     mem::ManuallyDrop,
@@ -80,18 +81,13 @@ pub(crate) struct Account {
     /// Indicates whether this account represents a program.
     executable: u8,
 
-    /// Account's original data length when it was serialized for the
-    /// current program invocation.
+    /// Difference between the original data length and the current
+    /// data length.
     ///
-    /// The value of this field is lazily initialized to the current data length
-    /// and the [`SET_LEN_MASK`] flag on first access. When reading this field,
-    /// the flag is cleared to retrieve the original data length by using the
-    /// [`GET_LEN_MASK`] mask.
-    ///
-    /// Currently, this value is only used for `realloc` to determine if the
-    /// account data length has changed from the original serialized length beyond
-    /// the maximum permitted data increase.
-    original_data_len: u32,
+    /// This is used to track the original data length of the account
+    /// when the account is resized. The runtime guarantees that this
+    /// value is zero at the start of the instruction.
+    resize_delta: i32,
 
     /// Public key of the account.
     key: Pubkey,
@@ -105,20 +101,6 @@ pub(crate) struct Account {
     /// Length of the data. Modifiable by programs.
     pub(crate) data_len: u64,
 }
-
-/// Mask to indicate the original data length has been set.
-///
-/// This takes advantage of the fact that the original data length will not
-/// be greater than 10_000_000 bytes, so we can use the most significant bit
-/// as a flag to indicate that the original data length has been set and lazily
-/// initialize its value.
-const SET_LEN_MASK: u32 = 1 << 31;
-
-/// Mask to retrieve the original data length.
-///
-/// This mask is used to retrieve the original data length from the `original_data_len`
-/// by clearing the flag that indicates the original data length has been set.
-const GET_LEN_MASK: u32 = !SET_LEN_MASK;
 
 /// Wrapper struct for an `Account`.
 ///
@@ -176,6 +158,16 @@ impl AccountInfo {
     #[inline(always)]
     pub fn data_len(&self) -> usize {
         unsafe { (*self.raw).data_len as usize }
+    }
+
+    /// Returns the delta between the original data length and the current
+    /// data length.
+    ///
+    /// This value will be different than zero if the account has been resized
+    /// during the current instruction.
+    #[inline(always)]
+    pub fn resize_delta(&self) -> i32 {
+        unsafe { (*self.raw).resize_delta }
     }
 
     /// Returns the lamports in the account.
@@ -461,8 +453,8 @@ impl AccountInfo {
     /// Realloc the account's data and optionally zero-initialize the new
     /// memory.
     ///
-    /// Note:  Account data can be increased within a single call by up to
-    /// [`MAX_PERMITTED_DATA_INCREASE`] bytes.
+    /// The account data can be increased by up to [`MAX_PERMITTED_DATA_INCREASE`] bytes
+    /// within an instruction.
     ///
     /// Note: Memory used to grow is already zero-initialized upon program
     /// entrypoint and re-zeroing it wastes compute units.  If within the same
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/account_info.rs
```diff
@@ -437,56 +429,47 @@ impl AccountInfo {
...
-            }
-        };
+        let difference = new_len - current_len;
+        let accumulated_resize_delta = self.resize_delta() + difference;
 
-        // return early if the length increase from the original serialized data
+        // Return an error when the length increase from the original serialized data
         // length is too large and would result in an out of bounds allocation
-        if new_len.saturating_sub(original_len) > MAX_PERMITTED_DATA_INCREASE {
+        if accumulated_resize_delta > MAX_PERMITTED_DATA_INCREASE as i32 {
```
rustopian, line 497: @febo Would we encounter any problems if someone:

1. provides an existing account with some large initial size
2. calls `realloc(shrink_len, false)` to reduce its size (`resize_delta` is now negative)
3. calls `realloc(new_len, true)` again, with `zero_init` as `true`, to increase account size by more than `MAX_PERMITTED_DATA_INCREASE` – since the function only checks the overall delta, this will be allowed
4. `sol_memset_` now zero-inits more memory than allowed by `MAX_PERMITTED_DATA_INCREASE`, since `difference > MAX_PERMITTED_DATA_INCREASE` even though `accumulated_resize_delta <= MAX_PERMITTED_DATA_INCREASE`


For example, in this test code, the final line throws:
```
const HEADER: usize = core::mem::size_of::<Account>();
const INITIAL_LEN: usize = MAX_PERMITTED_DATA_INCREASE;
const TOTAL: usize   = HEADER + INITIAL_LEN + 8;

let mut large_account_data = [0u8; TOTAL];
let alignment_offset = large_account_data.as_ptr() as usize % 8;
let large_slice = &mut large_account_data[alignment_offset..];

// 1. Pretend runtime gives us account with data_len of 10_240
large_slice[80..88].copy_from_slice(&(INITIAL_LEN as u64).to_le_bytes());

let exploit = AccountInfo {
    raw: large_slice as *const _ as *mut Account,
};

assert_eq!(exploit.data_len(), INITIAL_LEN);
assert_eq!(exploit.resize_delta(), 0);

// 2. shrink by the 10_240-byte cap → delta = −10_240
let shrink_len = INITIAL_LEN - MAX_PERMITTED_DATA_INCREASE;
exploit.realloc(shrink_len, false).unwrap();

// 3. grow by cap *2 (20_480) — now we are zero-initting 2 * MAX_PERMITTED_DATA_INCREASE in one shot
let grow_len = shrink_len + MAX_PERMITTED_DATA_INCREASE * 2;

assert!(exploit.realloc(grow_len, true).is_err(),
        "Vulnerability? Trying to zero-init more memory than allowed");
```

Possibly there could be some kind of attack where someone creates a large account over a period of time and then repeatedly cuts it down ~10MB and sizes it up ~10MB with `zero_init`.

rustopian (commented):
Looks good overall, but I do wonder about a possible exploit of multiple `realloc()` calls given the limitations of the `delta` approach - see comment

sdk/pinocchio/src/account_info.rs
febo: It should not be a problem. The data memory layout for each account is allocated by the runtime at the start of the instruction:
```
+------------------------+---------------------------------------+
|  ORIGINAL_DATA_LENGTH  |  MAX_PERMITTED_DATA_INCREASE (10KiB)  |
+------------------------+---------------------------------------+
```
So you could realloc the account to `0` and then realloc it again up to `ORIGINAL_DATA_LENGTH + MAX_PERMITTED_DATA_INCREASE`, since it will still be within the bounds of the account memory. The runtime allocated memory does not shrink until the end of the instruction – realloc only changes the length of the account data pointer.

The important check is that the accumulated resize (delta in relation to `ORIGINAL_DATA_LENGTH`) cannot be over `MAX_PERMITTED_DATA_INCREASE`, otherwise the scenario that you described can happen – access to memory that does not belong to the account.

rustopian: Got it, thanks. I was initially concerned about fee discrepancies, thinking (paranoid) that the 10KiB limit might be used elsewhere to bill for worst-case writes (or something similar affecting fees), but that doesn't seem to be the case.

PR looks good to me.

rustopian (approved):

ilya-bobyr (dismissed):
Using zero initialization to both avoid length assignment and make the logic simpler is really smart :)

sdk/pinocchio/src/account_info.rs
```diff
@@ -420,8 +412,8 @@ impl AccountInfo {
     /// Realloc the account's data and optionally zero-initialize the new
     /// memory.
     ///
-    /// Note:  Account data can be increased within a single call by up to
-    /// [`MAX_PERMITTED_DATA_INCREASE`] bytes.
+    /// The account data can be increased by up to [`MAX_PERMITTED_DATA_INCREASE`] bytes
+    /// within an instruction.
     ///
     /// Note: Memory used to grow is already zero-initialized upon program
     /// entrypoint and re-zeroing it wastes compute units.  If within the same
```
LStan: Probably this `Note` about `zero_init` should be changed as [here](https://github.com/anza-xyz/solana-sdk/pull/176/files)

joncinque (commented):
The change looks good to me!

There is an edge case if someone calls `close()` and then `resize()` within the same instruction.

With the current code, `resize()` would incorrectly return an error for going over the limit, even if the account previously had some space.

How about setting the difference during `close()`? `close_unchecked()` can stay as it is, but it would be good to add a comment saying to avoid resizing afterwards.

sdk/pinocchio/src/account_info.rs
febo: Good point, added a similar note.

febo:
> The change looks good to me!
> 
> There is an edge case if someone calls `close()` and then `resize()` within the same instruction.
> 
> With the current code, `resize()` would incorrectly return an error for going over the limit, even if the account previously had some space.
> 
> How about setting the difference during `close()`? `close_unchecked()` can stay as it is, but it would be good to add a comment saying to avoid resizing afterwards.

Good point! Fixed `close()` as suggested.

sdk/pinocchio/src/account_info.rs
LStan: You added a new note saying that `zero_init` being `false` will not be supported, but you didn't remove the old one about suggesting to use `false`. They contradict each other. And the memory used to grow will not be zero-initialized after [anza-xyz/agave#5871](https://github.com/anza-xyz/agave/pull/5871)

febo: Yeah, that was intentional – this is why the note mentions "in the future." The current behaviour works and uses less CUs. The deprecation is to "warn" people that the behaviour will go away, but if you are writing a program today, you can take advantage of the `zero_init` parameter. 

Eventually `realloc` will be removed and you will only have `resize`.

febo: > They contradict each other. And the memory used to grow will not be zero-initialized after [anza-xyz/agave#5871](https://github.com/anza-xyz/agave/pull/5871)

It will be always zeroed, that is why the `zero_init` parameter is going away. You won't have the option to not zero-initialize the data.

PR #176, opened 2025-06-09
Improve program entrypoint

febo:
### Problem

The current program entrypoint does not translate to very efficient bytecode, as can be seen from an assembly implementation of entrypoint – e.g., cavey's [`asmr`](https://github.com/cavemanloverboy/asmr).

### Solution

Tweak the implementation to improve its efficiency, borrowing ideas from the assembly implementation (credits to @cavemanloverboy).

One key difference is that the entrypoint includes inlined code to parse accounts, which reduces the number of jumps required and therefore reduces CUs.

### Results

image

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -393,9 +393,10 @@ pub unsafe fn deserialize<const MAX_ACCOUNTS: usize>(
     input = input.add(size_of::<u64>());
 
     let instruction_data = { from_raw_parts(input, instruction_data_len) };
+    let input = input.add(instruction_data_len);
 
     // program id
-    let program_id: &Pubkey = &*(input.add(instruction_data_len) as *const Pubkey);
+    let program_id: &Pubkey = &*(input as *const Pubkey);
 
     (program_id, processed, instruction_data)
 }


cavemanloverboy:
i just found two more 1 CU/account optimizations, so hold your horses here. One of them might be unusable unfortunately but we will see.

febo:
> i just found two more 1 CU/account optimizations, so hold your horses here. One of them might be unusable unfortunately but we will see.

Nice – this one will go after #166 goes in.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -157,6 +163,13 @@ macro_rules! program_entrypoint {
     };
 }
 
+/// Align a pointer to the BPF alignment of `u128`.
+macro_rules! align_pointer {
+    ($ptr:ident) => {
+        (($ptr as usize + (BPF_ALIGN_OF_U128 - 1)) & !(BPF_ALIGN_OF_U128 - 1)) as *mut u8
+    };
+}
```
joncinque: Why does this align to `u128`? Shouldn't it align to `u64` per the serialization spec?

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -166,61 +179,71 @@ macro_rules! program_entrypoint {
...
+
+        input = input.add(STATIC_ACCOUNT_DATA + size_of::<u64>());
+        input = input.add((*account).data_len as usize);
+        input = align_pointer!(input);
+
+        to_process -= 1;
+
+        'remaining: while to_process > 0 {
+            // Increment the accounts pointer to the next account.
+            accounts = accounts.add(1);
```
joncinque: If `MAX_TX_ACCOUNTS` ever increases, it's possible to overwrite the accounts array, correct? it would be nice to include a check to make sure the array is enough here.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -166,61 +179,71 @@ macro_rules! program_entrypoint {
...
+            input = input.add(size_of::<u64>());
+
+            if (*account).borrow_state != NON_DUP_MARKER {
+                accounts.write(AccountInfo {
+                    raw: accounts_slice.add((*account).borrow_state as usize) as *mut Account,
+                });
+                to_process -= 1;
+                // Return to the loop to process the next account (if there is one).
+                continue 'remaining;
+            }
```
joncinque: nit: maybe it's just me, but I find it more legible to avoid the `continue` statement, ie
```suggestion
            if (*account).borrow_state != NON_DUP_MARKER {
                accounts.write(AccountInfo {
                    raw: accounts_slice.add((*account).borrow_state as usize) as *mut Account,
                });
            } else {
              // rest of logic
            }
            to_process -= 1;
```

sdk/pinocchio/src/entrypoint/mod.rs
febo: That is the name of the constant on the [SDK](https://github.com/anza-xyz/solana-sdk/blob/332ac21f02749e3341ca81f0ad5cc86c2e23d6de/program-entrypoint/src/lib.rs#L383). 😊 I think it is meant to represent the alignment of an `u128` in BPF, which is `8`. We could rename it if that makes it more clear – agree that the name is a bit confusing.

sdk/pinocchio/src/entrypoint/mod.rs
febo: That is a trick to make the compiler generate "good" assembly. 😊 This way we save about `2` CUs per account by avoiding the `else` branch, for which we would have to "jump in" and "jump out". The label is intentional to try to make it more clear.

The compiler is still doing something not optimal – @LucasSte is having a look at it.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Yeah, I was thinking that we can have one version of the entrypoint that allocates space for `255` accounts, which is the theoretical maximum given that the duplicated flag is an `u8`, and another where we add back the logic to skip accounts that we don't have space.

The reason for it is that `255` accounts take about `2KiB` of stack, which will be ok once we have dynamic stack frames, but at the moment might be too much. After dynamic stack frames, we can deprecate the one that takes an arbitrary number of accounts.

This is not yet done on this PR.

sdk/pinocchio/src/entrypoint/mod.rs
joncinque: I mainly want to make sure I'm not missing anything, but I guess you could use any `usize` whose value is equal to 8 :sweat_smile:

sdk/pinocchio/src/entrypoint/mod.rs
joncinque: Is the concern about CU usage from checking the number of accounts passed in? If so, maybe we can have an unchecked entrypoint which omits the length check

sdk/pinocchio/src/entrypoint/mod.rs
joncinque: What if we copy the `to_process -= 1;` line into both branches? That would remove the `continue` statement and should give exactly the same performance, right? If not, please add a comment explaining the code :smile:

sdk/pinocchio/src/entrypoint/mod.rs
febo: Yeah, uses less CU and simplifies the logic. We can't omit the length check if we don't use the maximum, since we still need to skip the "extra" accounts to get to the instruction data. Perhaps the best thing for now is to offer both, so you either use the maximum or specify how many accounts are expected; once dynamic stack frames are available, stack memory won't be an issue for this.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Almost the same, there is still one extra jump on the generated assembly. Perhaps with a later version of the platform tools that Lucas is working on, this code generation is improved.

joncinque: Bah, that stinks, ah well. Can you add a comment explaining the code then?

sdk/pinocchio/src/entrypoint/mod.rs
joncinque: Gotcha, in that case, can you put a comment to that effect so we know why there's no check in this code? Or offering both versions, as you mentioned

sdk/pinocchio/src/entrypoint/mod.rs
febo: Added an explanation.

febo:
@cavemanloverboy Found a few more savings with a small tweak in the code.

febo:
@joncinque Put the PR back to draft to test a suggestion from @cavemanloverboy

febo:
Current benchmark:

| Name | CUs |
|------|------|
| Account (0) | 9 |
| Account (1) | 13 |
| Account (2) | 22 |
| Account (3) | 36 |
| Account (4) | 45 |
| Account (5) | 52 |
| Account (6) | 72 |
| Account (7) | 75 |
| Account (8) | 80 |
| Account (16) | 154 |
| Account (32) | 280 |
| Account (64) | 541 |

nlgripto:
let him cook

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -200,153 +206,278 @@ macro_rules! program_entrypoint {
...
+                        process_accounts!(3 => (input, accounts, accounts_slice));
+                    }
+                } else {
+                    // 2..1 accounts left to process.
+                    if to_process > 2 {
+                        process_accounts!(2 => (input, accounts, accounts_slice));
+                    } else if to_process > 1 {
+                        process_accounts!(1 => (input, accounts, accounts_slice));
+                    }
+                }
```
illia-bobyr: *minor*

Have you considered a `match` for this?

```suggestion
                // There might be remaining accounts to process.
                match to_process {
                    5 => process_accounts!(4 => (input, accounts, accounts_slice)),
                    4 => process_accounts!(3 => (input, accounts, accounts_slice)),
                    3 => process_accounts!(2 => (input, accounts, accounts_slice)),
                    2 => process_accounts!(1 => (input, accounts, accounts_slice)),
                    1 => (),
                    _ => {
                        // SAFETY: `while` loop above makes sure that `to_process` has 1 to 5
                        // entries left.
                        unsafe { core::hint::unreachable_unchecked() }
                    }
                };
```

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -200,153 +206,278 @@ macro_rules! program_entrypoint {
...
+        input = align_pointer!(input);
+
+        if processed > 1 {
+            // The number of accounts to process (`to_process`) is limited to
+            // `MAX_ACCOUNTS`, which is the capacity of the accounts array. When there
+            // are more accounts to process than the maximum, we still need to skip the
+            // remaining accounts (`to_skip`) to move the input pointer to the instruction
+            // data. At the end, we return the number of accounts processed (`processed`),
+            // which represents the accounts initialized in the `accounts` slice.
+            let mut to_process = processed;
```
illia-bobyr: While one can infer it from the logic, the fact that `to_process` is actually "the number of accounts to process plus one" is a subtle nuance.
I think it would be good to note it in the comment, as you are already explaining the logic here.

I assume you are doing it to avoid an extra subtraction instruction?[^1]

I would even argue that it could be reasonable to put it directly in the variable name, calling it, maybe, `to_process_plus_one`.  It would remove the extra mental transformation the reader currently needs to do when they read code that uses `to_process`.

[^1]: I normally expect the compiler to be able to combine arithmetic operations and even do some more complex transformations.
For example, if `to_process` is only used within this function, which it is, the compiler can instead decrement the `match`/`if` constants you are comparing it against, and skip the decrement.

febo:
@joncinque PR updated and ready for another review. 😊

sdk/pinocchio/src/entrypoint/mod.rs
cavemanloverboy: the point of this was to manually unroll the binary search. did you verify that this match statement does **not** increase cus? if it doesn't increase it, i'm in favor of this so long as we add a comment that the compiler can figure out the binary search (in case someone changes in the future)

sdk/pinocchio/src/entrypoint/mod.rs
cavemanloverboy: i agree here, it should be free to change from `to_process` to `to_process_plus_one` via a `let to_process_plus_one = to_process;` (with appropriate coment) so long as `to_process` is unused after that

sdk/pinocchio/src/entrypoint/mod.rs
illia-bobyr: I must admit, I didn't realize it.
Maybe a short note explaining the optimization could help the future readers.

febo: A `match` statement increases CUs – in the end it generates a "standard" jump table, which in the worse case will do 4 comparisons. The "manual" one uses 3 comparisons at most, and 2 for most of the values. 

I will add a comment explaining the rationale of the nested `if` statements.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Yeah, good idea.

sdk/pinocchio/src/entrypoint/mod.rs
illia-bobyr: This is probably out of scope for the PR by now.

I started wondering if there are some alternatives to the "binary search" that have similarly good properties.
In particular, I noticed that as written we produce 16 identical code blocks for the account processing, in total.
Not sure if any of them gets optimized away, but if you are saying that the binary search tree makes it into the final code, then all the blocks are probably there as well.

The very first code block is an optimization for the case when there is only one account.
But the rest 15 are used to process accounts.
And the longest "uninterrupted" sequence of accounts we process is 5 accounts at a time.
I guess, the size of the program is less important, but it still costs something to deploy it.

On x86, jump table is a single jump: https://godbolt.org/z/f7E5P3YKs

This:
```rust
    match iterations {
        3 => { process(); process(); process(); }
        2 => { process(); process(); }
        1 => { process(); }
        0 => {}
        _ => unreachable!(),
    }
```

Turns into this:
```asm
.LBB1_1:
        lea     rax, [rip + .LJTI1_0]
        movsxd  rcx, dword ptr [rax + 4*rbx]
        add     rcx, rax
        jmp     rcx
.LBB1_3:
        call    example::process::h15029326abde9722
.LBB1_4:
        call    example::process::h15029326abde9722
.LBB1_5:
        call    example::process::h15029326abde9722
.LBB1_6:
        add     rsp, 16
        pop     rbx
        ret
.LJTI1_0:
        .long   .LBB1_6-.LJTI1_0
        .long   .LBB1_5-.LJTI1_0
        .long   .LBB1_4-.LJTI1_0
        .long   .LBB1_3-.LJTI1_0
```

There are no comparisons.
But maybe with the SBF backend can not produce computed jumps here?
The instruction set seems to have [the necessary instruction](https://github.com/7etsuo/Learn-Solana-BPF-Assembly/blob/main/docs/reference/03_instruction_set.md#jump-instructions).

If I pass the index into `process()`, then it gets a bit more confusing.
Though, `process_n_accounts!(@process_account => (input, accounts, accounts_slice))` calls are identical.  All the state change is a side effect of the call.
Though, there is a lot of code that is inlined, so the compiler might miss the fact that they are indeed identical.

cavemanloverboy: yea there's nothing like this in bpf. afaiu, in bpf the jump is to a particular labeled block (or not) if a condition is true, e.g.
```
jeq r2, 5,  process_5

process_5:
  // bpf to process 5
```

the full list of jump conditions:
image

unless call with a register can do the trick?

cavemanloverboy: example: 
image

cavemanloverboy: example 
image

cavemanloverboy: there **does** seem to be some semblance of a binary search here, looking at the `r2 > 0x2` ix which is the second ix

cavemanloverboy: it does seem to do a **decent** job.. we should try the match arm just to see what it comes up with
image

febo: Good call, it actually improved the CUs in a few cases at the sacrifice of one case:
```
| Name         | CUs | Delta |
|--------------|-----|-------|
| Account (0)  | 9   |   --  |
| Account (1)  | 13  |   --  |
| Account (2)  | 22  |   --  |
| Account (3)  | 34  |   --  |
| Account (4)  | 43  |   --  |
| Account (5)  | 51  |   +1  |
| Account (6)  | 67  |   -3  |
| Account (7)  | 71  |   -2  |
| Account (8)  | 78  |   --  |
| Account (16) | 149 |   -3  |
| Account (32) | 276 |   -2  |
| Account (64) | 538 |   --  |
```
I recall having tested a `match` statement, but perhaps it was before adding the early `if to_process == 2` test.

illia-bobyr: I was playing with godbolt a bit and this code was producing exactly the same results as running the loop `% 4` times explicitly first and then dealing with the tail using a match:

```rust
    loop {
        match iterations {
            0 => {}
            1 => { process(counter); }
            2 => { process(counter); process(counter); }
            3 => { process(counter); process(counter); process(counter); }
            4 | _ => { process(counter); process(counter); process(counter); process(counter); }
        }
        if iterations > 4 {
            iterations -= 4;
        } else {
            break;
        }
    }
```

https://godbolt.org/z/GPxP8o9o5

I think it is the best from the standpoint of readability.
It still produces 7 calls to `process`.
4 are combined in a single block, for the case when `iterations > 4`, and 3 are used with a jump table.

From the standpoint of the code size, and considering in this particular case we do not have too many accounts, I think a just table inside a loop would be better.
But I do not know how to make it produce that.

illia-bobyr: > Good call, it actually improved the CUs in a few cases at the sacrifice of one case:
> 
> ```
> | Name         | CUs | Delta |
> |--------------|-----|-------|
> | Account (0)  | 9   |   --  |
> | Account (1)  | 13  |   --  |
> | Account (2)  | 22  |   --  |
> | Account (3)  | 34  |   --  |
> | Account (4)  | 43  |   --  |
> | Account (5)  | 51  |   +1  |
> | Account (6)  | 67  |   -3  |
> | Account (7)  | 71  |   -2  |
> | Account (8)  | 78  |   --  |
> | Account (16) | 149 |   -3  |
> | Account (32) | 276 |   -2  |
> | Account (64) | 538 |   --  |
> ```
> 
> I recall having tested a `match` statement, but perhaps it was before adding the early `if to_process == 2` test.

Seems like you are doing profile guided optimization, but manually :smiley:

illia-bobyr: As for the just instructions, I'm not sure if I'm looking at the current spec, but this page says that the jump instructions could use either an immediate argument as well as a register: https://github.com/7etsuo/Learn-Solana-BPF-Assembly/blob/main/docs/reference/02_opcode_tables.md#jump-instructions
Which should allow for just table construction.

I do not see any jump tables in your disassembly.
Maybe @LucasSte knows if jump tables are supported in our current backend?

febo: > Seems like you are doing profile guided optimization, but manually 😃

Yeah, the compiler is a bit unpredictable – sometimes you change a single line and things are significantly different. 😅

Tried your second suggestion with the `match` inside the `loop`. A bit of mix results, but definitely worse for a large number of accounts.

```
| Name         | CUs | Delta |
|--------------|-----|-------|
| Account (0 ) | 9   |   --  |
| Account (1)  | 13  |   --  |
| Account (2)  | 27  |   +5  |
| Account (3)  | 33  |   -1  |
| Account (4)  | 42  |   -1  |
| Account (5)  | 58  |   +7  |
| Account (6)  | 63  |   -4  |
| Account (7)  | 68  |   -3  |
| Account (8)  | 77  |   -1  |
| Account (16) | 147 |   -2  |
| Account (32) | 287 |  +11  |
| Account (64) | 568 |  +30  |
```

LucasSte: What version of platform tools are you using? I merged the switch simplify pass recently https://github.com/anza-xyz/llvm-project/pull/153. It is available on platform tools v1.49 onwards.

LucasSte: I'd encourage you to test with v1.50, because it enables a pass to simplify branches, so you'll see even different results.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -330,21 +330,26 @@ pub unsafe fn parse(
                     to_process -= 5;
                 }
 
-                // There might be remininag accounts to process.
-                if to_process > 3 {
-                    // 4..3 accounts left to process.
-                    if to_process > 4 {
+                // There might be remaining accounts to process.
+                match to_process {
```
illia-bobyr: This one is `to_process_plus_one` as well, right?

```suggestion
                match to_process_plus_one {
```

sdk/pinocchio/src/entrypoint/mod.rs
febo: Using platform-tools `v1.50` we get even more improvements:
```
| Name         | CUs | Delta |
|--------------|-----|-------|
| Account (0)  |   9 |   --  |
| Account (1)  |  13 |   --  |
| Account (2)  |  21 |   -1  |
| Account (3)  |  34 |   --  |
| Account (4)  |  42 |   -1  |
| Account (5)  |  49 |   -1  |
| Account (6)  |  64 |   -6  |
| Account (7)  |  68 |   -5  |
| Account (8)  |  75 |   -3  |
| Account (16) | 140 |  -12  |
| Account (32) | 258 |  -20  |
| Account (64) | 501 |  -37  |
```

illia-bobyr: > Yeah, the compiler is a bit unpredictable – sometimes you change a single line and things are significantly different. 😅

Fernando The Compiler Whisperer.

LucasSte: @cavemanloverboy example generates different code in v1.50:

This is for the case with 16 match arms. The compiler builds a lookup table:

```
entrypoint:
	ldxdw r1, [r1 + 0]
	jgt r1, 16, LBB0_3
	mov64 r2, r1
	lsh64 r2, 32
	rsh64 r2, 32
	mov64 r3, 129023
	rsh64 r3, r2
	and64 r3, 1
	jeq r3, 0, LBB0_3
	lsh64 r1, 3
	lddw r2, .Lswitch.table.entrypoint
	add64 r2, r1
	ldxdw r1, [r2 + 0]
	mov64 r2, 1
	call sol_log
LBB0_3:
	mov64 r0, 0
	exit
```

sdk/pinocchio/src/entrypoint/mod.rs
febo: Good catch - fixed!

Agree, it is a bit annoying to have both `parse` and `parse_into` but there are a couple of more differences related to the skip logic [here](https://github.com/anza-xyz/pinocchio/pull/176/files#diff-8da710d703b7201cf381567ae858204192e3e6351ad82555c37606976a296f23R306) and [here](https://github.com/anza-xyz/pinocchio/pull/176/files#diff-8da710d703b7201cf381567ae858204192e3e6351ad82555c37606976a296f23R420-R422). We could split the implementation into separate macros, but I wonder if it will make it harder to follow than having two separate functions.

illia-bobyr: Splitting it into several macros may indeed decrease code readability.

What is the root cause for these differences?
If it is just the performance optimizations for certain values of the `MAX_ACCOUNTS` type argument, then, possibly, they could be handled by compile time `if` statements?

The first link points to this line:

```rust
        let mut accounts = accounts.as_mut_ptr() as *mut AccountInfo;
```
(Just after the `if processed > 0 {` check).
I think it is identical between `parse` and `parse_into`.
Wrong link, or am I missing something?

The second block, you linked, is this:
```rust
            let mut to_process_plus_one = min(processed, MAX_ACCOUNTS);
            let mut to_skip = processed - to_process_plus_one;
            processed = to_process_plus_one;
```

`to_skip` will be removed when `MAX_ACCOUNTS` is equal to `MAX_TX_ACCOUNTS` if the skip logic below is removed by a compile time check, as I was suggesting in the previous comment.

As for the `min()` - I think it actually points to a discrepancy between the two implementations.
In once case you check that you do not write outside the output `accounts` argument, and in the other - you do not.
I think it probably should be consistent.
Both versions should have the `min()` call, or none should.

If `parse` does not have the `min` call because you know that the SVM will never pass more than `MAX_TX_ACCOUNTS` accounts, then an compile time `if` statement will do it:

```rust
            // SAFETY: We know that the SVM will never pass more than `MAX_TX_ACCOUNTS` accounts to
            // the program, and so we can skip the `min()` check, if we know that the destination
            // `accounts` is big enough.
            let mut to_process_plus_one = if MAX_ACCOUNTS < MAX_TX_ACCOUNTS {
                min(processed, MAX_ACCOUNTS)
            } else {
                processed
            };
```

illia-bobyr: Here is a diff, for convenience:

```diff
@@ -1,9 +1,9 @@
-pub unsafe fn parse(
+pub unsafe fn parse_into<const MAX_ACCOUNTS: usize>(
     mut input: *mut u8,
-    accounts: &mut [MaybeUninit<AccountInfo>; MAX_TX_ACCOUNTS],
+    accounts: &mut [MaybeUninit<AccountInfo>; MAX_ACCOUNTS],
 ) -> (&'static Pubkey, usize, &'static [u8]) {
     // Number of accounts to process.
-    let processed = *(input as *const u64) as usize;
+    let mut processed = *(input as *const u64) as usize;
     // Skip the number of accounts (8 bytes).
     input = input.add(size_of::<u64>());

@@ -22,8 +22,19 @@
         input = align_pointer!(input);

         if processed > 1 {
-            // Counter for the number of accounts remaining.
-            let mut to_process_plus_one = processed;
+            // The number of accounts to process (`to_process_plus_one`) is limited to
+            // `MAX_ACCOUNTS`, which is the capacity of the accounts array. When there
+            // are more accounts to process than the maximum, we still need to skip the
+            // remaining accounts (`to_skip`) to move the input pointer to the instruction
+            // data. At the end, we return the number of accounts processed (`processed`),
+            // which represents the accounts initialized in the `accounts` slice.
+            //
+            // Note that `to_process_plus_one` includes the first (already processed)
+            // account to avoid decrementing the value. The actual number of remaining
+            // accounts to process is `to_process_plus_one - 1`.
+            let mut to_process_plus_one = min(processed, MAX_ACCOUNTS);
+            let mut to_skip = processed - to_process_plus_one;
+            processed = to_process_plus_one;

             // This is an optimization to reduce the number of jumps required
             // to process the accounts. The macro `process_accounts` will generate
@@ -59,6 +70,27 @@
                     }
                 }
             }
+
+            // Process any remaining accounts to move the offset to the instruction
+            // data (there is a duplication of logic but we avoid testing whether we
+            // have space for the account or not).
+            while to_skip > 0 {
+                // Marks the account as skipped.
+                to_skip -= 1;
+
+                // Read the next account.
+                let account: *mut Account = input as *mut Account;
+                // Adds an 8-bytes offset for:
+                //   - rent epoch in case of a non-duplicated account
+                //   - duplicated marker + 7 bytes of padding in case of a duplicated account
+                input = input.add(size_of::<u64>());
+
+                if (*account).borrow_state == NON_DUP_MARKER {
+                    input = input.add(STATIC_ACCOUNT_DATA);
+                    input = input.add((*account).data_len as usize);
+                    input = align_pointer!(input);
+                }
+            }
         }
     }
```

febo: Nice, this worked!

Added an extra check compile-time check:
```rust
const {
  assert!(
    MAX_ACCOUNTS <= MAX_TX_ACCOUNTS,
    "MAX_ACCOUNTS must be less than or equal to MAX_TX_ACCOUNTS"
  );
}
```
There is really no reason to have `MAX_ACCOUNTS` greater than `MAX_TX_ACCOUNTS`.

illia-bobyr: Cool!

Would it make sense to keep the name `parse()`, or `parse_into()` matches the existing naming convention better?

febo: Good point, reverted back to `deserialize` since it is the current name being used.

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -200,153 +183,219 @@ macro_rules! program_entrypoint {
...
-    input = input.add(core::mem::size_of::<u64>());
+    input = input.add(size_of::<u64>());
 
-    let instruction_data = { core::slice::from_raw_parts(input, instruction_data_len) };
-    input = input.add(instruction_data_len);
+    let instruction_data = { from_raw_parts(input, instruction_data_len) };
 
     // program id
-    let program_id: &Pubkey = &*(input as *const Pubkey);
+    let program_id: &Pubkey = &*(input.add(instruction_data_len) as *const Pubkey);
```
illia-bobyr: *minor*

I would argue it is better to adjust `input` value right after the next "chunk" is processed.
So, I think, previous version was better.

It makes the code paragraphs more independent.
So when you move them, or perform other kinds of changes, the chances you make a mistake are lower.

    let instruction_data = { from_raw_parts(input, instruction_data_len) };
    let input = input.add(instruction_data_len);

    // program id
    let program_id: &Pubkey = &*(input as *const Pubkey);

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -602,3 +651,162 @@ unsafe impl core::alloc::GlobalAlloc for NoAllocator {
...
+        // Number of accounts.
+        input.write(&(accounts as u64).to_le_bytes(), 0);
+        let mut offset = size_of::<u64>();
+
+        for i in 0..accounts {
+            // Account data.
+            let mut account = [0u8; STATIC_ACCOUNT_DATA + size_of::<u64>()];
+            account[0] = NON_DUP_MARKER;
+            // Set the accounts data length. The actual account data is zeroed.
+            account[80..88].copy_from_slice(&i.to_le_bytes());
```
illia-bobyr, line 739: *minor*

Is `80` the `STATIC_ACCOUNT_DATA`?

```suggestion
            let mut account = [0u8; STATIC_ACCOUNT_DATA + size_of::<u64>()];
            account[0] = NON_DUP_MARKER;
            // Set the accounts data length. The actual account data is zeroed.
            account[STATIC_ACCOUNT_DATA..(STATIC_ACCOUNT_DATA + size_off::<u64>())].copy_from_slice(&i.to_le_bytes());
```

ilya-bobyr (approved):
Did one more pass and found a few minor things.
But overall it is good :)

sdk/pinocchio/src/entrypoint/mod.rs
febo: > Is `80` the `STATIC_ACCOUNT_DATA`?

No, static data is the size of the `Account` struct (88) plus the data increase padding (10kb) – this is the minimum "data" (bytes) of an `Account` even when the account data is empty.

https://github.com/anza-xyz/pinocchio/blob/main/sdk/pinocchio/src/entrypoint/mod.rs#L36

PR #177, opened 2025-06-10
tighten docs/tests for pubkey

paahaad:
Added Docs and testcases for pubkey crate.

### DIFF
--- a/sdk/pubkey/README.md
+++ b/sdk/pubkey/README.md
@@ -48,8 +48,7 @@ pinocchio_pubkey::declare_id!("Ping111111111111111111111111111111111111111");
 
 3) Creating a `Pubkey` from a `&str`:
 ```rust
-let address = String::from("7qtAvP4CJuSKauWHtHZJt9wmQRgvcFeUcU3xKrFzxKf1");
-let owner = pinocchio_pubkey::from_str(&address);
+const OWNER: pinocchio::pubkey::Pubkey = pinocchio_pubkey::from_str("7qtAvP4CJuSKauWHtHZJt9wmQRgvcFeUcU3xKrFzxKf1");
 ```
 
 ## License
--- a/sdk/pubkey/src/lib.rs
+++ b/sdk/pubkey/src/lib.rs
@@ -4,6 +4,17 @@ pub use five8_const::decode_32_const;
 pub use pinocchio;
 
 /// Convenience macro to define a static `Pubkey` value.
+/// 
+/// This macro validates the pubkey at compile time, ensuring it's a valid
+/// base58-encoded 32-byte public key.
+/// 
+/// # Examples
+/// 
+/// ```rust
+/// use pinocchio_pubkey::pubkey;
+/// 
+/// const MY_PUBKEY: pinocchio::pubkey::Pubkey = pubkey!("11111111111111111111111111111111");
+/// ```
 #[macro_export]
 macro_rules! pubkey {
     ( $id:literal ) => {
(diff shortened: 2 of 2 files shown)


paahaad:
@febo sir, is it helpfull or should I close this?

sdk/pubkey/src/lib.rs
```diff
@@ -4,6 +4,17 @@ pub use five8_const::decode_32_const;
 pub use pinocchio;
 
 /// Convenience macro to define a static `Pubkey` value.
+/// 
+/// This macro validates the pubkey at compile time, ensuring it's a valid
```
febo: Not always in compile time – it depends on the input. The `const` qualifier allows it to be done in compile time when the input is static.

sdk/pubkey/README.md
```diff
@@ -48,8 +48,7 @@ pinocchio_pubkey::declare_id!("Ping111111111111111111111111111111111111111");
 
 3) Creating a `Pubkey` from a `&str`:
 ```rust
-let address = String::from("7qtAvP4CJuSKauWHtHZJt9wmQRgvcFeUcU3xKrFzxKf1");
-let owner = pinocchio_pubkey::from_str(&address);
+const OWNER: pinocchio::pubkey::Pubkey = pinocchio_pubkey::from_str("7qtAvP4CJuSKauWHtHZJt9wmQRgvcFeUcU3xKrFzxKf1");
```
febo: I am not sure what is the benefit of changing this example.

sdk/pubkey/src/lib.rs
```diff
@@ -13,8 +24,21 @@ macro_rules! pubkey {
 
 /// Convenience macro to define a static `Pubkey` value representing the program ID.
 ///
-/// This macro also defines a helper function to check whether a given pubkey is
-/// equal to the program ID.
+/// This macro also defines helper functions to check whether a given pubkey is
+/// equal to the program ID, providing a complete program identity management solution.
```
febo: The part of "providing a complete program identity management solution." is strange, does not read right.

sdk/pubkey/src/lib.rs
```diff
@@ -13,8 +24,21 @@ macro_rules! pubkey {
...
+/// 
+/// # Examples
+/// 
+/// ```rust
+/// use pinocchio_pubkey::declare_id;
+/// 
+/// declare_id!("11111111111111111111111111111111");
```
febo, line 36: This example needs to be more complete, e.g., show how the helper functions can be used.

sdk/pubkey/src/lib.rs
```diff
@@ -35,8 +59,51 @@ macro_rules! declare_id {
     };
 }
 
-/// Create a `Pubkey` from a `&str`.
+/// Create a `Pubkey` from a `&str` at compile time.
```
febo: Same comment about the "compile time".

sdk/pubkey/src/lib.rs
```diff
@@ -35,8 +59,51 @@ macro_rules! declare_id {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_const_pubkey_creation() {
+        const SYSTEM_PROGRAM: Pubkey = from_str("11111111111111111111111111111111");
+        // This test just ensures compilation works
+        assert_eq!(SYSTEM_PROGRAM.len(), 32);
```
febo, line 102: This needs more asserts.

febo:
> @febo sir, is it helpfull or should I close this?

It could eventually be helpful, but needs some fixes. Left some comments.

sdk/pubkey/src/lib.rs
paahaad: added more uses of helper function.

sdk/pubkey/src/lib.rs
paahaad: added more asserts with different cases

sdk/pubkey/src/lib.rs
```diff
@@ -36,7 +75,105 @@ macro_rules! declare_id {
...
+/// const SYSTEM_PROGRAM: pinocchio::pubkey::Pubkey = 
+///     pinocchio_pubkey::from_str("11111111111111111111111111111111");
+/// ```
 #[inline(always)]
 pub const fn from_str(value: &str) -> pinocchio::pubkey::Pubkey {
     decode_32_const(value)
 }
+
+/// Type alias for the Pubkey type for convenience.
+pub type Pubkey = pinocchio::pubkey::Pubkey;
```
febo, line 91: This is strange and unnecessary. You can just import the `Pubkey` type if you need it. Most likely you want to import the type for your tests.

febo:
pinocchio now uses the `Address` type from the SDK.

PR #178, opened 2025-06-11
Add cargo miri test to CI

febo:
### Problem

Currently CI is not using `cargo miri`, which can help detect undefined behaviour.

### Solution

Add a `miri` step to CI and fix raised issues.

cc: @d0nutptr


joncinque (approved):
Very nice!

d0nutptr:
Look correct to me :)

PR #179, opened 2025-06-12
Add `Copy` bound to `copy_val` helper

febo:
### Problem

Currently `copy_val` accepts `?Sized` types, which is unsound – `?Sized` allows unsized types to be used in the copy operation. Additionally, it must require that `T` is a `Copy` type to make sure the copy is allowed.

### Solution

Add `Copy` to the bound of `T` in `copy_val`.

cc: @d0nutptr

### DIFF
--- a/sdk/pinocchio/src/memory.rs
+++ b/sdk/pinocchio/src/memory.rs
@@ -47,13 +47,13 @@ pub unsafe fn sol_memcpy(dst: &mut [u8], src: &[u8], n: usize) {
 /// are of same type and `impl Copy`.
 ///
 /// This helper will be useful to optimize CU if
-/// copied size is > 32 bytes.
+/// copied size is `> 32` bytes.
 ///
-/// For value smaller than 32 bytes, `dst = src`
+/// For value smaller than `32` bytes, `dst = src`
 /// will be emitted to register assignment. For
-/// values larger than 32 bytes, compiler will
+/// values larger than `32` bytes, compiler will
 /// generate excess boilerplate to `sol_memcpy_`.
-/// So if T's size is known to be > 32 bytes,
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/memory.rs
```diff
@@ -47,33 +47,36 @@ pub unsafe fn sol_memcpy(dst: &mut [u8], src: &[u8], n: usize) {
...
+/// So if `T`'s size is known to be `> 32` bytes,
 /// this helper should be used.
 ///
 /// # Arguments
 ///
 /// - `dst` - Destination reference to copy to
 /// - `src` - Source reference to copy from
 #[inline]
-pub fn copy_val<T: ?Sized>(dst: &mut T, src: &T) {
+pub fn copy_val<T: Copy>(dst: &mut T, src: &T) {
```
d0nutptr, line 64: I believe this also needs to drop `dst` before copying over, otherwise `dst` may not be properly cleaned up

febo: When the type `T` is `Copy`, then it should not have a "custom" `Drop` right?

illia-bobyr: Seems like it is indeed not allowed.

> Error E0184: The Copy trait was implemented on a type with a Drop implementation.
> ```rust
> Erroneous code example:
> 
> #[derive(Copy)]
> struct Foo; // error!
> 
> impl Drop for Foo {
>     fn drop(&mut self) {
>     }
> }
> ```

https://doc.rust-lang.org/stable/error_codes/E0184.html

d0nutptr: Oh! Good catch. Thanks :)

PR #180, opened 2025-06-12
Fix `assign` unsoundness

febo:
### Problem

The current implementation of `AccountInfo::assign` is unsound since it takes an immutable reference to `owner` and casts it to a mutable one:
```
&(*self.raw).owner as *const _ as *mut Pubkey
```

### Solution

Use a mutable reference to `owner` directly. This PR also clarifies the safety assumptions for `owner()` and `assign()`.

cc: @d0nutptr

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -2,7 +2,7 @@
 use core::{
     marker::PhantomData,
     mem::ManuallyDrop,
-    ptr::NonNull,
+    ptr::{write, NonNull},
     slice::{from_raw_parts, from_raw_parts_mut},
 };
 
@@ -123,8 +123,8 @@ impl AccountInfo {
     ///
     /// # Safety
     ///
-    /// A reference returned by this method is invalidated when [`Self::assign`]
-    /// is called.
+    /// It is undefined behaviour to use [`Self::assign`] while there is an active reference
+    /// to the `owner` returned by this method.
     #[inline(always)]
     pub unsafe fn owner(&self) -> &Pubkey {
         &(*self.raw).owner
@@ -173,18 +173,18 @@ impl AccountInfo {
     /// Checks if the account is owned by the given program.
     #[inline(always)]
     pub fn is_owned_by(&self, program: &Pubkey) -> bool {
-        unsafe { &(*self.raw).owner == program }
+        unsafe { self.owner() == program }
     }
 
     /// Changes the owner of the account.
     ///
     /// # Safety
     ///
-    /// Using this method invalidates any reference returned by [`Self::owner`].
+    /// It is undefined behaviour to use this method while there is an active reference
+    /// to the `owner` returned by [`Self::owner`].
     #[inline(always)]
     pub unsafe fn assign(&self, new_owner: &Pubkey) {
-        #[allow(invalid_reference_casting)]
-        core::ptr::write_volatile(&(*self.raw).owner as *const _ as *mut Pubkey, *new_owner);
+        write(&mut (*self.raw).owner, *new_owner);
     }
 
     /// Return true if the account borrow state is set to the given state.


sdk/pinocchio/src/account_info.rs
```diff
@@ -123,8 +123,8 @@ impl AccountInfo {
     ///
     /// # Safety
     ///
-    /// A reference returned by this method is invalidated when [`Self::assign`]
-    /// is called.
+    /// It is undefined behaviour to use [`Self::assign`] while there is an active reference
+    /// to the `owner` returned by this method.
     #[inline(always)]
     pub unsafe fn owner(&self) -> &Pubkey {
```
febo: I wonder if we should revert the `unsafe` here. The issue is with `assign()`, so having that one as unsafe might be sufficient.

joncinque: I'd suggest keeping it as `unsafe` to make it clear that there are unsound ways to use the function

joncinque (dismissed):
Looks great!

sdk/pinocchio/src/account_info.rs
illia-bobyr: Seems correct to me.

---

I think if you change the return type to `Pubkey`, then both functions become safe.
According to the ["Pointer to reference conversion"](https://doc.rust-lang.org/core/ptr/index.html#pointer-to-reference-conversion), only the lifetimes of the references matter.

And if `owner()` does not return a reference, then the lifetimes of both the shared and the mutable references to the `self.raw.owner` bytes are contained within the `owner()` and `assign()` function calls respectively.

I think `AccountInfo` should already be `!Send`, due to the `*mut T` [being `!Send`](https://doc.rust-lang.org/std/marker/trait.Send.html#impl-Send-for-*mut+T).
And so, it is impossible to construct two simultaneously alive references to the `owner` memory then.

illia-bobyr: > I'd suggest keeping it as `unsafe` to make it clear that there are unsound ways to use the function

One problem is that if you have too many `unsafe` blocks, you tend to pay less attention when checking each one of them.
Assuming it is much more common to call `owner()`, by removing `unsafe` we reduce the mental burden of using this API.
And in turn, if someone uses `assign()` then now will hopefully pay more attention, when verifying that the `assign()` call they are performing is correct.

After all, they only need verification of the `assign()` invocations.
If they check that it works correctly, then everything is fine.

At the same time, an addition of a new `owner()` call to an existing, undefined behavior free, code base may now cause undefined behavior.
So, maybe both indeed need to be `unsefe`, as long as `owner()` returns a reference and there is no runtime checking.

stegaBOB: If assign didn't exist, owner would be safe. I'm in favor of marking owner as safe as well.

d0nutptr: I misread this earlier, thinking this was `assign` :) 

I believe the removal of `unsafe` in the signature is fine here. It's `assign` that has the stricter guarantees it must uphold.

febo: Ok, latest commit remove the `unsafe` on owner. UB is introduced if you use `assign` with an active owner reference, so `assign` being `unsafe` should be enough.

illia-bobyr: I think Jon is right in suggesting you may want to keep it.
It is possible to add an `owner()` call to cause an undefined behavior, right?

I think if you really want to remove `unsafe` here, it would be better to return `Pubkey` by value.
I would think, you would actually get exactly the same binary in most cases.
And it is guaranteed that you can not have UB.

joncinque (approved):

sdk/pinocchio/src/account_info.rs
febo: > I think Jon is right in suggesting you may want to keep it.

Spoke to Jon offline and he is fine with removing the `unsafe` here.

> It is possible to add an `owner()` call to cause an undefined behavior, right?

The only situation that this can happen is that if you add the call right before  (in the same scope) `assign`, which will cause UB in `assign`; adding it after is fine.

> I think if you really want to remove `unsafe` here, it would be better to return `Pubkey` by value. I would think, you would actually get exactly the same binary in most cases. And it is guaranteed that you can not have UB.

The problem is that you are copying the value, which uses more CU.

illia-bobyr: > > It is possible to add an `owner()` call to cause an undefined behavior, right?
> 
> The only situation that this can happen is that if you add the call right before (in the same scope) `assign`, which will cause UB in `assign`; adding it after is fine.

It is UB for a mutable reference to have an overlapping lifetime with a shared reference, if we consider just the basic rules.
So the order should not matter.
It is probably not very likely that you would want to preserve the reference you get from the `owner()` call, but you may.

For example, like this:

```rust
let owner = account.owner();

if owner == one_id() {
    account.assign(get_new_owner());
}

if owner == second_id() {
    /* do something else */
}
```

If initially you did not have the second check, the program is correct.
After you add the `owner == second_id()` check, it has UB.

> > I think if you really want to remove `unsafe` here, it would be better to return `Pubkey` by value. I would think, you would actually get exactly the same binary in most cases. And it is guaranteed that you can not have UB.
> 
> The problem is that you are copying the value, which uses more CU.

Have you actually seen that it affects the produced code?
I understand that this is not a trivial question.
I just wonder if you have already done some research here.

Here is a godbolt compilation that shows identical results for a version that uses references and a version that passes data by value, when compiled with O3: https://rust.godbolt.org/z/9ra85bqeE
It does not mean that every program will be compiled like this, of course.
Godbolt does not support any of the stock Rust BPF backends, and we actually have our own SBF backend.
Plus, everything is inlined in this case.
Yet, the diff shows no changes at all, except for some minor layout differences in `main()`.

With regard to the different backends, I think most of the relevant optimizations are probably happening before the code generation.
But the inlining - this part is important.
In a different application inlining might happen differently, and you get a different result.
I've noticed that you add `#[inline(always)]` annotations everywhere.
So, I think, the chances that it would be processed identically are actually very high.

Most of the time, I would imagine, what you do to the `owner()` result is that you compare the content.
Which would require the same instructions, regardless of if you got it by value or by reference[^1].
If the compiler is able to see the source of the key bytes, it should compare directly to that bytes.  There is no reason to copy them, nor to copy the pointer itself.

[^1]: If you remove -O3 from the example above, you can actually see that while the by reference version skips the copy, it inserts an extra function that does the comparison.

febo: > It is UB for a mutable reference to have an overlapping lifetime with a shared reference, if we consider just the basic rules. So the order should not matter. It is probably not very likely that you would want to preserve the reference you get from the `owner()` call, but you may.

Yeah, this is what I was trying to say – if the call to the `owner()` is before the call to `assign()` and the reference stays "alive". Your exemple is definitely more clear than my description. 😅

> Have you actually seen that it affects the produced code? I understand that this is not a trivial question. I just wonder if you have already done some research here.

Here is a part of the assembly produced on a program with your code example:

* Returning a reference:
```
LBB0_13:
	ldxdw r1, [r2 + 0]
	add64 r2, r1
	add64 r2, 8
	ldxdw r6, [r10 - 512]
	add64 r6, 40
	mov64 r1, r6
	mov64 r3, 32
	call memcmp
	lsh64 r0, 32
	rsh64 r0, 32
	jeq r0, 0, LBB0_15
```
* Returning a copy:
```
LBB0_13:
	ldxdw r1, [r2 + 0]
	ldxdw r6, [r10 - 544]
	ldxdw r3, [r6 + 40]
	stxdw [r10 - 32], r3
	ldxdw r3, [r6 + 48]
	stxdw [r10 - 24], r3
	ldxdw r3, [r6 + 56]
	stxdw [r10 - 16], r3
	ldxdw r3, [r6 + 64]
	stxdw [r10 - 8], r3
	add64 r2, r1
	add64 r2, 8
	mov64 r1, r10
	add64 r1, -32
	mov64 r3, 32
	call memcmp
	lsh64 r0, 32
	rsh64 r0, 32
	jeq r0, 0, LBB0_15
```
On the second listing you can see the copy happening.

illia-bobyr: Considering `memcmp` is comparing `r3` bytes with addresses in `r1` and `r2`, the copy here is redundant.
Especially, when everything is already inlined within a single block.
If I understand correctly, we copy the key from `r6 + 40` to `r10 - 32` and then use `r10 - 32` for comparison.
We could have just used `r6 + 40`.
Which would essentially give us the first version, just with a different order of register assignments before the call.

I do not want to block this PR, so totally feel free to merge it.
But I wonder if this is something @LucasSte would want to look into?

LucasSte: > Which would essentially give us the first version, just with a different order of register assignments before the call.

I didn't follow your point. If you return a copy, one would expect the compiler to actually copy the bytes, right? Are you suggesting that we should not actually copy in that case?

I checked the code in the godbolt site. In the copy example, we indeed copy the return value to the stack, according to the call convention. There is actually a limit the compiler use to decide whether to promote a copy argument/return value to a reference argument/return value. Certainly that can be lowered.

LucasSte: > Here is a godbolt compilation that shows identical results for a version that uses references and a version that passes data by value, when compiled with O3: https://rust.godbolt.org/z/9ra85bqeE

The x86 has enough registers to hold the both keys, so the copy in godbolt happens implicitly (memory -> registers -> then compare). This is why both codes are the same. We do not have as many registers, so stack spills are more common.

illia-bobyr: > > Which would essentially give us the first version, just with a different order of register assignments before the call.
> 
> I didn't follow your point. If you return a copy, one would expect the compiler to actually copy the bytes, right?
> Are you suggesting that we should not actually copy in that case?

I would expect it to avoid copying bytes if it is unnecessary.
There are many optimizations that ensure unnecessary copies are not performed.
I think they are quite essential in supporting the modern style of writing in C++ and in Rust.
I only expect the result of the assembly execution to match what the Rust code is suggesting should happen.
My understanding is that if the assembly matched the higher level language 1 to 1, there would be no reasons for the undefined behavior to exist.

> I checked the code in the godbolt site. In the copy example, we indeed copy the return value to the stack, according to the call convention. There is actually a limit the compiler use to decide whether to promote a copy argument/return value to a reference argument/return value. Certainly that can be lowered.

We often deal with addresses that are 32 bytes in size.
If there is a parameter that would allow the compiler to avoid unnecessary copies of objects of this size, I think, it would make sense to look into adjusting it.

> The x86 has enough registers to hold the both keys, so the copy in godbolt happens implicitly (memory -> registers -> then compare). This is why both codes are the same. We do not have as many registers, so stack spills are more common.

I do not think we are comparing values in registers.

Here is how I understand the second code block example that Febo shared:

```gas
LBB0_13:
    # I do not fully understand this assignment.  I'm not familiar with the SVM
    # calling conventions.  But, eventually, `*(r2 + 0) * 2 + 8` is the target
    # Pubkey we are comparing the account owner Pubkey with.  `r2` is only used
    # to compute this value.
    ldxdw r1, [r2 + 0]

    # `r10` is probably stack?
    # We access data with rather big offsets relative to `r10` in both code
    # versions.  And we subtract from it, which matches how stack stack pointer
    # is normally used.
    # `r6` is probably `self` we got from the caller.
    ldxdw r6, [r10 - 544]

    # Copy 32 bytes from `r6 + 40` to `r10 - 32`.
    # This is going to be the second argument to our `memcmp` call.
    ldxdw r3, [r6 + 40]
    stxdw [r10 - 32], r3
    ldxdw r3, [r6 + 48]
    stxdw [r10 - 24], r3
    ldxdw r3, [r6 + 56]
    stxdw [r10 - 16], r3
    ldxdw r3, [r6 + 64]
    stxdw [r10 - 8], r3
    # At this point `r10 - 32` and `r6 + 40` point to identical 32 bytes.

    # Prepare a `memcmp` call.
    # `r2` is `*(r2 + 0) * 2 + 8`
    add64 r2, r1
    add64 r2, 8
    # `r1` is `r10 - 32`
    mov64 r1, r10
    add64 r1, -32
    # `r3` is `32`
    mov64 r3, 32
    call memcmp

    # Jump if `memcmp` returned zero in the lower 32 bits of `r0`.
    lsh64 r0, 32
    rsh64 r0, 32
    jeq r0, 0, LBB0_15
```

There is a redundant copy here.
We copy 32 bytes from `r6 + 40` to `r10 - 32`, and then call `memcmp(r10 - 32, r1 + 8, 32)`.
We can just call `memcmp(r6 + 40, r1 + 8, 32)` to get the same result.
This copy is redundant.
It does not seem to matter how many registers you have.

LucasSte: Dereferencing a pointer implies loading it to a register or copying the value if it does not fit there.

I understand what you are pointing in the SBF target. Since everything is inlined, the copy to the stack appears unnecessary.

I want to bring here that in the example in godbolt compiled for x86 or aarch64, you see the same code for both the copied version and the reference version. That happens because both targets have enough registers to hold two pub keys and then compare them with compare instructions. For them, you actually see the pub keys being loaded in registers and then compared. That won't change depending on the case, since the machine instructions are actually the same (if you pass by reference, you'll also need to load them to registers before the compare instructions).

In the SBF target, there is not enough registers to hold the pubkeys. Regardless of whether we emit compares or the syscall, the copy will be explicit. (There is threshold defined for when to invoke the syscall and when to compare with instructions).

If we want to extrapolate that and do a fair comparison, you can set the size of the pub key to 1024 bytes. The copied version (in x86 and aarch64) will have a call to memcpy (dereferencing the pointer) and another call to memcmp, even though you could just forward the pointer to memcmp. See https://godbolt.org/z/z3GaeddEs.

illia-bobyr: Thank you for the explanation.
I still do not fully understand the mechanics, but it is indeed happening as you are saying it does.

A diff is a bit easier for me to read.
Here is a diff of the two versions with the Pubkey size increased to 1024: https://rust.godbolt.org/z/33raodq7M
It is pretty strange that it first uses all the xmm registers to copy everything in memory, and then does an unnecessary `memcpy` call for the "by value" case.

LucasSte: I believe what's causing the copy is not the "by value" or "by copy" case. Every argument or return passed by value whose size is larger than 128 bits is promoted to a reference.

The pointer dereference is triggering a copy. For example, if you do `let a = *(ptr as *const u64)`, you are copying the value of the pointer to a variable. Likewise, `(*self.raw).owner` triggers a copy.

illia-bobyr: > I believe what's causing the copy is not the "by value" or "by copy" case. Every argument or return passed by value whose size is larger than 128 bits is promoted to a reference.
> 
> The pointer dereference is triggering a copy. For example, if you do `let a = *(ptr as *const u64)`, you are copying the value of the pointer to a variable. Likewise, `(*self.raw).owner` triggers a copy.

`self.raw` is a pointer to an `Account`, while `(*self.raw).owner` is a `Pubkey`.
We see an unnecessary copy of the `Pubkey`.
You showed me an example where an unnecessary copy of a `Pubkey` was elided in BPF.
Though, your example could have been a special case of the return value optimization (at least this is what it is called in C++).

Assuming the compiler can avoid copying an object of the size of `Pubkey` in a general case.  I would appreciate it if you could help me understand what operation in this particular example might be preventing the optimizations to kick in.

d0nutptr:
lgtm

PR #181, opened 2025-06-16
Improve sysvar get error handling

febo:
### Problem

Currently the implementation of `impl_sysvar_get` macro and `get_sysvar` return a `e.into()` when the syscall fails to return `SUCCESS`. This generates `ProgramError::Custom(_)` errors, although there are no custom errors associated with them.

### Solution

There are only [two](https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/sysvar.rs#L228-L239) situations where a sysvar get will return an error code:
1. if you pass a wrong combination of length and offset, which tries to copy data outside the sysvar length
2. If you pass an invalid sysvar id

This PR updates the error handling to return a `ProgramError::InvalidArgument` for the first case and `ProgramError::UnsupportedSysvar` for the second.

### DIFF
--- a/sdk/pinocchio/src/sysvars/mod.rs
+++ b/sdk/pinocchio/src/sysvars/mod.rs
@@ -11,7 +11,7 @@ pub mod fees;
 pub mod instructions;
 pub mod rent;
 
-/// Return value indicating that the  `offset + length` is greater than the length of
+/// Return value indicating that the `offset + length` is greater than the length of
 /// the sysvar data.
 //
 // Defined in the bpf loader as [`OFFSET_LENGTH_EXCEEDS_SYSVAR`](https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/sysvar.rs#L172).
@@ -93,9 +93,9 @@ pub unsafe fn get_sysvar_unchecked(
         match result {
             crate::SUCCESS => Ok(()),
             OFFSET_LENGTH_EXCEEDS_SYSVAR => Err(ProgramError::InvalidArgument),
-            // An invalid sysvar id and unexpected errors are folded
-            // into `UnsupportedSysvar`.
-            SYSVAR_NOT_FOUND | _ => Err(ProgramError::UnsupportedSysvar),
+            SYSVAR_NOT_FOUND => Err(ProgramError::UnsupportedSysvar),
+            // Unexpected errors are folded into `UnsupportedSysvar`.
+            _ => Err(ProgramError::UnsupportedSysvar),
         }
     }
 


sdk/pinocchio/src/sysvars/mod.rs
```diff
@@ -11,6 +11,11 @@ pub mod fees;
 pub mod instructions;
 pub mod rent;
 
+/// Return value to indicate that the  `offset + length` is greater than the length of
+/// the sysvar data.
+#[cfg(target_os = "solana")]
+const OFFSET_LENGTH_EXCEEDS_SYSVAR: u64 = 1;
```
illia-bobyr: When I read "to indicate", it seems as if the error value is being defined here.
Maybe use "that indicates"?
Also, the PR description has a reference to the definition site of this constant.
I think it makes sense to have this information in the comment here.

```suggestion
/// Return value that indicates that the  `offset + length` is greater than the length of
/// the sysvar data.
///
/// Defined in the bpf loader (`agave/programs/bpf_loader/src/syscalls`) as
/// `OFFSET_LENGTH_EXCEEDS_SYSVAR`.
#[cfg(target_os = "solana")]
const OFFSET_LENGTH_EXCEEDS_SYSVAR: u64 = 1;
```

Could be in an ordinary comment as well, rather than the doc comment, depending on if you think it should be exposed in the docs:

```rust
/// Return value that indicates that the  `offset + length` is greater than the length of
/// the sysvar data.
// Defined in the bpf loader (`agave/programs/bpf_loader/src/syscalls`) as
// `OFFSET_LENGTH_EXCEEDS_SYSVAR`.
#[cfg(target_os = "solana")]
const OFFSET_LENGTH_EXCEEDS_SYSVAR: u64 = 1;
```

sdk/pinocchio/src/sysvars/mod.rs
```diff
@@ -76,7 +83,8 @@ pub unsafe fn get_sysvar_unchecked(
 
         match result {
             crate::SUCCESS => Ok(()),
-            e => Err(e.into()),
+            OFFSET_LENGTH_EXCEEDS_SYSVAR => Err(ProgramError::InvalidArgument),
+            _ => Err(ProgramError::UnsupportedSysvar),
```
illia-bobyr: *minor*

I wonder if `_` is the right choice for the `SYSVAR_NOT_FOUND` value.
It seems reasonable to fail with the `UnsupportedSysvar` for an unexpected error.
But maybe mentioning `SYSVAR_NOT_FOUND` explicitly would be better?
I do not know if the following will compile without warnings:

```suggestion
            OFFSET_LENGTH_EXCEEDS_SYSVAR => Err(ProgramError::InvalidArgument),
            SYSVAR_NOT_FOUND | _ /* Unexpected errors are folded into `UnsupportedSysvar` */
              => Err(ProgramError::UnsupportedSysvar),
```

Also, obviously, need to add a definition of `SYSVAR_NOT_FOUND`.

sdk/pinocchio/src/sysvars/mod.rs
febo: Your suggestion works, just tweaked a bit the comment.

sdk/pinocchio/src/sysvars/mod.rs
febo: Your suggestion works, just tweaked a bit the comment.

ilya-bobyr (dismissed):

sdk/pinocchio/src/sysvars/mod.rs
```diff
@@ -76,7 +92,10 @@ pub unsafe fn get_sysvar_unchecked(
 
         match result {
             crate::SUCCESS => Ok(()),
-            e => Err(e.into()),
+            OFFSET_LENGTH_EXCEEDS_SYSVAR => Err(ProgramError::InvalidArgument),
+            SYSVAR_NOT_FOUND => Err(ProgramError::UnsupportedSysvar),
```
febo, line 96: @ilya-bobyr Simplified the match arm since clippy complained when I made a similar change on the SDK – likely we don't have all warnings enabled in this repo, but since we should fix that it is better to update the code.

ilya-bobyr (approved):

PR #182, opened 2025-06-17
system: Fix signer flag

febo:
### Problem

Current both `allocate_with_seed` and `assign_with_seed` mark the first account as a writable signer, while the program only expects the account to be writable.

### Solution

Update the instruction builders to mark the account as writable only.

cc: @d0nutptr


d0nutptr:
LGTM :)

ilya-bobyr (approved):

joncinque (approved):

rustopian (approved):

PR #183, opened 2025-06-18
Add rust-toolchain.toml

rustopian:


rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.86.0"
```
febo: We probably want this version to match the one from [Cargo.toml](https://github.com/anza-xyz/pinocchio/blob/main/Cargo.toml#L18). Currently is set to rust `1.79.0` since it is the one supported by platform-tools `v1.43` but we are already using rust `1.84.1` (platform-tools `v1.48`) in other parts, including CI - we could bump everything to `1.84.1`.

Given that pinocchio is meant to be used on-chain, we should match the version to the "target" Solana CLI/platform-tools, otherwise the build might fail.

rustopian: Didn't see that, makes sense, thanks. FYI [solana-sdk is at 1.85.1](https://github.com/anza-xyz/solana-sdk/blob/master/rust-toolchain.toml)

febo: The sdk has a slightly different setup since crates are also used off-chain. There is a script that checks on-chain dependencies ([check-msrv.sh](https://github.com/anza-xyz/solana-sdk/blob/master/scripts/check-msrv.sh)) by building a program and it is currently set at `1.81.0` here: https://github.com/anza-xyz/solana-sdk/blob/master/program/Cargo.toml#L12

febo (approved):
Thanks!

PR #184, opened 2025-06-19
token: Fix instruction data length

febo:
### Problem

The helpers `initialize_mint`, `initialize_mint_2` and `set_authority` for the token program include an optional `Pubkey` in their instruction data. When  the optional pubkey ise not present, the length of the instruction data is not being updated, which creates an UB situation.

### Solution

Adjust the instruction data length according to whether the optional pubkey is present or not.

cc: @d0nutptr


d0nutptr:
Looks good to me!

joncinque (approved):
Looks great!

PR #186, opened 2025-07-01
log: Add unsafe to Log trait

febo:
### Problem

The `Log` trait has memory safety implications in its implementation, but it is not marked as `unsafe`. The implementation must ensure that the value returned by any of the methods correctly reflects the actual number of bytes written to the buffer. Returning a value greater than the number of bytes written to the buffer will result in undefined behavior, since it will lead to reading uninitialized memory from the buffer.

### Solution

Add `unsafe` to `Log` trait to make it explicit that the trait has to be carefully implemented.

cc: @d0nutptr

### DIFF
--- a/sdk/log/crate/src/logger.rs
+++ b/sdk/log/crate/src/logger.rs
@@ -191,7 +191,14 @@ pub enum Argument {
 }
 
 /// Trait to specify the log behavior for a type.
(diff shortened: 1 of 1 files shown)


d0nutptr:
LGTM

PR #187, opened 2025-07-01
Expand docs on InstructionContext

febo:
### Problem

The lazy entrypoint uses an `InstructionContext`, which is created passing the input parameters given by the SVM loader. Apart from this observation, the safety note does not explain the alignment requirements nor provide more information about the input paramters.

### Solution

Expand the docs to include details of the alignment requirements, the input parameters expected and link to external documentation.

cc: @d0nutptr


d0nutptr:
LGTM

PR #188, opened 2025-07-01
log: Calculate maximum digits for `Log` implementation

febo:
### Problem

Currently the `Log` implementation for unsigned integer types specify the maximum number of digits that a value can have manually.

### Solution

Simplify the implementation by calculating the value automatically:
```rust
const MAX_DIGITS: usize = const { <type>::MAX.ilog10() as usize + 1 };
```
This does not have a performance implication since `MAX_DIGITS` is calculated at compile time.

cc: @d0nutptr

### DIFF
--- a/sdk/log/crate/src/logger.rs
+++ b/sdk/log/crate/src/logger.rs
@@ -224,7 +224,7 @@ macro_rules! impl_log_for_unsigned_integer {
             #[inline]
             fn write_with_args(&self, buffer: &mut [MaybeUninit<u8>], args: &[Argument]) -> usize {
                 // The maximum number of digits that the type can have.
-                const MAX_DIGITS: usize = const { $type::MAX.ilog10() as usize + 1 };
+                const MAX_DIGITS: usize = $type::MAX.ilog10() as usize + 1;
 
                 if buffer.is_empty() {
                     return 0;


sdk/log/crate/src/logger.rs
```diff
@@ -212,10 +212,13 @@ pub trait Log {
 
 /// Implement the log trait for unsigned integer types.
 macro_rules! impl_log_for_unsigned_integer {
-    ( $type:tt, $max_digits:literal ) => {
+    ( $type:tt ) => {
         impl Log for $type {
             #[inline]
             fn write_with_args(&self, buffer: &mut [MaybeUninit<u8>], args: &[Argument]) -> usize {
+                // The maximum number of digits that the type can have.
+                const MAX_DIGITS: usize = const { $type::MAX.ilog10() as usize + 1 };
```
joncinque: nit: i don't think this makes a difference, but is the second `const` needed? I was able to do this in a test, outside of a macro:
```suggestion
                // The maximum number of digits that the type can have.
                const MAX_DIGITS: usize = $type::MAX.ilog10() as usize + 1;
```

joncinque (approved):
Looks great!

sdk/log/crate/src/logger.rs
febo: You are right, I think we can omit the second `const`.

joncinque (dismissed):

febo:
@joncinque Rebased.

joncinque (approved):

PR #189, opened 2025-07-01
Add manual default implementation to `Rent` sysvar

febo:
### Problem

Currently the `Rent` sysvar includes a default implemention using the `Default` derive. This makes all values to be set to `0`, which differs from the default values used in the SDK.

### Solution

This PR adds a manual implementation for the `Default` trait with the expected default values.

### DIFF
--- a/sdk/pinocchio/src/sysvars/rent.rs
+++ b/sdk/pinocchio/src/sysvars/rent.rs
@@ -52,7 +52,7 @@ pub const ACCOUNT_STORAGE_OVERHEAD: u64 = 128;
 
 /// Rent sysvar data
 #[repr(C)]
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct Rent {
     /// Rental rate in lamports per byte-year
     pub lamports_per_byte_year: u64,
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Does this need a `Default` implementation at all? It's a bit dangerous to have it, especially because we're considering changing it for https://github.com/solana-foundation/solana-improvement-documents/pull/194

I'd probably recommend removing it, but if it is needed, it should match the sdk's version

febo:
> Does this need a `Default` implementation at all? It's a bit dangerous to have it, especially because we're considering changing it for [solana-foundation/solana-improvement-documents#194](https://github.com/solana-foundation/solana-improvement-documents/pull/194)
> 
> I'd probably recommend removing it, but if it is needed, it should match the sdk's version

Good point! At the moment it is needed since the `Sysvar` [trait](https://github.com/anza-xyz/pinocchio/blob/main/sdk/pinocchio/src/sysvars/mod.rs#L28) requires it. But we can change that as well, since it is not really required.

I made the `Sysvar` change locally and everything else works.

joncinque (approved):
Looks great!

PR #192, opened 2025-07-07
pubkey: Add `derive_address` helper

febo:
### Problem

Even when a program stores a bump to derive a program address, it is necessary to use the `create_program_address` syscall to validate the derivation. The base cost of the syscall is `1500` CUs.

In most cases, the program has the correct seeds for the derivation, so it would be sufficient to just perform the derivation and compare it against the expected resulting address.

### Solution

This PR adds a new helper `derive_address` that Instead of using the `create_program_address`, it uses the `sol_sha256` syscall to perform the derivation. This significantly reduces the amount of CUs consumed.

To derive a PDA with seeds `&[program_id, &[255]]`:
| function       | CUs |
|------------| --- |
| `create_program_address` | 🟥 1519 (+1363) |
| (*new*) `derive_address`  |  🟩 156 |

There is also compile-time constant version `derive_address_const`.

Since this functionality requires a change of dependency, the PR includes a refactoring to add the new dependency behind a feature flag – although the flag is enabled by default.

> [!NOTE]
> The `derive_address` and `derive_address_const` do not validate if the address is a valid PDA (i.e., off-curve address).

### DIFF
--- a/sdk/pubkey/src/lib.rs
+++ b/sdk/pubkey/src/lib.rs
@@ -1,6 +1,8 @@
 #![no_std]
 
 #[cfg(feature = "const")]
+#[doc(hidden)]
+// Re-export dependencies used in macros.
 pub mod reexport {
     pub use pinocchio::pubkey::Pubkey;
 }


sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+
+/// Derive a [program address][pda] from the given seeds, bump and program id.
+///
+/// [pda]: https://solana.com/docs/core/pda
+///
+/// This function differs from [`pinocchio::pubkey::create_program_address`] in that it
+/// does not perform a validation to ensure that the derived address is a valid
+/// (off-curve) program derived address. It is intended for use in cases where the
+/// seeds, bump, and program id are known to be valid, and the caller wants to derive
+/// the address without incurring the cost of the `create_program_address` syscall.
```
illia-bobyr, line 38: *subjective opinion*

I think it could be helpful if you mention the performance improvements here, in the function doc.

Most users are not going to see the CU numbers from the PR description.
And, as a user, I'm deciding if the performance advantage here is worth the safety check I'm losing when I use `derive_address` instead of `create_progra_address`.

One can argue that if you know that your bump value is correct, the check is redundant in any case.
But when you would specifically tell me that the check is responsible for 90% of the CU cost of a `create_progra_address` call, it might affect my decision quite a lot.

Essentially, I suggest that you put the PR description into this function doc.

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+pub fn derive_address<const N: usize>(seeds: &[&[u8]; N], bump: u8, program_id: &Pubkey) -> Pubkey {
+    const {
+        if N > MAX_SEEDS {
+            panic!("number of seeds must be less than `MAX_SEEDS`");
+        }
```
illia-bobyr:
```suggestion
        assert!(N <= MAX_SEEDS, "number of seeds must be less than `MAX_SEEDS`");
```

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+
+    let bump = [bump];
+
+    // SAFETY: `data` is guaranteed to have enough space for `MAX_SEEDS + 3`
+    // elements, and `MAX_SEEDS` is as large as `N`.
+    unsafe {
+        data.get_unchecked_mut(i).write(bump.as_ref());
+        data.get_unchecked_mut(i + 1).write(program_id.as_ref());
+        data.get_unchecked_mut(i + 2).write(PDA_MARKER.as_ref());
+    }
```
illia-bobyr: *personal education*

I should learn how you check the CU usage of a snippet.
But if you still want to check an alternative version yourself, here is a possible suggestion.

I wonder if it is possible to avoid dealing with `MaybeUninit` and some of the unsafe functions from it by using `array::from_fn`.

```rust
    use core::array;

    let bump = [bump];

    let data: [&[u8]; N + 3] = array::from_fn(|i| {
        if i < N {
            // SAFETY: `seeds` has `N` elements.
            unsafe { seeds.get_unchecked(i) }
        } else if i == N {
            &bump
        } else if i == N + 1 {
            &program_id
        } else if i == N + 2 {
            &PDA_MARKER
        }
    });
```

I would imagine the compiler should be able to remove any zero initialization and be able to directly initialize all the elements.  But I have not checked it.

sdk/pubkey/src/lib.rs
buffalojoec: Consider a `WARN:` here? Maybe it's gratuitous...

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+                pda.as_mut_ptr() as *mut _ as *mut u8,
+            );
+        }
+
+        // SAFATE: `pda` has been initialized by the syscall.
+        unsafe { pda.assume_init() }
+    }
+
+    #[cfg(not(target_os = "solana"))]
+    unreachable!("deriving a pda is only available on target `solana`");
```
illia-bobyr, line 88: Most other functions in `pinnochio` use `black_box` on all the arguments and do nothing for platform dependent code.
`create_program_address()` is one exception as it panics.

In any case, I think a `panic!` might be a better macro to use here.
As on non-Solana platform this spot is reachable.

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+        let mut pda = MaybeUninit::<[u8; 32]>::uninit();
+
+        // SAFETY: `data` has `i + 3` elements initialized.
+        unsafe {
+            sol_sha256(
+                data.as_ptr() as *const _ as *const u8,
+                (i + 3) as u64,
+                pda.as_mut_ptr() as *mut _ as *mut u8,
+            );
+        }
```
illia-bobyr: *minor*

You can use `N` instead of `i` here, I think?
Not sure if it would affect the compiler output.

Also, a single pointer cast should be enough, no?

```suggestion
        // SAFETY: `data` has `i + 3` elements initialized.
        unsafe {
            sol_sha256(
                data.as_ptr() as *const u8,
                (i + 3) as u64,
                pda.as_mut_ptr() as *mut u8,
            );
        }
```

I think clippy notices unnecessary casts.

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+
+        // SAFATE: `pda` has been initialized by the syscall.
```
buffalojoec:
```suggestion
        // SAFETY: `pda` has been initialized by the syscall.
```

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+/// does not perform a validation to ensure that the derived address is a valid
+/// (off-curve) program derived address. It is intended for use in cases where the
+/// seeds, bump, and program id are known to be valid, and the caller wants to derive
+/// the address without incurring the cost of the `create_program_address` syscall.
+pub fn derive_address<const N: usize>(seeds: &[&[u8]; N], bump: u8, program_id: &Pubkey) -> Pubkey {
+    const {
+        if N > MAX_SEEDS {
+            panic!("number of seeds must be less than `MAX_SEEDS`");
+        }
+    }
```
buffalojoec: You aren't checking the length of any one seed.

https://github.com/anza-xyz/solana-sdk/blob/78daa206c2f9b6068da4c67703f9e0579490ec48/pubkey/src/lib.rs#L906-L908

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,118 @@
...
+use core::mem::MaybeUninit;
+
+use pinocchio::pubkey::{Pubkey, MAX_SEEDS, PDA_MARKER};
+#[cfg(target_os = "solana")]
+use pinocchio::syscalls::sol_sha256;
+
+#[cfg(feature = "const")]
+use const_crypto::{bs58::decode_pubkey, sha2::Sha256};
+#[cfg(feature = "const")]
+pub use pinocchio::pubkey::Pubkey as __Pubkey;
```
illia-bobyr: *project-layout*

I think there is a slightly better way to organize crate re-exports.
So, I am going to make an unsolicited suggestion here.

The idea is to have a dedicated submodule to re-export any dependencies that macros use.
I call it `reexport`, and it looks something like this (in the `lib.rs`):

```rust
// Includes re-exports needed to implement macros
#[doc(hidden)]
pub mod reexport {
   pub use const_crypto;
   /* ... */
}
```

Which later allows you to write the macro body like this:
```rust
rust_macros! my_useful_macro {
    ($arg:ident) => {
        use $crate::reexport::{
            const_crypto::sha2::Sha256,
            /* ... */
        };

        let mut hasher = Sha256::new();
        /* ... */
    }
}
```

The advantage is that this structure scales to the whole package and in the specific macro body you can still see where symbols are coming from.

A few popular libraries follow this approach with slight variations:
* `tokio` uses this approach: https://github.com/tokio-rs/tokio/blob/aff24dfbebf79c698abf16db24930d478c099252/tokio/src/macros/mod.rs#L34-L36
(Though they call the submodule `support`).
* `zerocopy`: https://github.com/dorryspears/zerocopy/blob/727959e3037081d4e0abd51fe8f26f0eff60ebba/src/macro_util.rs#L512-L523
* `libstd` uses a variant of this approach: https://github.com/fabricedesre/rust/blob/master/src/libstd/lib.rs#L207-L211

sdk/pubkey/src/lib.rs
febo: Maybe  an `# Important` heading will be more visible?

sdk/pubkey/src/lib.rs
febo: Yeah, not including any "runtime" checks – perhaps we should have an `unchecked` version without any checks and one with the checks.

sdk/pubkey/src/lib.rs
febo: Using an expression `N + 3` was my first idea, but we can't use generic in `const` operations:

image

sdk/pubkey/src/lib.rs
buffalojoec: Sure, wfm!

sdk/pubkey/src/lib.rs
febo: Yeah, it is a bit inconsistent at the moment. But I think we should be moving to `panic!` since they should not be used off-chain.

sdk/pubkey/src/lib.rs
buffalojoec: I think it just depends on the purpose of this API. After all, you're still going to end up with a PDA mismatch, and if you abuse it, you'll exhaust all your CUs.

febo: The main purpose is for cases where you have the seeds for the derivation (e.g., ["my pda"]) and the bump stored in the account. So the likelihood of having an invalid seed length should be zero.

So maybe we don't need an "unchecked" version – is this what you are suggesting?

sdk/pubkey/src/lib.rs
febo: Here we can use `N`, might end up being the same but it is clearer to have `N` here.

sdk/pubkey/src/lib.rs
buffalojoec: Yes, exactly. Grown-ups only.

sdk/pubkey/src/lib.rs
illia-bobyr: Ah, sorry.
I've tested an intermediate version with `N`, but did not try one with `N + 3`.

Just out of curiosity, I've spent some time trying to understand the limitations better.

Seems like one still needs `typenum` and `generic_array` to do this in the current version of Rust: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ad724135fdcaad3b2f8b759033cdee35
Not sure if you want to add those libraries here.

The assignment gets a bit less readable:
```rust
    let data: GenericArray<&[u8], <U<N> as Add<U3>>::Output> = GenericArray::generate(|i| {
        if i < N {
            // SAFETY: `seeds` has `N` elements.
            unsafe { seeds.get_unchecked(i) }
        } else if i == N {
            bump
        } else if i == N + 1 {
            program_id
        } else if i == N + 2 {
            marker
        } else {
            unreachable!("i should never reach N + 3")
        }
    });
```

And you need a `where` clause that makes sure that 3 can be added to `N` and it is a valid array length:
```rust
pub fn derive_address<const N: usize>(
    seeds: &[&[u8]; N],
    bump: u8, program_id: &Pubkey,
) -> Pubkey
where
    Const<N>: ToUInt,
    <Const<N> as ToUInt>::Output: Add<U3>,
    <<Const<N> as ToUInt>::Output as Add<U3>>::Output: ArrayLength,
{
```

ilya-bobyr (dismissed):
Looks good :)

sdk/pubkey/src/lib.rs
febo: Hmmm, I feel like the `MaybeUninit` keeps the code clearer and without extra dependencies. 😅

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,142 @@
 #![no_std]
 
-pub use five8_const::decode_32_const;
-pub use pinocchio;
+#[cfg(feature = "const")]
+pub mod reexport {
+    pub use pinocchio::pubkey::Pubkey;
+}
```
illia-bobyr: I think it is nice to add a `#[doc(hidden)]` attribute.
To avoid this macro being visible in the generated docs.

```suggestion
#[cfg(feature = "const")]
#[doc(hidden)]
// Re-export dependencies used in macros.
pub mod reexport {
    pub use pinocchio::pubkey::Pubkey;
}
```

sdk/pubkey/src/lib.rs
illia-bobyr: It is your call :)
I think it is fine either way.

sdk/pubkey/src/lib.rs
febo: My bad, the intention was to have included the `#[doc(hidden)]` attribute.

sdk/pubkey/src/lib.rs
```diff
@@ -1,9 +1,144 @@
...
+/// the expected resulting address.
+///
+/// # Important
+///
+/// This function differs from [`pinocchio::pubkey::create_program_address`] in that it
+/// does not perform a validation to ensure that the derived address is a valid
+/// (off-curve) program derived address. It is intended for use in cases where the
+/// seeds, bump, and program id are known to be valid, and the caller wants to derive
+/// the address without incurring the cost of the `create_program_address` syscall.
+pub fn derive_address<const N: usize>(seeds: &[&[u8]; N], bump: u8, program_id: &Pubkey) -> Pubkey {
```
joncinque, line 39: Sorry for coming in late here, but is there a reason to require the `bump` parameter?

If the user is just hashing things together, there's a chance that they also simply used `create_program_address` without a bump, and it was valid, in which case, they can't use this function.

febo: Good point! So when you have a bump, you can simply include it with the seeds, right? Then the signature would be:
```rust
pub fn derive_address<const N: usize>(
  seeds: &[&[u8]; N],
  program_id: &Pubkey
) -> Pubkey;
```

joncinque: Yep exactly!

febo: Nice, I will change it.

rustopian:
> it is necessary to use the `create_program_address` syscall to validate the derivation.

It's only necessary if you are *not* subsequently doing any signing with the program address, right?

In that case, using `derive_address` is a choice to perform redundant validation early (whether for redundant safety or just for better errors), since an invalid address will fail validation upon `invoke_signed`. (In p-ata right now it's skipped entirely when bump is provided.)

sdk/pubkey/src/lib.rs
illia-bobyr: > Good point! So when you have a bump, you can simply include it with the seeds, right? Then the signature would be:
> 
> ```rust
> pub fn derive_address<const N: usize>(
>   seeds: &[&[u8]; N],
>   program_id: &Pubkey
> ) -> Pubkey;
> ```

Do we expect users to understand how to add bump to the list of seeds used in the key generation, in other places?

I wonder if this could also make it easier to forget to include the `bump` value.
And debugging an issue like this is annoying, as there are no other checks that point to the root cause.

I wonder if a more ergonomic API would be something like this:

```rust
pub fn derive_address<const N: usize>(
  seeds: &[&[u8]; N],
  bump: Option<u8>,
  program_id: &Pubkey,
) -> Pubkey;
```

The docs will then clarify that a `bump` of `None` will not cause the `bump` value to be included in the list of `seeds`.  And that it is a case for address generated via a `create_program_address()` call.

joncinque: Hm, typically users should know to include the bump, since you need to add it when using `invoke_signed` currently. With that said, people still routinely post questions the Solana Stack Exchange because they're PDA seeds are incorrect, so anything to help is probably useful.

All that to say, I'm not opposed to using an `Option<u8>`, it sounds like a nice middle-ground!

febo:
> > it is necessary to use the `create_program_address` syscall to validate the derivation.
> 
> It's only necessary if you are _not_ subsequently doing any signing with the program address, right?
> 
> In that case, using `derive_address` is a choice to perform redundant validation early (whether for redundant safety or just for better errors), since an invalid address will fail validation upon `invoke_signed`. (In p-ata right now it's skipped entirely when bump is provided.)

Yes, exactly – only necessary if you need to do the validation in your program. When used as a signer, the runtime will do the derivation as you mentioned.

PR #152, opened 2025-05-05
zero-copy slot hashes sysvar (with checked alternatives)

rustopian:
Adds SlotHashes sysvar support. Checked and unchecked zero-copy paths are both included. Intended to replace [121](https://github.com/anza-xyz/pinocchio/pull/121).

Binary search is used, but zerocopy `Iterator` and copy `fetch` / `fetch_into` / `fetch_into_unchecked` are also implemented for those with different SlotHashes access needs.

PR #193, opened 2025-07-08
chore : cleanup token crate

Nagaprasadvr:
1. rename from_bytes to from_bytes_unchecked
2. update docs


programs/token/src/state/mint.rs
```diff
@@ -74,16 +74,19 @@ impl Mint {
...
+        ))
     }
 
     /// Return a `Mint` from the given bytes.
     ///
     /// # Safety
     ///
-    /// The caller must ensure that `bytes` contains a valid representation of `Mint`.
+    /// The caller must ensure that `bytes` contains a valid representation of `Mint`. This method does not
+    /// perform a length validation.
```
illia-bobyr: Is there also an alignment requirement?
`Mint` is declared as `repr(C)` but it is not necessary needs to be byte aligned based on the current definition.
It currently probably is.

When working on the slot hashes sysvar @rustopian added this kind of check to make sure that a struct is indeed byte aligned:
```rust
// Fail compilation if `SlotHashEntry` is not byte-aligned.
const _: [(); 1] = [(); mem::align_of::<SlotHashEntry>()];
```
https://github.com/anza-xyz/pinocchio/pull/152/files#diff-edc4b9ebabc6ab34940d998f4a7d2f4049f78b7a9a054253507887f08289556eR55-R56

febo: > Is there also an alignment requirement?

Alignment requirement is `1`. So we could add the compile-time check for alignment, but probably not strictly necessary.

illia-bobyr: As we are adding a comment explaining the requirements on the input, we could mention the alignment requirements there as well.

Nagaprasadvr: will do

programs/token/src/state/token.rs
```diff
@@ -89,16 +89,19 @@ impl TokenAccount {
...
+        ))
     }
 
     /// Return a `TokenAccount` from the given bytes.
     ///
     /// # Safety
     ///
-    /// The caller must ensure that `bytes` contains a valid representation of `TokenAccount`.
+    /// The caller must ensure that `bytes` contains a valid representation of `TokenAccount` and its aligned to 1 byte. This method does not
+    /// perform a length validation.
```
illia-bobyr:
```suggestion
    /// The caller must ensure that `bytes` contains a valid representation of `TokenAccount`, and
    /// it is properly aligned to be interpreted as an instance of `TokenAccount`.
    /// At the moment `TokenAccount` has an alignment of 1 byte.
    /// This method does not perform a length validation.
```

We format code and comments not to exceed 100 characters in with, I think.

Also, the alignment requirement of 1 byte is somewhat confusing.
It essentially says "any alignment works".
It would be more accurate so say that this method is only safe when `bytes` is aligned correctly for the current alignment of `TokenAccount`.
Which is 1 for the moment.

PR #194, opened 2025-07-10
feat: publicly expose num_instructions in instruction introspection

deanmlittle:
Currently, there is no way to expose the number of instructions in an instruction introspection sysvar account. This feature publicly exposes the number of signatures, also using the same function inline in the checked version of `load_instruction_at`.


deanmlittle:
accidentally misnamed the branch, opened a [new PR here](https://github.com/anza-xyz/pinocchio/pull/195)

PR #196, opened 2025-07-10
pubkey: Optional bump parameter

febo:
### Problem

PR #192  added `derive_address` and `derive_address_const` helpers, both expect a `bump` value in addition to the seeds. While a `bump` value is generally used to ensure a valid PDA (off-curve) is generated, it is not a requirement – it is possible to have a valid PDA without a `bump` value.

### Solution

<del>This PR removes the individual `bump` parameter from `derive_address` and `derive_address_const`. When a `bump` value is used in the derivation, its value can be specified in the seeds array.</del>

This PR makes the individual `bump` parameter from `derive_address` and `derive_address_const` optional.

### DIFF
--- a/sdk/pubkey/src/lib.rs
+++ b/sdk/pubkey/src/lib.rs
@@ -80,7 +80,7 @@ pub fn derive_address<const N: usize>(
     {
         let mut pda = MaybeUninit::<[u8; 32]>::uninit();
 
-        // SAFETY: `data` has `N + 2` elements initialized.
+        // SAFETY: `data` has `i + 2` elements initialized.
         unsafe {
             sol_sha256(
                 data.as_ptr() as *const u8,


illia-bobyr:
Maybe still keep it as an `Option<u8>`?
Here are my arguments from: https://github.com/anza-xyz/pinocchio/pull/192#discussion_r2195931590

---

Do we expect users to understand how to add bump to the list of seeds used in the key generation in other places?

This could also make it easier to forget to include the `bump` value.
And debugging an issue like this is annoying, as there are no other checks that point to the root cause.

I wonder if a more ergonomic API would be something like this:

```rust
pub fn derive_address<const N: usize>(
  seeds: &[&[u8]; N],
  bump: Option<u8>,
  program_id: &Pubkey,
) -> Pubkey;
```

The docs will then clarify that a `bump` of `None` will not cause the `bump` value to be included in the list of `seeds`. And that it is a case for address generated via a `create_program_address()` call.

---

Jon also replied here: https://github.com/anza-xyz/pinocchio/pull/192#discussion_r2195943010

sdk/pubkey/src/lib.rs
```diff
@@ -36,46 +39,50 @@ use pinocchio::syscalls::sol_sha256;
...
-        assert!(
-            N <= MAX_SEEDS,
-            "number of seeds must be less than MAX_SEEDS"
-        );
+        assert!(N < MAX_SEEDS, "number of seeds must be less than MAX_SEEDS");
     }
 
     const UNINIT: MaybeUninit<&[u8]> = MaybeUninit::<&[u8]>::uninit();
-    let mut data = [UNINIT; MAX_SEEDS + 3];
+    let mut data = [UNINIT; MAX_SEEDS + 2];
```
illia-bobyr, line 52: Why reduce the maximum number of allowed seeds here?
Was it incorrect in the version with `bump`?

sdk/pubkey/src/lib.rs
```diff
@@ -36,46 +39,50 @@ use pinocchio::syscalls::sol_sha256;
...
     {
         let mut pda = MaybeUninit::<[u8; 32]>::uninit();
 
-        // SAFETY: `data` has `i + 3` elements initialized.
+        // SAFETY: `data` has `N + 2` elements initialized.
         unsafe {
             sol_sha256(
                 data.as_ptr() as *const u8,
-                (N + 3) as u64,
+                (N + 2) as u64,
```
illia-bobyr: This needs to use `i` now.

```suggestion
                (i + 1) as u64,
```

sdk/pubkey/src/lib.rs
febo: Yeah, the `bump` counts as a seed.

sdk/pubkey/src/lib.rs
```diff
@@ -14,68 +14,77 @@ use pinocchio::pubkey::{Pubkey, MAX_SEEDS, PDA_MARKER};
...
         unsafe {
             data.get_unchecked_mut(i).write(seeds.get_unchecked(i));
         }
         i += 1;
     }
 
-    let bump = [bump];
+    // TODO: replace this with `as_slice` when the MSRV is upgraded
+    // to `1.84.0+`.
+    let bump_seed = [bump.unwrap_or_default()];
```
illia-bobyr, line 66: *general discussion*

As bump, program id and the marker have known sizes, I started wondering what would a version that passes them as a single seed would look like.
The hasher concatenates all the bytes, so it should result in the same hash.

This may increase the CU cost, though.
I do not know how CUs are computed for the `sol_sha256` call.

Conditional `bump` makes things somewhat verbose, but the following worked in my test:

```rust

    const MAX_EXTRA_SEEDS_SIZE: usize = 1 + size_of::<Pubkey>() + size_of_val(PDA_MARKER);
    let extra_seeds: [u8; MAX_EXTRA_SEEDS_SIZE] = if let Some(bump) = bump {
        const PUBKEY_START: usize = 1;
        const PUBKEY_END: usize = PUBKEY_START + size_of::<Pubkey>();
        const PDA_MARKER_START: usize = PUBKEY_END;
        const PDA_MARKER_END: usize = PDA_MARKER_START + size_of_val(PDA_MARKER);
        array::from_fn(|i| {
            if i == 0 {
                bump
            } else if i >= PUBKEY_START && i < PUBKEY_END {
                program_id[i - PUBKEY_START]
            } else if i >= PDA_MARKER_START && i < PDA_MARKER_END {
                PDA_MARKER[i - PDA_MARKER_START]
            } else {
                unsafe { core::hint::unreachable_unchecked() }
            }
        })
    } else {
        const PUBKEY_START: usize = 0;
        const PUBKEY_END: usize = PUBKEY_START + size_of::<Pubkey>();
        const PDA_MARKER_START: usize = PUBKEY_END;
        const PDA_MARKER_END: usize = PDA_MARKER_START + size_of_val(PDA_MARKER);
        array::from_fn(|i| {
            if i >= PUBKEY_START && i < PUBKEY_END {
                program_id[i - PUBKEY_START]
            } else if i >= PDA_MARKER_START && i < PDA_MARKER_END {
                PDA_MARKER[i - PDA_MARKER_START]
            } else {
                0
            }
        })
    };

    if bump.is_some() {
        unsafe { data.get_unchecked_mut(i) }.write(&extra_seeds);
    } else {
        unsafe { data.get_unchecked_mut(i) }.write(&extra_seeds[..MAX_EXTRA_SEEDS_SIZE - 1]);
    }
```

On one hand the compiler should be able to remove bounds checks for all the `program_id[_]`/`PDA_MARKER[_]` accesses above, as the ranges are already checked, and the sizes are known at compilation time.
But maybe using `.get_unchecked(_)` makes sense as well, just to be sure.

sdk/pubkey/src/lib.rs
```diff
@@ -14,68 +14,77 @@ use pinocchio::pubkey::{Pubkey, MAX_SEEDS, PDA_MARKER};
...
+        data.get_unchecked_mut(i).write(program_id.as_ref());
+        data.get_unchecked_mut(i + 1).write(PDA_MARKER.as_ref());
     }
 
     #[cfg(target_os = "solana")]
     {
         let mut pda = MaybeUninit::<[u8; 32]>::uninit();
 
-        // SAFETY: `data` has `i + 3` elements initialized.
+        // SAFETY: `data` has `N + 2` elements initialized.
```
illia-bobyr: Should this say `i + 2`?

```suggestion
        // SAFETY: `data` has `i + 2` elements initialized.
```

ilya-bobyr (approved):
Looks good to me :)

ilya-bobyr (approved):

sdk/pubkey/src/lib.rs
joncinque: This is definitely worth trying out in a follow-up PR, thanks for the suggestion!

PR #197, opened 2025-07-14
`feat` : Add InitMultisig ixs

Nagaprasadvr:
1. Add InitMultisig
2. Add InitMultisig2 cpis


programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,75 @@
...
+///   1. `[]` Rent sysvar
+///   2. ..`2+N`. `[]` The signer accounts, must equal to N where `1 <= N <=
+///      11`.
+pub struct InitializeMultisig<'a> {
+    /// Multisig Account.
+    pub multisig: &'a AccountInfo,
+    /// Rent sysvar Account.
+    pub rent_sysvar: &'a AccountInfo,
+    /// Signer Accounts
+    pub multisig_signers: Vec<&'a AccountInfo>,
```
sonicfromnewyoke: i believe this vector can be replaced with the slice or even with the const generic array

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        // Instruction data layout:
+        // -  [0]: instruction discriminator (1 byte, u8)
+        // -  [1]: m (1 byte, u8)
+        let mut instruction_data = [UNINIT_BYTE; 2];
+
+        // Set discriminator as u8 at offset [0]
+        write_bytes(&mut instruction_data, &[2]);
+        // Set number of signers (m) at offset 1
+        write_bytes(&mut instruction_data[1..2], &[self.m]);
```
illia-bobyr: I think it is better to avoid uninitialiezd and unsafe, if possible.

```suggestion
        let instruction_data = [2, self.m];
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use crate::{write_bytes, UNINIT_BYTE};
+
+/// Initialize a new Multisig.
+///
+/// ### Accounts:
+///   0. `[writable]` The multisig account to initialize.
+///   1. `[]` Rent sysvar
+///   2. ..`2+N`. `[]` The signer accounts, must equal to N where `1 <= N <=
+///      11`.
```
illia-bobyr: The "must be equal to N" part is confusing to me.
What must be equal to N?
I suggest you rephrase it like this:
```suggestion
///   2. ..`2+N`. `[]` The N signer accounts, where N is between 1 and 11.
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,75 @@
...
+            accounts: account_metas.as_slice(),
+            data: unsafe { from_raw_parts(instruction_data.as_ptr() as _, 2) },
+        };
+
+        let mut account_infos = Vec::with_capacity(2 + self.multisig_signers.len());
+        account_infos.push(self.multisig);
+        account_infos.push(self.rent_sysvar);
+        account_infos.extend_from_slice(self.multisig_signers.as_slice());
+
+        slice_invoke_signed(&instruction, account_infos.as_slice(), signers)
```
illia-bobyr: *style*

`Vec::as_slice()` is the same as `&_`, but the latter is more common, I think.
```suggestion
        let instruction = Instruction {
            program_id: &crate::ID,
            accounts: &account_metas,
            data: unsafe { from_raw_parts(instruction_data.as_ptr() as _, 2) },
        };

        let mut account_infos = Vec::with_capacity(2 + self.multisig_signers.len());
        account_infos.push(self.multisig);
        account_infos.push(self.rent_sysvar);
        account_infos.extend_from_slice(&self.multisig_signers);

        slice_invoke_signed(&instruction, &account_infos, signers)
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        // Set discriminator as u8 at offset [0]
+        write_bytes(&mut instruction_data, &[2]);
+        // Set number of signers (m) at offset 1
+        write_bytes(&mut instruction_data[1..2], &[self.m]);
+
+        let instruction = Instruction {
+            program_id: &crate::ID,
+            accounts: account_metas.as_slice(),
+            data: unsafe { from_raw_parts(instruction_data.as_ptr() as _, 2) },
```
illia-bobyr: This `unsafe` should be unnecessary, as well as the conversion.

```suggestion
            data: &instruction_data,
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub m: u8,
+}
+
+impl InitializeMultisig<'_> {
+    #[inline(always)]
+    pub fn invoke(&self) -> ProgramResult {
+        self.invoke_signed(&[])
+    }
+
+    pub fn invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
```
illia-bobyr: Considering Pinocchio is trying to be as fast as possible, would it make sense to call this method `slice_invoke_signed()`?
This would match the method it actually uses from the `cpi` module.
And would leave a space for an `invoke_signed()` method that has a compile time specified number of accounts.

programs/token/src/instructions/initialize_multisig.rs
illia-bobyr: I wonder if there is an ergonomic to generalize over the notion of "either a slice with a size known at compile time or a slice with a runtime defined size".  Though, such a generalization would probably be out of scope for this PR.

But there is no reason to restrict the type to a `Vec` - a slice is strictly better.

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,75 @@
...
+        // Account metadata
+        let mut account_metas = Vec::with_capacity(2 + self.multisig_signers.len());
+        account_metas.push(AccountMeta::writable(self.multisig.key()));
+        account_metas.push(AccountMeta::readonly(self.rent_sysvar.key()));
+        account_metas.extend(
+            self.multisig_signers
+                .iter()
+                .map(|a| AccountMeta::readonly(a.key())),
```
illia-bobyr: *bikeshedding*

I always annoy everyone with this in code reviews.
You can destructure `self` to avoid some of the "syntactic noise".

`self` is only used as a container for arguments.
(Not even sure why put some arguments into `self` and some into the methods).
But in any case, as soon as `invoke_signed()` is called, you only care for the individual fields inside `self`, not about `self` as a whole.

Destructuring makes it more obvious and removes all the `self.` prefixes, that have no actual meaning here:

```suggestion
        let &Self {
            multisig,
            rent_sysvar,
            multisig_signers,
            m,
        } = self;

        // Account metadata
        let mut account_metas = Vec::with_capacity(2 + multisig_signers.len());
        account_metas.push(AccountMeta::writable(multisig.key()));
        account_metas.push(AccountMeta::readonly(rent_sysvar.key()));
        account_metas.extend(
            multisig_signers
                .iter()
                .map(|a| AccountMeta::readonly(a.key())),
```

For all the other types except `Self` you can also destructure directly in the function arguments.
But if you do it for a `self`, function can not be called as a method any more.

programs/token/src/instructions/initialize_multisig.rs
Nagaprasadvr: yeah i totally forgot we just need to do this

programs/token/src/instructions/initialize_multisig.rs
Nagaprasadvr: cool , should i also add invoke_signed method with compile time num of accounts constant

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,71 @@
+use pinocchio::{
+    account_info::AccountInfo,
+    cpi::slice_invoke_signed,
+    instruction::{AccountMeta, Instruction, Signer},
+    ProgramResult,
+};
+
+extern crate alloc;
+use alloc::vec::Vec;
```
sonicfromnewyoke: you can drop it, its not used in this file anymore

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    pub fn slice_invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
+        let &Self {
+            multisig,
+            rent_sysvar,
+            multisig_signers,
+            m,
+        } = self;
+        // Account metadata
+        let mut account_metas = Vec::with_capacity(1 + multisig_signers.len());
```
sonicfromnewyoke: take a look at how similar manipulation was done by @publicqi for the [memo-program](https://github.com/anza-xyz/pinocchio/blob/main/programs/memo/src/instructions/mod.rs#L43)

after replacing the vector you will be able to remove `vec` from imports

programs/token/src/instructions/initialize_multisig_2.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    pub fn slice_invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
+        let &Self {
+            multisig,
+            multisig_signers,
+            m,
+        } = self;
+
+        // Account metadata
+        let mut account_metas = Vec::with_capacity(1 + multisig_signers.len());
```
sonicfromnewyoke: the same recommendations as for `initialize_multisig`

programs/token/src/instructions/initialize_multisig.rs
illia-bobyr: It probably makes sense to try and see what it looks like.
It can save a few CUs.
One problem is that the `multisig_signers` size is still not known at compile time.
And it would then require you to use `slice_invoke_signed()`.

Overall, it is probably not that big of a deal.
I would imagine that initialization instruction is not the most frequently used one.

But it would be nice to figure out a patter of how these should be described in Pinocchio.
It seems for the underlying CPI functions @febo(?) was trying to optimize the case when the size is known at compile time.
There are similar optimizations for the account initializations.
And, I think, most programs do know these numbers at the compilation time.

Using a struct, like `InitializeMultisig`, for some arguments, makes it a bit harder to express these optimizations.

---

You can probably do it in a separate PR, as long as name this function `slice_invoke_signed()` in this PR.

programs/token/src/instructions/initialize_multisig.rs
illia-bobyr: All those `unsafe` and `MaybeUninit` make me sad.
I wonder if there is a common pattern, that can be extracted, such that most functions would not have to go through all these motions of allocation of uninitialized memory.
It is easier to make a mistake with these lower level tools.

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,71 @@
...
+        // Instruction data layout:
+        // -  [0]: instruction discriminator (1 byte, u8)
+        // -  [1]: m (1 byte, u8)
+        let instruction_data = &[2, m];
+
+        let instruction = Instruction {
+            program_id: &crate::ID,
+            accounts: &account_metas,
+            data: instruction_data,
+        };
```
illia-bobyr: *minor*

If you use the same name for the variable as the field name, you can use a short assignment syntx.
As this method is so small, it is probably fine rename `accounts_metas` to just `accounts`, and `instruction_data` to just `data.
And then you'll have:

```suggestion
        let instruction = Instruction {
            program_id: &crate::ID,
            accounts,
            data,
        };
```

Nagaprasadvr: yeah i would have done that , but i just tried to be uniform with other methods just followed the same conventions

illia-bobyr: It is your call.
I would approve a more idiomatic code, to be honest.
Unnecessary renaming is not great, even if it is repeated everywhere :)
Also, it does seem like there is an opportunity to have a macro or something the would reduce the repetition.

sonicfromnewyoke: @ilya-bobyr your point is valid, but maybe it will be better to make such a refactoring in the dedicated MR and apply to all the CPIs, wdyt?

programs/token/src/instructions/initialize_multisig.rs
sonicfromnewyoke: rust is unjustifiably secure even in the places where this is not needed.
In this case we have to pay with CUs for the rust's "paranoia"...

programs/token/src/instructions/initialize_multisig_2.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        // Instruction data layout:
+        // -  [0]: instruction discriminator (1 byte, u8)
+        // -  [1]: m (1 byte, u8)
+        let data = &[2, m];
+
+        let instruction = Instruction {
+            program_id: &crate::ID,
+            accounts: unsafe {
+                core::slice::from_raw_parts(account_metas.as_ptr() as _, num_accounts)
```
illia-bobyr: I think it is better to import `core::slice`.
It is better to keep the symbol names short, to a reasonable minimum.
Also, the cast is unnecessary.

```rust
use core::slice;
```
```suggestion
                slice::from_raw_parts(account_metas.as_ptr(), num_accounts)
```

programs/token/src/instructions/initialize_multisig_2.rs
```diff
@@ -0,0 +1,110 @@
...
+            // - Index 0 is always present
+            account_metas
+                .get_unchecked_mut(0)
+                .write(AccountMeta::writable(multisig.key()));
+        }
+
+        for i in 1..(1 + multisig_signers.len()) {
+            unsafe {
+                // SAFETY:
+                // - `i` in 1..(1 + multisig_signers.len()) is guaranteed less than MAX_ALLOWED_ACCOUNTS
```
illia-bobyr: > `i` in 1..(1 + multisig_signers.len()) is guaranteed less than MAX_ALLOWED_ACCOUNTS

I do not see a check for this.

programs/token/src/instructions/initialize_multisig_2.rs
Nagaprasadvr: rust complains if we dont cast it

programs/token/src/instructions/initialize_multisig_2.rs
Nagaprasadvr: added a check for MAX_SIGNERS

programs/token/src/instructions/initialize_multisig_2.rs
illia-bobyr: Ah, it is the `MaybeUninit`.
I should have checked, sorry.

ilya-bobyr (commented):
Looks good to me.
@febo, @joncinque, do anyone of you would like to approve?

programs/token/src/instructions/initialize_multisig_2.rs
```diff
@@ -0,0 +1,112 @@
...
+                acc_metas.get_unchecked_mut(i).write(AccountMeta::readonly(
+                    multisig_signers.get_unchecked(i - 1).key(),
+                ));
+            }
+        }
+
+        // Instruction data layout:
+        // -  [0]: instruction discriminator (1 byte, u8)
+        // -  [1]: m (1 byte, u8)
+        let data = &[2, m];
```
joncinque: The data here is incorrect and should be 19: https://github.com/solana-program/token/blob/17f59a8cc22e2fe0d1e756aa54e1a2583665c71e/program/src/instruction.rs#L552
```suggestion
        let data = &[19, m];
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+
+impl InitializeMultisig<'_> {
+    pub const MAX_ALLOWED_ACCOUNTS: usize = 1 + 1 + 11; // 1 multisig + 1 rent_sysvar + 11 MAX_SIGNERS
+
+    #[inline(always)]
+    pub fn invoke(&self) -> ProgramResult {
+        self.slice_invoke_signed(&[])
+    }
+
+    pub fn slice_invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
```
joncinque: This is more a question for @febo since I see this in the other instruction implementations, but why do we pass in `signers` at all? This instruction requires no signatures

joncinque (commented):
Just a couple of questions, then we can merge this in

programs/token/src/instructions/initialize_multisig_2.rs
Nagaprasadvr: ah made a huge mistake here , will fix it

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+///   0. `[writable]` The multisig account to initialize.
+///   1. `[]` Rent sysvar
+///   2. ..`2+N`. `[]` The N signer accounts, where N is between 1 and 11.
+pub struct InitializeMultisig<'a> {
+    /// Multisig Account.
+    pub multisig: &'a AccountInfo,
+    /// Rent sysvar Account.
+    pub rent_sysvar: &'a AccountInfo,
+    /// Signer Accounts
+    pub multisig_signers: &'a [&'a AccountInfo],
```
febo: Better to have a different lifetime for the array:
```suggestion
    pub multisig_signers: &'b [&'a AccountInfo],
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+    ProgramResult,
+};
+
+/// Initialize a new Multisig.
+///
+/// ### Accounts:
+///   0. `[writable]` The multisig account to initialize.
+///   1. `[]` Rent sysvar
+///   2. ..`2+N`. `[]` The N signer accounts, where N is between 1 and 11.
+pub struct InitializeMultisig<'a> {
```
febo: Another lifetime for the array of signers:
```suggestion
pub struct InitializeMultisig<'a, 'b> {
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Rent sysvar Account.
+    pub rent_sysvar: &'a AccountInfo,
+    /// Signer Accounts
+    pub multisig_signers: &'a [&'a AccountInfo],
+    /// The number of signers (M) required to validate this multisignature
+    /// account.
+    pub m: u8,
+}
+
+impl InitializeMultisig<'_> {
```
febo:
```suggestion
impl InitializeMultisig<'_, '_> {
```

programs/token/src/instructions/initialize_multisig.rs
febo: As Jon mentioned, there are no signers in this instruction so we only need `invoke`.

Nagaprasadvr: Cool will change accordingly

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub const MAX_ALLOWED_ACCOUNTS: usize = 1 + 1 + 11; // 1 multisig + 1 rent_sysvar + 11 MAX_SIGNERS
```
febo: I think the important value here is the number of signers, so that should be the constant value.
```suggestion
    /// Maximum number of multisignature signers.
    pub const MAX_SIGNERS: usize = 11;
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+        // MAX_SIGNERS = 11
+        if multisig_signers.len() > 11 {
+            return Err(ProgramError::InvalidArgument);
+        }
+
+        let num_accounts = 2 + multisig_signers.len();
+
+        // Account metadata
+        const UNINIT_META: MaybeUninit<AccountMeta> = MaybeUninit::<AccountMeta>::uninit();
+        let mut acc_metas = [UNINIT_META; Self::MAX_ALLOWED_ACCOUNTS];
```
febo:
```suggestion
        let mut acc_metas = [UNINIT_META; 2 + Self::MAX_SIGNERS];
```

febo:
```suggestion
        let mut acc_metas = [UNINIT_META; 2 + Self::MAX_SIGNERS];
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub fn slice_invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
+        let &Self {
+            multisig,
+            rent_sysvar,
+            multisig_signers,
+            m,
+        } = self;
+
+        // MAX_SIGNERS = 11
+        if multisig_signers.len() > 11 {
```
febo:
```suggestion
        if multisig_signers.len() > Self::MAX_SIGNERS {
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        let instruction = Instruction {
+            program_id: &crate::ID,
+            accounts: unsafe { slice::from_raw_parts(acc_metas.as_ptr() as _, num_accounts) },
+            data,
+        };
+
+        // Account info array
+        const UNINIT_INFO: MaybeUninit<&AccountInfo> = MaybeUninit::uninit();
+        let mut acc_infos = [UNINIT_INFO; Self::MAX_ALLOWED_ACCOUNTS];
```
febo:
```suggestion
        let mut acc_infos = [UNINIT_INFO; 2 + Self::MAX_SIGNERS];
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+        for i in 2..(2 + multisig_signers.len()) {
+            unsafe {
+                // SAFETY:
+                // - `i` in 2..(2 + multisig_signers.len()) is guaranteed less than MAX_ALLOWED_ACCOUNTS
+                // - `i - 2` < multisig_signers.len()
+                acc_metas.get_unchecked_mut(i).write(AccountMeta::readonly(
+                    multisig_signers.get_unchecked(i - 2).key(),
+                ));
+            }
+        }
```
febo: Can you try the following simplification?
```suggestion
        for (account_meta, signer) in acc_metas[2..].iter_mut().zip(multisig_signers.iter()) {
            account_meta.write(AccountMeta::readonly(signer.key()));
        }
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+        for i in 2..(2 + multisig_signers.len()) {
+            unsafe {
+                // SAFETY:
+                // - `i` in 2..(2 + multisig_signers.len()) is guaranteed less than MAX_ALLOWED_ACCOUNTS
+                // - `i - 2` < multisig_signers.len()
+                acc_infos
+                    .get_unchecked_mut(i)
+                    .write(multisig_signers.get_unchecked(i - 2));
+            }
+        }
```
febo: Same here.
```suggestion
        for (account_info, signer) in acc_infos[2..].iter_mut().zip(multisig_signers.iter()) {
            account_info.write(signer);
        }
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        slice_invoke_signed(
+            &instruction,
+            unsafe { slice::from_raw_parts(acc_infos.as_ptr() as _, num_accounts) },
+            signers,
+        )
```
febo: Better to use `invoke_with_bounds` since the maximum number of accounts is known so we can reduce the amount of stack used.
```suggestion
        invoke_with_bounds::<{ 2 + Self::MAX_SIGNERS }>(
            &instruction,
            unsafe { slice::from_raw_parts(acc_infos.as_ptr() as _, num_accounts) },
        )
```

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,120 @@
+use core::{mem::MaybeUninit, slice};
+
+use pinocchio::{
+    account_info::AccountInfo,
+    cpi::slice_invoke_signed,
```
febo:
```suggestion
    cpi::invoke_with_bounds,
```

febo (commented):
Left a few comments/nits. The same apply to `initialize_multisig_2`.

programs/token/src/instructions/initialize_multisig_2.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub multisig: &'a AccountInfo,
+    /// Signer Accounts
+    pub multisig_signers: &'a [&'a AccountInfo],
+    /// The number of signers (M) required to validate this multisignature
+    /// account.
+    pub m: u8,
+}
+
+impl InitializeMultisig2<'_> {
+    pub const MAX_ALLOWED_ACCOUNTS: usize = 1 + 11; // 1 multisig + 11 MAX_SIGNERS
```
febo: Should use the same constant as `initialize_multisig`.

Nagaprasadvr:
i have resolved all the comments @febo

programs/token/src/instructions/initialize_multisig.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct InitializeMultisig<'a, 'b>
+where
+    'a: 'b,
+{
+    /// Multisig Account.
+    pub multisig: &'a AccountInfo,
+    /// Rent sysvar Account.
+    pub rent_sysvar: &'a AccountInfo,
+    /// Signer Accounts
+    pub multisig_signers: &'b [&'a AccountInfo],
```
febo: On the token program, we use `signers`.
```suggestion
    pub signers: &'b [&'a AccountInfo],
```

programs/token/src/instructions/initialize_multisig_2.rs
```diff
@@ -0,0 +1,94 @@
...
+///   0. `[writable]` The multisig account to initialize.
+///   1. ..`1+N`. `[]` The N signer accounts, where N is between 1 and 11.
+pub struct InitializeMultisig2<'a, 'b>
+where
+    'a: 'b,
+{
+    /// Multisig Account.
+    pub multisig: &'a AccountInfo,
+    /// Signer Accounts
+    pub multisig_signers: &'b [&'a AccountInfo],
```
febo:
```suggestion
    pub signers: &'b [&'a AccountInfo],
```

febo (commented):
One final nit.

PR #198, opened 2025-07-15
fix: use `pinocchio` workspace dep in the `memo` program

sonicfromnewyoke:
### Problem
memo program uses pinned `pinocchio` dependency

### Solution
use workspace dependency instead of pinned one

cc @publicqi

### DIFF
--- a/programs/memo/Cargo.toml
+++ b/programs/memo/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "pinocchio-memo"
 description = "Pinocchio helpers to invoke Memo program instructions"
-version = "0.1.0"
+version = "0.1.1"
 edition = { workspace = true }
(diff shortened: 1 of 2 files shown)


programs/memo/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "pinocchio-memo"
 description = "Pinocchio helpers to invoke Memo program instructions"
-version = "0.1.0"
+version = "0.1.1"
```
febo: Not sure if we need the version bump, since the dependency is still the same version. In any case, the version bump is done by the "publish" workflow so we don't need to do them manually.

sonicfromnewyoke: reverted

febo (approved):

PR #200, opened 2025-07-15
Clarify the use of constant values

febo:
### Problem

There are a few places where "raw" values (magic numbers) are being used, which might not clearly describe their use.

### Solution

This PR improves the documentation for these values. Additionally, it updates the name of the function used to calculate a pointer offset to a field of a type (`offset` -> `field_at_offset`) to better represent its purpose – this is a function used with "raw" values.

This is an alternative to PR #191

### DIFF
--- a/scripts/setup/solana.dic
+++ b/scripts/setup/solana.dic
@@ -64,3 +64,4 @@ RPC
 ed25519
 performant
 syscall/S
+bitmask
--- a/sdk/pinocchio/src/instruction.rs
+++ b/sdk/pinocchio/src/instruction.rs
@@ -73,35 +73,35 @@ pub struct Account<'a> {
     _account_info: PhantomData<&'a AccountInfo>,
 }
 
-/// Return a pointer to a type `U` given a source pointer for `T` and the relative offset
-/// to `U` in units of `T`.
+/// Return a pointer to a type `U` given type `T` has a field of type `U` at the specified
+/// offset (in bytes) from the start of the `T` type.
 ///
 /// # Safety
 ///
-/// The caller must ensure that the `source` pointer is valid for `T` and that the offset is
-/// within the bounds of the memory region pointed to by `source`. The resulting pointer must
-/// at an aligned memory location that can be safely cast to `*const U`.
+/// The caller must ensure that the `ptr` is a valid pointer to a type `T`, and `ptr + offset`
+/// points to bytes that are properly aligned for `U` and represent a bit pattern that is a
+/// valid instance of `U`.
 ///
 /// If any of this requirements is not valid, this function leads to undefined behavior.
 #[inline(always)]
-const unsafe fn offset<T, U>(source: *const T, offset: usize) -> *const U {
+const unsafe fn field_at_offset<T, U>(ptr: *const T, offset: usize) -> *const U {
     // SAFETY: The caller ensures that the offset is valid for the type `T` and that
     // the resulting pointer is valid for type `U`.
-    unsafe { (source as *const u8).add(offset) as *const U }
+    unsafe { (ptr as *const u8).add(offset) as *const U }
 }
 
 impl<'a> From<&'a AccountInfo> for Account<'a> {
     fn from(account: &'a AccountInfo) -> Self {
         Account {
             // SAFETY: offset `8` is the `key` field in the `Account` struct.
-            key: unsafe { offset(account.raw, 8) },
+            key: unsafe { field_at_offset(account.raw, 8) },
             // SAFETY: offset `72` is the `lamports` field in the `Account` struct.
-            lamports: unsafe { offset(account.raw, 72) },
+            lamports: unsafe { field_at_offset(account.raw, 72) },
             data_len: account.data_len() as u64,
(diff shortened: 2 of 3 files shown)


febo:
@joncinque @ilya-bobyr Instead of adding constant, this PR improves the documentation of the "raw" (constant) values in the code. Not sure whether this is better than the changes on #191 – I slightly prefer this one since we don't need to come up with names to constants that might not clearly describe their use.

sdk/pinocchio/src/instruction.rs
```diff
@@ -73,19 +73,35 @@ pub struct Account<'a> {
...
 }
 
+/// Return a pointer to a type `U` given a source pointer for `T` and the relative offset
+/// to `U` in units of `T`.
+///
+/// # Safety
+///
+/// The caller must ensure that the `source` pointer is valid for `T` and that the offset is
+/// within the bounds of the memory region pointed to by `source`. The resulting pointer must
+/// at an aligned memory location that can be safely cast to `*const U`.
```
illia-bobyr:
```suggestion
/// within the bounds of the memory region pointed to by `source`. The resulting pointer must
/// point at an aligned memory location that can be safely cast to `*const U`.
```

sdk/pinocchio/src/instruction.rs
```diff
@@ -73,19 +73,35 @@ pub struct Account<'a> {
...
+///
+/// If any of this requirements is not valid, this function leads to undefined behavior.
 #[inline(always)]
-const fn offset<T, U>(ptr: *const T, offset: usize) -> *const U {
-    unsafe { (ptr as *const u8).add(offset) as *const U }
+const unsafe fn offset<T, U>(source: *const T, offset: usize) -> *const U {
```
illia-bobyr: Unrelated to the documentation you are adding (which is great by the way), I think this function, as it is, is the ultimate foot gun :)
It removes pretty much all the constraints on the source and the destination types, and performs no checks as to their relationship.

While I can see why you want something like that, I wonder if it is possible to be more precise as to what it should do.
And as a result, potentially make it a bit safer.

It seems like a more specific function can replace the `offset()` uses in this module.
Essentially, a function that is like `offset()`, but with `T` being specifically `*const u8`.
And this would allow the name and the semantics to be more precise as well.
It would be called something like `at_offset()`, or, if we want to be more verbose, `val_at_offset()`.  Or something like that.
For the higher level abstractions, "slice" is a well known abstraction.
But I do not know if there is a common terminology for pointers.

If we were working with slices, it could be called `as_slice_transmute()`.
Maybe for pointers it should be called `at_offset_transmute()`?
[`transmute()` docs](https://doc.rust-lang.org/std/mem/fn.transmute.html) for reference.
(Even for `transmute()` the compiler is [trying to provide some safety](https://doc.rust-lang.org/std/mem/trait.TransmuteFrom.html).)

```rust
/// Return a pointer to a type `T` give there is an instance of `T` at the specified offset.
///
/// # Safety
///
/// The caller must ensure that the `ptr` is valid, and `ptr + offset` points to bytes that are
/// properly aligned for `T` and represent a bit pattern that is a valid instance of `T`.
///
/// If any of this requirements is not valid, this function leads to undefined behavior.
#[inline(always)]
const unsafe fn at_offset_transmute<T>(ptr: *const u8, offset: usize) -> *const T {
```

ilya-bobyr (commented):
I like all the comments.
I think they make the code much easier to follow.
Especially for someone who does not have the whole thing already in their head.

illia-bobyr:
> @joncinque @ilya-bobyr Instead of adding constant, this PR improves the documentation of the "raw" (constant) values in the code. Not sure whether this is better than the changes on #191 – I slightly prefer this one since we don't need to come up with names to constants that might not clearly describe their use.

I do not have a very strong opinion one way or another.
But I also think that this version seems better, even despite it having more raw constants, compared to #191.

sdk/pinocchio/src/instruction.rs
febo: I like your suggestion – went with a slightly modified variation:
```rust
const unsafe fn field_at_offset<T, U>(ptr: *const T, offset: usize) -> *const U;
```
The function is essentially extracting a pointer to a type `U` that is a field of a type `T` at the specified offset.

illia-bobyr: Now this commit contains changes beside just comments on constants :P

scripts/setup/solana.dic
```diff
@@ -63,4 +63,5 @@ callee
 RPC
 ed25519
 performant
-syscall/S
\ No newline at end of file
+syscall/S
+bitmask
```
illia-bobyr, line 67: *minor*

Seems like this file does not have a new line at the end.
Here is one set of arguments as to why this might matter: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

joncinque (approved):
Looks good to me!

sdk/pinocchio/src/instruction.rs
febo: True – updated the PR name and description to better reflect the changes in it.

PR #201, opened 2025-07-16
perf: more aggressive inlining in `memo` and `ata` programs

sonicfromnewyoke:
### Problem
**system** and **token** programs inline the `invoke_signed` call for reaching better performance when **memo** and **ata** programs dont

### Solution
more aggressive inlining,  which leads to saving up to `~100` CU per call

### Potential Troubles

```
rustc-LLVM ERROR: Branch target out of insn range
```

should be fixed on the user's code side, that is responsible for using **memo** or **ata** client

### DIFF
--- a/programs/associated-token-account/Cargo.toml
+++ b/programs/associated-token-account/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "pinocchio-associated-token-account"
 description = "Pinocchio helpers to invoke Associated Token Account program instructions"
-version = "0.1.2"
+version = "0.1.1"
 edition = { workspace = true }
 license = { workspace = true }
 readme = "./README.md"
--- a/programs/memo/Cargo.toml
+++ b/programs/memo/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "pinocchio-memo"
 description = "Pinocchio helpers to invoke Memo program instructions"
-version = "0.1.1"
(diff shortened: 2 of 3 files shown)


programs/associated-token-account/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "pinocchio-associated-token-account"
 description = "Pinocchio helpers to invoke Associated Token Account program instructions"
-version = "0.1.1"
```
febo: Version bumps are done by the "publish" workflow, so we can revert this change.

programs/memo/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "pinocchio-memo"
 description = "Pinocchio helpers to invoke Memo program instructions"
-version = "0.1.0"
+version = "0.1.1"
```
febo: Same here, version bumps are done by the "publish" workflow so we can revert this change.

programs/associated-token-account/Cargo.toml
sonicfromnewyoke: my bad, reverted

programs/memo/Cargo.toml
sonicfromnewyoke: my bad, reverted

febo (approved):
Great – thanks!

PR #202, opened 2025-07-17
Add `resize_unchecked` to `AccountInfo`

febo:
### Problem

The current behaviour of `AccountInfo::realloc` allowing to skip zero-initializing the memory is deprecated. Programs should not rely that newly-allocated memory is zeroed. A new `resize` method was introduced to zero-initialize the data, although programs might define alternative initialization that do not require the data to be zeroed.

### Solution

This PR modifies the `realloc` implementation to use `resize`, thus ignoring the `zero_init` parameter and always zero-initializing the data. Additionally, it adds a `resize_unchecked`, which can be used when the program performs the account data borrow check and initialization of the allocated memory.

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -456,77 +456,29 @@ impl AccountInfo {
         Ok(())
     }
 
-    /// Realloc the account's data and optionally zero-initialize the new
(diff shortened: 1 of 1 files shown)


febo:
This will be refactored to provide a `MaybeUninit<&[u8]>` return.

PR #203, opened 2025-07-17
Ignore `zero_init` parameter

febo:
### Problem

The current behaviour of `AccountInfo::realloc` allowing to skip zero-initializing the memory is deprecated, although `realloc` still accepts a `zero_init` parameter. A new `resize` method was introduced in #174.

### Solution

This PR modifies the `realloc` implementation to use `resize`, thus ignoring the `zero_init` parameter and always zero-initializing the data.


joncinque (approved):
Looks great!

PR #207, opened 2025-07-18
pubkey: Use qualified name for `Pubkey`

febo:
### Problem

PR #192 added a "hidden" `reexport` module to be used by the macros, but macros are now importing the `Pubkey` type. This creates problems when the callsite for the macro already imports the same type.

### Solution

Use the qualified name for `Pubkey` to avoid importing the type.


joncinque (approved):
Makes sense!

PR #209, opened 2025-07-19
Fix duplicated account parsing

febo:
### Problem

PR #176 refactored the entrypoint and inlined the processing of accounts, but the parsing of duplicated accounts is wrongly storing the address to an `AccountInfo` instead of an `Account`.

### Solution

De-reference the `AccountInfo` to store the address of its `Account`. As a side-effect of the fix, CU consumption improved:

| Name | CUs | Delta from current |
|------|------|-------|
| Account (0) | 9 | -- |
| Account (1) | 13 | -- |
| Account (2) | 13 | -8 |
| Account (3) | 33 | -1 |
| Account (4) | 41 | -1 |
| Account (5) | 48 | -1 |
| Account (6) | 63 | -1 |
| Account (7) | 67 | -1 |
| Account (8) | 74 | -1 |
| Account (16) | 139 | -1 |
| Account (32) | 257 | -1 |
| Account (64) | 500 | -1 |

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -225,9 +225,7 @@ macro_rules! process_n_accounts {
         $input = $input.add(size_of::<u64>());
 
         if (*account).borrow_state != NON_DUP_MARKER {
-            $accounts.write(AccountInfo {
-                raw: $accounts_slice.add((*account).borrow_state as usize) as *mut Account,
-            });
+            clone_account_info($accounts, $accounts_slice, (*account).borrow_state);
         } else {
             $accounts.write(AccountInfo { raw: account });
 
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -258,6 +256,29 @@ macro_rules! process_accounts {
...
+///
+/// # Safety
+///
+/// The caller must ensure that:
+///   - `accounts` pointer must point to an array of `AccountInfo`s where
+///      the new `AccountInfo` will be written.
+///   - `accounts_slice` pointer must point to a slice of `AccountInfo`s
+///     already initialized.
+///   - `index` is a valid index in the `accounts_slice`.
+#[cold]
```
febo, line 274: This was needed to stop the compiler trying to optimize the parsing of duplicated accounts and increasing the overall CU consumption.

illia-bobyr: I think it is better to add this explanation as a comment next to the attribute.

ilya-bobyr (approved):

sdk/pinocchio/src/entrypoint/mod.rs
febo: Good call!

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -766,6 +857,30 @@ mod tests {
...
+
+        // Duplicated accounts should have the same `data_len` as the
+        // last unique account.
+        for account in accounts[unique..].iter() {
+            let account_info = unsafe { account.assume_init_ref() };
+            let duplicated = unsafe { accounts[unique - 1].assume_init_ref() };
+
+            assert_eq!(account_info.raw, duplicated.raw);
+            assert_eq!(account_info.data_len(), duplicated.data_len());
+        }
```
joncinque: Would it be worth adding a check that borrowing `unique` increases the borrow count on the duplicates? Or that changing the data on one of the duplicates reflects in `unique`?

joncinque (dismissed):
Ah too bad this snuck in, but looks good to me! I feel silly because I *almost* asked for a test with duplicates in the previous PR.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Oh, good idea. I think it is somewhat covered by the raw pointer check, but it is a nice way to show they are the same.

febo:
> Ah too bad this snuck in, but looks good to me! I feel silly because I _almost_ asked for a test with duplicates in the previous PR.

I feel the same. There was not a good reason not to include one.

PR #214, opened 2025-07-29
docs: Format `entrypoint` documentation

ilya-bobyr:
Format doc-comments and comment to be 100 characters wide.

Use quotes when referencing feature names.
`` `"no_std"` feature ``, rather than `` "`no_std`" feature ``, or `` `no_std` feature ``.
This seems to be the common convention.

Also added a few automatic links in the doc comments, where previously it was just quoted name of a type.

---

Not sure what do you guys think about this kind of changes.
Some people call it "churn" and do not accept them.

I personally think that it would be much better if these things would be automatically enforced, similar to how `fmt` deals with many other formatting details.
But it does not seem like it can do it.

If merged, I would next add a `.git-blame-ignore-revs` and add this commit SHA in there.
GitHub will pick it up automatically:
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

Git in general will require a bit of manual configuration, but it seems to be a pretty common approach:
https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-filefile

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -43,16 +43,17 @@ const STATIC_ACCOUNT_DATA: usize = size_of::<Account>() + MAX_PERMITTED_DATA_INC
 
 /// Declare the program entrypoint and set up global handlers.
 ///
-/// The main difference from the standard (SDK) [`entrypoint`](https://docs.rs/solana-program-entrypoint/latest/solana_program_entrypoint/macro.entrypoint.html)
-/// macro is that this macro represents an entrypoint that does not perform allocations or copies
-/// when reading the input buffer.
+/// The main difference from the standard (SDK) [`entrypoint`] macro is that this macro represents
+/// an entrypoint that does not perform allocations or copies when reading the input buffer.
 ///
-/// This macro emits the common boilerplate necessary to begin program execution, calling a
-/// provided function to process the program instruction supplied by the runtime, and reporting
-/// its result to the runtime.
+/// [`entrypoint`]: https://docs.rs/solana-program-entrypoint/latest/solana_program_entrypoint/macro.entrypoint.html
 ///
-/// It also sets up a [global allocator] and [panic handler], using the [`crate::default_allocator!`]
-/// and [`crate::default_panic_handler!`] macros.
+/// This macro emits the common boilerplate necessary to begin program execution, calling a provided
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -109,15 +110,17 @@ const STATIC_ACCOUNT_DATA: usize = size_of::<Account>() + MAX_PERMITTED_DATA_INC
...
-/// When the program and all its dependencies are `no_std`, it is necessary to set a
-/// `#[panic_handler]` to handle panics. This is done by the [`crate::nostd_panic_handler`](https://docs.rs/pinocchio/latest/pinocchio/macro.nostd_panic_handler.html)
-/// macro. In this case, it is not possible to use the `entrypoint`
-/// macro. Use the [`crate::program_entrypoint!`] macro instead and set up the allocator and panic
-/// handler manually.
+/// linker or not. The `entrypoint` macro will set up a default panic "hook", that works with the
+/// `#[panic_handler]` set by the `std`. Therefore, this macro should be used when the program or
+/// any of its dependencies are dependent on the `std` library.
+///
+/// When the program and all its dependencies are `"no_std"`, it is necessary to set a
```
joncinque: `no_std` isn't a crate feature, but rather an attribute to be used as `#![no_std]`, like `#[cfg(test)]`, so I would keep it as it was. I don't see it referred to as `"no_std"` anywhere.

The rust embedded book just uses `no_std`: https://docs.rust-embedded.org/book/intro/no-std.html

joncinque (commented):
Looks good to me! Just a couple of small things.

We should be able to enforce this with https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#comment_width, no? Pinocchio just doesn't have a rustfmt.toml like [Agave](https://github.com/anza-xyz/agave/blob/master/rustfmt.toml) or the [SDK](https://github.com/anza-xyz/solana-sdk/blob/master/rustfmt.toml)

As for the blame skip, I'm not too fussed about it personally.

sdk/pinocchio/src/entrypoint/mod.rs
illia-bobyr: My bad - thank you for checking.
I thought `no_std` is a feature as well.

joncinque (approved):
Looks great!

PR #219, opened 2025-08-09
token-2022: Remove unnecessary `invoke_signed`

febo:
### Problem

There are a few instructions that provide an `invoke_signed` method although the instruction does not require any signer.

### Solution

Remove the unnecessary `invoke_signed`.


rustopian (approved):
Same here, looks good.

PR #225, opened 2025-08-20
`feat` : scaffold token-2022 extensions

Nagaprasadvr:
1. setup extensions module
2. add Extensions enum and helpers to get and parse extension data from account data bytes


Taylor123:
Would love to see this get published into the token2022 crate

febo:
Thanks for the initial work, this has been implemented in #349.

PR #228, opened 2025-08-21
Feat: Add debug/copy derives and enable missing debug/copy lint

stegaBOB:
Types that can implement copy generally should: https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy

For debug, the rust API guidelines suggest debug should always be implemented in public types: https://rust-lang.github.io/api-guidelines/debuggability.html#all-public-types-implement-debug-c-debug


sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -668,6 +668,7 @@ mod alloc {
 
 #[cfg(not(feature = "std"))]
 /// An allocator that does not allocate memory.
+#[derive(Debug, Copy, Clone)]
```
febo: Let's keep them with a consistent order:
```suggestion
#[derive(Clone, Copy, Debug)]
```

sdk/pinocchio/src/instruction.rs
```diff
@@ -37,7 +37,7 @@ pub struct ProcessedSiblingInstruction {
 /// This struct contains the same information as an [`AccountInfo`], but has
 /// the memory layout as expected by `sol_invoke_signed_c` syscall.
 #[repr(C)]
-#[derive(Clone)]
+#[derive(Clone, Debug, Copy)]
```
febo:
```suggestion
#[derive(Clone, Copy, Debug)]
```

sdk/pinocchio/src/sysvars/clock.rs
```diff
@@ -33,7 +33,7 @@ pub type UnixTimestamp = i64;
 ///
 /// All members of `Clock` start from 0 upon network boot.
 #[repr(C)]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default, Debug)]
```
febo:
```suggestion
#[derive(Clone, Copy, Debug, Default)]
```

sdk/pinocchio/src/sysvars/fees.rs
```diff
@@ -4,7 +4,7 @@ use super::{clock::DEFAULT_MS_PER_SLOT, Sysvar};
 use crate::impl_sysvar_get;
 
 /// Fee calculator for processing transactions
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Default)]
```
febo: Is there a motivation for adding `Default` here?

sdk/pinocchio/src/sysvars/fees.rs
```diff
@@ -74,7 +74,7 @@ impl FeeRateGovernor {
 }
 
 /// Fees sysvar
-#[derive(Debug)]
+#[derive(Debug, Default, Copy, Clone)]
```
febo: Same here, is there a motivation for adding `Default` here?
```suggestion
#[derive(Clone, Copy, Debug, Default)]
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -13,6 +13,7 @@ pub const INSTRUCTIONS_ID: Pubkey = [
     0xc1, 0x24, 0xc6, 0x8f, 0x21, 0x56, 0x75, 0xa5, 0xdb, 0xba, 0xcb, 0x5f, 0x08, 0x00, 0x00, 0x00,
 ];
 
+#[derive(Debug, Copy, Clone)]
```
febo:
```suggestion
#[derive(Clone, Copy, Debug]
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -125,7 +126,7 @@ impl<'a> TryFrom<&'a AccountInfo> for Instructions<Ref<'a, [u8]>> {
 }
 
 #[repr(C)]
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq, Debug)]
```
febo:
```suggestion
#[derive(Clone, Debug, Eq, PartialEq)]
```

sdk/pinocchio/src/sysvars/instructions.rs
```diff
@@ -212,7 +213,7 @@ const IS_SIGNER: u8 = 0b00000001;
 const IS_WRITABLE: u8 = 0b00000010;
 
 #[repr(C)]
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq, Debug, Copy)]
```
febo:
```suggestion
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
```

sdk/pinocchio/src/sysvars/rent.rs
```diff
@@ -52,7 +52,7 @@ pub const ACCOUNT_STORAGE_OVERHEAD: u64 = 128;
 
 /// Rent sysvar data
 #[repr(C)]
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default, Copy)]
```
febo: We specifically removed `Default` implementation of sysvars [here](https://github.com/anza-xyz/pinocchio/pull/189).
```suggestion
#[derive(Clone, Copy, Debug)]
```

stegaBOB: Ah I think that must have been a mistake when I rebased my PR

sdk/pinocchio/src/sysvars/fees.rs
stegaBOB: I think that must have been a mistake. Not sure why I added that
```suggestion
#[derive(Debug, Clone, Copy)]
```

sdk/pinocchio/src/sysvars/fees.rs
stegaBOB:
```suggestion
#[derive(Copy, Clone, Debug)]
```
Not sure why. Was not intentional

sdk/pinocchio/src/sysvars/clock.rs
```diff
@@ -33,7 +33,7 @@ pub type UnixTimestamp = i64;
 ///
 /// All members of `Clock` start from 0 upon network boot.
 #[repr(C)]
-#[derive(Copy, Clone)]
+#[derive(Clone, Copy, Debug, Default)]
```
stegaBOB:
```suggestion
#[derive(Clone, Copy, Debug)]
```

febo (dismissed):
Great, thanks!

LStan:
Is it a good idea to derive `Copy` in code that tends to be zero-copy? Especially for sysvars (and perhaps `Account` in `instruction.rs`).
Imagine I have the following code:
```
let clock = Clock::get()?;
some_fn(clock, ...);
```
Here, a move happens, so I don't really want to bother to use reference.
However, after some time, I add more code after `some_fn` that uses `clock` variable.
Without the `Copy` trait, the compiler will tell me that I’m using a moved variable, and I'll consider changing the parameter of `some_fn` from `Clock` to `&Clock`. But with the `Copy` trait, I might not realize that a copy is happening.

febo:
> Is it a good idea to derive `Copy` in code that tends to be zero-copy? Especially for sysvars (and perhaps `Account` in `instruction.rs`). Imagine I have the following code:
> 
> ```
> let clock = Clock::get()?;
> some_fn(clock, ...);
> ```
> 
> Here, a move happens, so I don't really want to bother to use reference. However, after some time, I add more code after `some_fn` that uses `clock` variable. Without the `Copy` trait, the compiler will tell me that I’m using a moved variable, and I'll consider changing the parameter of `some_fn` from `Clock` to `&Clock`. But with the `Copy` trait, I might not realize that a copy is happening.

This is a good point. Perhaps would be prudent to not implement `Copy` in our case since we should be minimizing copies in a program. Agree with you that `Copy` makes "moves" silent in the sense that every assignment is a copy. What do you think @stegaBOB?

stegaBOB:
I still think if a type can be copy it probably should (which is what the std docs say as well). I'd imagine the compiler would probably optimize any unnecessary copies out anyway, although that would perhaps be worth benchmarking. If you really don't want the type to be copy, you could always make a newtype wrapper around the struct and not impl copy on that. 

```rs
#[derive(Clone)]
#[repr(transparent)]
pub struct RentNoCopy(Rent);
```

Having it copy allows more choices. In star frame, we plop sysvars in a Cell, which is only possible since they're copy: https://github.com/staratlasmeta/star_frame/blob/eb1d9f8b8a3a10e88bfc08bd459433881cce56b6/star_frame/src/context.rs#L17

febo:
> I still think if a type can be copy it probably should (which is what the std docs say as well). I'd imagine the compiler would probably optimize any unnecessary copies out anyway, although that would perhaps be worth benchmarking. If you really don't want the type to be copy, you could always make a newtype wrapper around the struct and not impl copy on that.
> 
> ```rust
> #[derive(Clone)]
> #[repr(transparent)]
> pub struct RentNoCopy(Rent);
> ```
> 
> Having it copy allows more choices. In star frame, we plop sysvars in a Cell, which is only possible since they're copy: https://github.com/staratlasmeta/star_frame/blob/eb1d9f8b8a3a10e88bfc08bd459433881cce56b6/star_frame/src/context.rs#L17

Yeah, that is also a good point. 😊 Given that it is possible to go from `Copy` to non-`Copy`, but not possible the other way around, let's keep it.

LStan:
The advice to implement `Copy` makes sense for `std` and non-constrained environment. But Pinocchio is used in a very constrained environment. The code in my example will not be optimized by the compiler, it will create a copy and use additional CUs without you realizing it.
Also, sysvars in the original sdk don't implement `Copy`, only `CloneZeroed`.
Maybe it is better to put `Copy` derivation into a feature?

febo:
> Maybe it is better to put `Copy` derivation into a feature?

This is probable the best compromise.

PR #229, opened 2025-08-21
Feat: Add try_maps on AccountInfo Ref/RefMut

stegaBOB:


Aursen:
Did you benchmark this kind of changes?

sonicfromnewyoke (commented):
@Aursen i expect small performance degradation

stegaBOB:
I'll benchmark it, but I doubt this would have any performance impacts. The compiler should optimize the error branch out.

stegaBOB:
@Aursen @sonicfromnewyoke https://rust.godbolt.org/z/9YeWsr1oP map with try_map infallible under the hood compiles down to the same assembly, so this should have 0 performance impact. This should be trivial for the compiler to optimize. One of the reasons Infallible exists :)

sonicfromnewyoke:
> @Aursen @sonicfromnewyoke https://rust.godbolt.org/z/9YeWsr1oP map with try_map infallible under the hood compiles down to the same assembly, so this should have 0 performance impact. This should be trivial for the compiler to optimize. One of the reasons Infallible exists :)

@stegaBOB it works even better than i expected, thanks

the last question: what do you think about inlining? Will it be better to use `#[inline(always)]` instead of putting the decision on the compiler??

Aursen:
> @stegaBOB it works even better than i expected, thanks
> 
> the last question: what do you think about inlining? Will it be better to use `#[inline(always)]` instead of putting the decision on the compiler??

I think for this kind of choice, @febo just benchmarks

stegaBOB:
> the last question: what do you think about inlining? Will it be better to use `#[inline(always)]` instead of putting the decision on the compiler??

I think `#[inline(always)]` should only really be done if you have serious benchmarks to justify it (or if the function is just empty). The compiler will probably make the right choice

febo:
> > @stegaBOB it works even better than i expected, thanks
> > the last question: what do you think about inlining? Will it be better to use `#[inline(always)]` instead of putting the decision on the compiler??
> 
> I think for this kind of choice, @febo just benchmarks

In this case `#[inline]` seems more appropriate since we don't know what the mapping function will be – it might simple, but it might not.

febo (dismissed):
👌

sdk/pinocchio/src/account_info.rs
```diff
@@ -715,16 +725,26 @@ impl<'a, T: ?Sized> RefMut<'a, T> {
...
+
+    /// Maps a mutable reference to a new type in a fallible map, returning an error if the map fails.
+    #[inline]
+    pub fn try_map<U: ?Sized, F, E>(orig: RefMut<'a, T>, f: F) -> Result<RefMut<'a, U>, E>
+    where
+        F: FnOnce(&mut T) -> Result<&mut U, E>,
     {
         // Avoid decrementing the borrow flag on Drop.
         let mut orig = ManuallyDrop::new(orig);
+        let value = f(&mut *orig)?;
```
febo: On a second thought, can we add a test for the `try_map`? Mainly to test the behaviour of the drop in case of failure. I wonder whether we need to invert these lines or not, i.e., first call the function before wrapping the `orig` with a `ManuallyDrop`.

stegaBOB: Will do! Good catch; I think you're right. We want it to get dropped on failure

PR #230, opened 2025-08-21
Add resize_unchecked method to account info

stegaBOB:
We retain control of the `RefMut` and need to be able to resize the account without going through the borrow check. This enables that, and is unsafe to indicate that the borrow check isn't being done.

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -495,6 +495,25 @@ impl AccountInfo {
         // Check wheather the account data is already borrowed.
         self.can_borrow_mut_data()?;
 
+        // SAFETY:
+        // We are checking if the account data is already borrowed, so we are safe to call
+        unsafe {
+            self.resize_unchecked(new_len)?;
+        }
+
+        Ok(())
+    }
+
+    /// Resizes the account's data and udpates the resize delta without checking if the account is already borrowed.
+    ///
+    /// The account data can be increased by up to [`MAX_PERMITTED_DATA_INCREASE`] bytes
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe because it does not check if the account data is already
+    /// borrowed.
+    #[inline(always)]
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/account_info.rs
```diff
@@ -495,6 +495,25 @@ impl AccountInfo {
...
+    }
+
+    /// Resizes the account's data and udpates the resize delta without checking if the account is already borrowed.
+    ///
+    /// The account data can be increased by up to [`MAX_PERMITTED_DATA_INCREASE`] bytes
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe because it does not check if the account data is already
+    /// borrowed.
```
febo:
```suggestion
    /// This method is unsafe because it does not check if the account data is already
    /// borrowed. The caller must guarantee that there are no active borrows to the account
    /// data.
```

sdk/pinocchio/src/account_info.rs
```diff
@@ -495,6 +495,25 @@ impl AccountInfo {
...
+        // SAFETY:
+        // We are checking if the account data is already borrowed, so we are safe to call
+        unsafe {
+            self.resize_unchecked(new_len)?;
+        }
+
+        Ok(())
+    }
+
+    /// Resizes the account's data and udpates the resize delta without checking if the account is already borrowed.
```
febo:
```suggestion
    /// Resize (either truncating or zero extending) the account's data.
```

sdk/pinocchio/src/account_info.rs
```diff
@@ -495,6 +495,25 @@ impl AccountInfo {
         // Check wheather the account data is already borrowed.
         self.can_borrow_mut_data()?;
 
+        // SAFETY:
+        // We are checking if the account data is already borrowed, so we are safe to call
+        unsafe {
+            self.resize_unchecked(new_len)?;
+        }
+
+        Ok(())
```
febo: nit:
```suggestion
        unsafe { self.resize_unchecked(new_len) }
```

febo (approved):

PR #231, opened 2025-08-24
Feat: implement Error for ProgramError

stegaBOB:
Stopgap solution until ProgramError is replaced with sdk types


publicqi:
ref: #80

PR #232, opened 2025-08-24
Feat: make AccountInfo::data_ptr public

stegaBOB:
Useful for unsafe operations that need to access the data pointer in a stable way without going through a reborrow

### DIFF
--- a/sdk/pinocchio/src/account_info.rs
+++ b/sdk/pinocchio/src/account_info.rs
@@ -930,8 +930,8 @@ mod tests {
     #[test]
     fn test_borrow_data() {
         // 8-bytes aligned account data.
-        let mut data = [0u64; size_of::<Account>() / size_of::<u64>()];
-        // Set the borrow state.
+        let mut data = [0u64; size_of::<Account>() / size_of::<u64>() + 1]; // extra byte at end for account data
+                                                                            // Set the borrow state.
         data[0] = NOT_BORROWED as u64;
         let account_info = AccountInfo {
             raw: data.as_mut_ptr() as *mut Account,
@@ -942,6 +942,13 @@ mod tests {
         assert!(account_info.can_borrow_mut_data().is_ok());
         assert!(account_info.can_borrow_lamports().is_ok());
         assert!(account_info.can_borrow_mut_lamports().is_ok());
+
+        // It should be sound to mutate the data through the data pointer while no other borrows exist
+        let data_ptr = account_info.data_ptr();
+        unsafe {
+            let data = core::slice::from_raw_parts_mut(data_ptr, 1); // Data is 1 byte long!
+            data[0] = 1;
+        }
 
         // Borrow immutable data (7 immutable borrows available).
         const ACCOUNT_REF: MaybeUninit<Ref<[u8]>> = MaybeUninit::<Ref<[u8]>>::uninit();
@@ -975,6 +982,8 @@ mod tests {
 
         // Borrow mutable data.
         let ref_mut = account_info.try_borrow_mut_data().unwrap();
+        // It should be sound to get the data pointer while the data is borrowed as long as we don't use it
+        let _data_ptr = account_info.data_ptr();
 
         // Check that we cannot borrow the data anymore.
         assert!(account_info.can_borrow_data().is_err());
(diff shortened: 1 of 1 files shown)


Aursen:
I'm in favor of this change too

sdk/pinocchio/src/account_info.rs
```diff
@@ -609,7 +609,7 @@ impl AccountInfo {
     }
 
     /// Returns the memory address of the account data.
-    fn data_ptr(&self) -> *mut u8 {
+    pub fn data_ptr(&self) -> *mut u8 {
         unsafe { (self.raw as *mut u8).add(core::mem::size_of::<Account>()) }
     }
```
febo, line 636: This is super unsafe, so let's flag it as such.

```suggestion
    ///
    /// # Safety
    ///
    /// This method is unsafe because it returns a raw pointer to the account data, thus leaving
    /// the borrow flag untouched. The caller must guarantee that there are no active borrows.
    ///
    /// Note that using any of the borrow data methods while holding the pointer retuned by this method
    /// could lead to undefined brehaviour.
    #[inline(always)]
    pub unsafe fn data_ptr(&self) -> *mut u8 {
        // SAFETY: The runtime serializes the account data after its "header".    
        unsafe { (self.raw as *mut u8).add(core::mem::size_of::<Account>()) }
    }
```

stegaBOB: Returning a raw pointer itself shouldn't be unsafe. Calling this method should be fine assuming the entrypoint serializes the account properly (if it doesn't, perhaps wrapping_add would be the safe way to go). 

The unsafe part would be on whoever is dereferencing the returned raw pointer, so it would make sense to document that, but I don't think it makes sense to mark this as unsafe too.

stegaBOB: > /// Note that using any of the borrow data methods while holding the pointer retuned by this method
> /// could lead to undefined behavior.

I don't think this is the case either. If you were to dereference the raw pointer that would definitely be a problem if there's an existing borrow, but just having a raw pointer to some data while a mutable reference to the data exists should be fine.

febo: > Returning a raw pointer itself shouldn't be unsafe. Calling this method should be fine assuming the entrypoint serializes the account properly (if it doesn't, perhaps wrapping_add would be the safe way to go).
> 
> The unsafe part would be on whoever is dereferencing the returned raw pointer, so it would make sense to document that, but I don't think it makes sense to mark this as unsafe too.

Yes, strickly speaking it is not unsafe, but I was trying to convey the idea that this is an `unsafe` API since you can easily end up with an UB – a bit similar to our discussion of the `owner_ref()` being an `unsafe` method.

Not sure whether the `unsafe` would be clear enough or end up being confusing. The point is that it should not be "encouraged" to use this over the borrows.

febo: > > /// Note that using any of the borrow data methods while holding the pointer retuned by this method
> > /// could lead to undefined behavior.
> 
> I don't think this is the case either. If you were to dereference the raw pointer that would definitely be a problem if there's an existing borrow, but just having a raw pointer to some data while a mutable reference to the data exists should be fine.

This is why it says "could lead to". 😉

febo: Perhaps doing a "safe" `data_ptr` with the borrow check and then an unsafe `data_ptr_unchecked` is a good compromise?

stegaBOB: I think it might be a bit confusing since the method isn't actually unsafe on its own. In std for example, methods that return raw pointers explain what circumstances invalidate them (in addition to all the normal pointer rules), but they aren't actually unsafe themselves: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_mut_ptr-1. 

> This is why it says "could lead to". 😉

I feel like this is a stretch hahaha. I'd argue it would be more accurate to say "dereferencing the raw pointer while references from borrow data methods are alive would be undefined behavior" or something along those lines.

sdk/pinocchio/src/account_info.rs
```diff
@@ -609,7 +609,7 @@ impl AccountInfo {
...
 
     /// Returns the memory address of the account data.
-    fn data_ptr(&self) -> *mut u8 {
+    pub fn data_ptr(&self) -> *mut u8 {
```
febo, line 634: Ok, so what about adding a "warning"?
```suggestion
    /// # Important
    ///
    /// Obtaining the raw pointer itself is safe, but dereferencing it requires
    /// the caller to uphold Rust’s aliasing rules. It is undefined behavior to dereference
    /// the pointer or write through it while any safe reference (e.g., from any of `borrow_data`
    /// or `borrow_mut_data` methods) to the same data is still alive.
    pub fn data_ptr(&self) -> *mut u8 {
```

febo:
@stegaBOB Could we also add a couple of tests so miri can validate the safety of this?

stegaBOB:
> @stegaBOB Could we also add a couple of tests so miri can validate the safety of this?

What circumstances do we want to test here?

sdk/pinocchio/src/account_info.rs
stegaBOB: Looks good to me!

febo (dismissed):
👌

febo (approved):

PR #233, opened 2025-08-26
entrypoint: remove unnecessary deser case

0x0ece:
Inspired by the work in https://github.com/solana-program/token/pull/81, I noticed that the initial deserialization had an unnecessary case. This is a hot path, therefore p-token.transfer() consumes 1 unnecessary CU to jump back.


febo (approved):
👍

PR #234, opened 2025-08-26
Safer Account

Buzzec:
Added safer code for Account. Removed copy and clone from Account. Changed lamports functions to operate on a Cell.

Should also be a bit more efficient in a few places.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ members = [
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/anza-xyz/pinocchio"
-rust-version = "1.79"
+rust-version = "1.80"
 
 [workspace.dependencies]
 five8_const = "0.1.4"
--- a/programs/token-2022/src/instructions/initialize_multisig.rs
+++ b/programs/token-2022/src/instructions/initialize_multisig.rs
@@ -85,8 +85,8 @@ impl InitializeMultisig<'_, '_, '_> {
         };
 
         // Account info array
-        const UNINIT_INFO: MaybeUninit<&AccountInfo> = MaybeUninit::uninit();
-        let mut acc_infos = [UNINIT_INFO; 2 + MAX_MULTISIG_SIGNERS];
+        let uninit_info: MaybeUninit<&AccountInfo> = MaybeUninit::uninit();
+        let mut acc_infos = [uninit_info; 2 + MAX_MULTISIG_SIGNERS];
 
         unsafe {
             // SAFETY:
--- a/programs/token/src/state/multisig.rs
+++ b/programs/token/src/state/multisig.rs
@@ -57,7 +57,7 @@ impl Multisig {
         if account_info.data_len() != Self::LEN {
             return Err(ProgramError::InvalidAccountData);
         }
-        if account_info.owner() != &ID {
+        if account_info.owner_is(&ID) {
             return Err(ProgramError::InvalidAccountOwner);
         }
         Ok(Self::from_bytes_unchecked(
--- a/programs/token/src/state/multisig.rs
+++ b/programs/token/src/state/multisig.rs
@@ -29,7 +29,9 @@ impl Multisig {
     /// This method performs owner and length validation on `AccountInfo`, safe borrowing
     /// the account data.
     #[inline]
-    pub fn from_account_info(account_info: &AccountInfo) -> Result<Ref<Multisig>, ProgramError> {
+    pub fn from_account_info(
+        account_info: &AccountInfo,
+    ) -> Result<Ref<'_, Multisig>, ProgramError> {
         if account_info.data_len() != Self::LEN {
             return Err(ProgramError::InvalidAccountData);
         }
--- a/programs/token/src/state/token.rs
+++ b/programs/token/src/state/token.rs
@@ -58,7 +58,7 @@ impl TokenAccount {
     #[inline]
     pub fn from_account_info(
         account_info: &AccountInfo,
-    ) -> Result<Ref<TokenAccount>, ProgramError> {
+    ) -> Result<Ref<'_, TokenAccount>, ProgramError> {
         if account_info.data_len() != Self::LEN {
             return Err(ProgramError::InvalidAccountData);
         }
--- a/programs/token/src/state/token.rs
+++ b/programs/token/src/state/token.rs
@@ -86,7 +86,7 @@ impl TokenAccount {
         if account_info.data_len() != Self::LEN {
             return Err(ProgramError::InvalidAccountData);
         }
-        if account_info.owner() != &ID {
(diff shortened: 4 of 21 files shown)


Cargo.toml
```diff
@@ -16,7 +16,7 @@ members = [
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/anza-xyz/pinocchio"
-rust-version = "1.79"
+rust-version = "1.80"
```
sonicfromnewyoke, line 19: maybe it'll be better to bump the version in the scope of the separate MR?

programs/token/src/state/token.rs
```diff
@@ -86,7 +86,7 @@ impl TokenAccount {
         if account_info.data_len() != Self::LEN {
             return Err(ProgramError::InvalidAccountData);
         }
-        if account_info.owner() != &ID {
+        if account_info.owner_is(&ID) {
```
sonicfromnewyoke: it breaks the logic

programs/token/src/state/multisig.rs
```diff
@@ -57,7 +59,7 @@ impl Multisig {
         if account_info.data_len() != Self::LEN {
             return Err(ProgramError::InvalidAccountData);
         }
-        if account_info.owner() != &ID {
+        if account_info.owner_is(&ID) {
```
sonicfromnewyoke: it breaks the logic, should be like follows 
```suggestion
        if !account_info.owner_is(&ID) {
```

sdk/pinocchio/src/account_info.rs
```diff
@@ -196,45 +234,22 @@ impl AccountInfo {
...
-    /// # Safety
-    ///
-    /// This method is unsafe because it does not return a `Ref`, thus leaving the borrow
-    /// flag untouched. Useful when an instruction has verified non-duplicate accounts.
-    #[allow(clippy::mut_from_ref)]
-    #[inline(always)]
-    pub unsafe fn borrow_mut_lamports_unchecked(&self) -> &mut u64 {
-        &mut (*self.raw).lamports
-    }
-
```
sonicfromnewyoke, line 237: what's the point of removing those 2 functions?

programs/token/src/state/token.rs
Buzzec: Whoops, will fix

programs/token/src/state/multisig.rs
Buzzec: Whoops will fix

sdk/pinocchio/src/account_info.rs
Buzzec: No longer necessary since lamports is a Cell. Making lamports a Cell makes these impossible, would have to be an UnsafeCell with more safety logic.

Cargo.toml
Buzzec: The issue is the add calls on NonNull, that was added in 1.80. Could keep it 1.79 but would need to call get, then add, then new_unchecked.

Buzzec:
Going to stack another PR to optimize the data locking now that lamports doesn't use it

Buzzec:
> Going to stack another PR to optimize the data locking now that lamports doesn't use it

Stacked PR: https://github.com/staratlasmeta/pinocchio/pull/1

Cargo.toml
Buzzec: Oof, CI is failing on this

febo:
@Buzzec This is great, thank you! Could you please bump the solana cli version to `2.3.0` here: https://github.com/anza-xyz/pinocchio/blob/main/Cargo.toml#L31

This will bump the rust version used in CI to `1.84.1`.

Buzzec:
> @Buzzec This is great, thank you! Could you please bump the solana cli version to `2.3.0` here: https://github.com/anza-xyz/pinocchio/blob/main/Cargo.toml#L31
> 
> This will bump the rust version used in CI to `1.84.1`.

Done

Buzzec:
With 1.84 we can fix the Miri issues as well, will do that in another PR

febo:
> With 1.84 we can fix the Miri issues as well, will do that in another PR

Good point – separate PR sounds good!

Buzzec:
> > With 1.84 we can fix the Miri issues as well, will do that in another PR
> 
> Good point – separate PR sounds good!

Stacked it https://github.com/staratlasmeta/pinocchio/pull/2

programs/token-2022/src/instructions/initialize_multisig.rs
```diff
@@ -85,8 +85,8 @@ impl InitializeMultisig<'_, '_, '_> {
         };
 
         // Account info array
-        const UNINIT_INFO: MaybeUninit<&AccountInfo> = MaybeUninit::uninit();
-        let mut acc_infos = [UNINIT_INFO; 2 + MAX_MULTISIG_SIGNERS];
+        let uninit_info: MaybeUninit<&AccountInfo> = MaybeUninit::uninit();
```
febo: Just wondering motivation for this change – is it more idiomatic to use a `let` in this scenario? I can see that the end result is the same, since the `let` is immutable.

Buzzec: Clippy warning about const mutability. Technically this would be fine but a const is a compile time inserted expression. So if there's interior mutability a const would have weird behavior.

~~`MaybeUninit` was marked as interior mutable for a while which caused the clippy lint to trigger. In more modern rust this isn't the case, rust even [recommends using a const block for this](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element).~~

`AccountInfo` has interior mutability. [Clippy Lint](https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const).

It only works here since the `MaybeUninit` is copy (`&T` is `Copy`)

Buzzec: Actually const blocks work here, will swap to that

febo:
@Buzzec Thanks for this, really nice stuff.  I started to look into the `Account`/`AccountInfo` changes. One thing that I noticed is that the performance of the entrypoint got a hit from the changes, can't pinpoint exactly where at the moment.

| Name         | CUs (before) | CUs (after) | Delta |
|--------------|--------------|-------------|-------|
| Account (1)  | 17           | 21          | +4    |
| Account (2)  | 30           | 37          | +7    |
| Account (3)  | 37           | 50          | +13   |
| Account (4)  | 45           | 63          | +18   |
| Account (8)  | 78           | 118         | +40   |
| Account (16) | 143          | 227         | +84   |
| Account (32) | 261          | 430         | +169  |
| Account (64) | 504          | 846         | +342  |

I am planning to add your changes incrementally to isolate what change is slowing things down.

Buzzec:
Any discoveries here?

PR #236, opened 2025-08-28
pinocchio: Add hint module

febo:
### Problem

It is possible to add hints to the compiler to help it optimize the code and save CUs – e.g., as it is used in [p-token](https://github.com/solana-program/token/pull/81).

### Solution

Add a new module `hint` with heper functions. This is similar to the experimental API from [`core::hint`](https://doc.rust-lang.org/core/hint/index.html).

Credits to @cavemanloverboy for this suggestion.

### DIFF
--- a/sdk/pinocchio/src/lib.rs
+++ b/sdk/pinocchio/src/lib.rs
@@ -264,3 +264,39 @@ pub const SUCCESS: u64 = 0;
 
 /// The result of a program execution.
 pub type ProgramResult = Result<(), program_error::ProgramError>;
+
+/// Module with functions to provide hints to the compiler about how code
+/// should be optimized.
+pub mod hint {
+    /// A "dummy" function with a hint to the compiler that it is unlikely to be
+    /// called.
+    ///
+    /// This function is used as a hint to the compiler to optimize other code paths
+    /// instead of the one where the function is used.
+    #[cold]
+    pub const fn cold_path() {}
+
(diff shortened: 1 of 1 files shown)


rustopian (approved):
Very nice and looking forward to using it.

sdk/pinocchio/src/lib.rs
```diff
@@ -264,3 +264,39 @@ pub const SUCCESS: u64 = 0;
...
+/// Module with functions to provide hints to the compiler about how code
+/// should be optimized.
+pub mod hint {
+    /// A "dummy" function with a hint to the compiler that it is unlikely to be
+    /// called.
+    ///
+    /// This function is used as a hint to the compiler to optimize other code paths
+    /// instead of the one where the function is used.
+    #[cold]
+    pub const fn cold_path() {}
```
rustopian, line 277: `#[inline(never)]` on `cold_path` doesn't seem to make a difference - perhaps there are some cases where it would - I'll try a bit more and if so will propose separately.

rustopian (approved):
Clearing my earlier suggestion for `#[inline(never)]`, seems not to matter. Approved as is.

sdk/pinocchio/src/lib.rs
rustopian: Added links to the mentioned code/discussion.

joncinque (approved):

PR #238, opened 2025-08-31
add syscalls-stubs

skrrb:
**problem**

pinocchio doesn't execute syscalls for non-solana targets, making it impossible to test/call program functions directly outside the solana runtime.

in the case of `solana_program`, this was solved with the [program stubs](https://docs.rs/solana-program/latest/solana_program/program_stubs/index.html)

**solution**
introduce a `syscalls-stubs` feature flag that replaces `black_box` hints with user custom definitions exposed via the `SyscallStubs` trait

### DIFF
--- a/sdk/pinocchio/Cargo.toml
+++ b/sdk/pinocchio/Cargo.toml
@@ -19,6 +19,7 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 
 [features]
 std = []
+syscalls-stubs = ["std"]
 
 [dev-dependencies]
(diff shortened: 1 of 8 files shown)


febo:
@skrrb Thanks for the PR. We are about to start the integration of pinocchio and SDK so we will be able to use the existing stub implementations.

skrrb:
> @skrrb Thanks for the PR. We are about to start the integration of pinocchio and SDK so we will be able to use the existing stub implementations.

cool. then i don't think it makes sense for this to be open anymore

PR #243, opened 2025-09-05
pinocchio: Use SIMD-0194 exemption threshold value

febo:
### Problem

[SIMD-0194](https://github.com/solana-foundation/solana-improvement-documents/pull/194) will deprecate the `exemption_threshold` value and set it to `1.0`. This will "break" the current logic to avoid `f64` operations in the minimum balance calculation.

### Solution

Use the SIMD-0194 value to avoid `f64` operations. This PR also adds deprecation notes for changes proposed in SIMD-0194.

### DIFF
--- a/sdk/pinocchio/src/sysvars/rent.rs
+++ b/sdk/pinocchio/src/sysvars/rent.rs
@@ -26,6 +26,12 @@ pub const RENT_ID: Pubkey = [
 /// - `$3.65` per megabyte year
 pub const DEFAULT_LAMPORTS_PER_BYTE_YEAR: u64 = 1_000_000_000 / 100 * 365 / (1024 * 1024);
 
+/// SIMD-0194
+///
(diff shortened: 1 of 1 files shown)


sdk/pinocchio/src/sysvars/rent.rs
```diff
@@ -150,6 +172,7 @@ impl Rent {
 
     /// Rent due for account that is known to be not exempt.
     #[inline]
+    #[allow(deprecated)]
     pub fn due_amount(&self, data_len: usize, years_elapsed: f64) -> u64 {
```
joncinque, line 176: This can be done in future work, but how about just deprecating this function?

febo: Yeah, that is a good point. We should be using the SDK crate for rent in the very near future, so we can make these changes there.

PR #118, opened 2025-04-01
Use `Address` from Solana SDK

febo:

PR #248, opened 2025-09-10
pinocchio: Add `pubkey_eq` helper

febo:
### Problem

In some cases, comparing pubkeys with the equality operator (`==`) results in "bloated" code, which eventually leads to an increased CU consumption.

### Solution

This PR adds a `pubkey_eq` helper, which compares pubkeys as a sequence of `u64` values.

### DIFF
--- a/sdk/pinocchio/src/pubkey.rs
+++ b/sdk/pinocchio/src/pubkey.rs
@@ -300,22 +300,22 @@ pub fn create_with_seed(
 
 #[cfg(test)]
 mod tests {
-    extern crate std;
-    use proptest::{
-        prelude::{any, prop_assert_eq},
-        proptest,
-    };
-    use std::format;
+    use crate::pubkey::{pubkey_eq, Pubkey, PUBKEY_BYTES};
 
-    use crate::pubkey::pubkey_eq;
+    #[test]
+    fn test_pubkey_eq_matches_default_eq() {
+        for i in 0..u8::MAX {
+            let p1: Pubkey = [i; PUBKEY_BYTES];
+            let p2: Pubkey = [i; PUBKEY_BYTES];
 
-    proptest! {
-            #[test]
-            fn test_pubkey_eq_matches_default_eq(
-                p1 in any::<[u8; 32]>(),
-                p2 in any::<[u8; 32]>()
-            ) {
-                prop_assert_eq!(pubkey_eq(&p1, &p2), p1 == p2);
-            }
+            assert_eq!(pubkey_eq(&p1, &p2), p1 == p2);
+        }
+
+        for i in 0..u8::MAX {
+            let p1: Pubkey = [i; PUBKEY_BYTES];
+            let p2: Pubkey = [u8::MAX - i; PUBKEY_BYTES];
+
+            assert_eq!(!pubkey_eq(&p1, &p2), p1 != p2);
+        }
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,55 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "autocfg"
-version = "1.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
-
-[[package]]
-name = "bit-set"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
-dependencies = [
- "bit-vec",
-]
-
-[[package]]
-name = "bit-vec"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
-
-[[package]]
-name = "bitflags"
-version = "2.9.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2261d10cca569e4643e526d8dc2e62e433cc8aba21ab764233731f8d369bf394"
-
-[[package]]
-name = "cfg-if"
-version = "1.0.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2fd1289c04a9ea8cb22300a459a72a385d7c73d3259e2ed7dcb2af674838cfa9"
-
-[[package]]
-name = "errno"
-version = "0.3.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
-dependencies = [
- "libc",
- "windows-sys",
-]
-
-[[package]]
-name = "fastrand"
-version = "2.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
-
-[[package]]
 name = "five8_const"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


sonicfromnewyoke (commented):
what do you think about `impl PartialEq for Pubkey` in the `sdk`?
to have API like follows:

```rust
mint_a.key().eq(mint_b.key())
```

febo:
> what do you think about `impl PartialEq for Pubkey` in the `sdk`? to have API like follows:
> 
> ```rust
> mint_a.key().eq(mint_b.key())
> ```

Yeah, definitely – we should have it on the `Address` [crate](https://github.com/anza-xyz/solana-sdk/tree/master/address).

sdk/pinocchio/src/sysvars/rent.rs
grod220, line 1: A potentially missed call site?

```rust
    pub fn from_account_info(account_info: &AccountInfo) -> Result<Ref<Rent>, ProgramError> {
        if account_info.key() != &RENT_ID {
        }
```

sdk/pinocchio/src/pubkey.rs
```diff
@@ -33,6 +35,24 @@ pub fn log(pubkey: &Pubkey) {
...
+    let p1_ptr = p1.as_ptr() as *const u64;
+    let p2_ptr = p2.as_ptr() as *const u64;
+
+    unsafe {
+        read_unaligned(p1_ptr) == read_unaligned(p2_ptr)
+            && read_unaligned(p1_ptr.add(1)) == read_unaligned(p2_ptr.add(1))
+            && read_unaligned(p1_ptr.add(2)) == read_unaligned(p2_ptr.add(2))
+            && read_unaligned(p1_ptr.add(3)) == read_unaligned(p2_ptr.add(3))
+    }
+}
```
grod220, line 54: Feels like a good candidate for prop testing. Something like

```rust
    proptest! {
        #[test]
        fn test_pubkey_eq_matches_default_eq(
            p1 in any::<[u8; 32]>(),
            p2 in any::<[u8; 32]>()
        ) {
            prop_assert_eq!(pubkey_eq(&p1, &p2), p1 == p2);
        }
}
```

sdk/pinocchio/src/cpi.rs
```diff
@@ -299,7 +299,7 @@ unsafe fn inner_invoke_signed_with_bounds<const MAX_ACCOUNTS: usize>(
             // In order to check whether the borrow state is compatible
             // with the invocation, we need to check that we have the
             // correct account info and meta pair.
-            if account_info.key() != account_meta.pubkey {
+            if !pubkey_eq(account_info.key(), account_meta.pubkey) {
```
rustopian: Any reason not to use `unlikely` here as well? Mismatched meta should be rare.

```suggestion
            if crate::hint::unlikely(!pubkey_eq(account_info.key(), account_meta.pubkey)) {
```

sdk/pinocchio/src/sysvars/rent.rs
febo: Good catch!

sdk/pinocchio/src/cpi.rs
febo: Good catch!

sdk/pinocchio/src/pubkey.rs
febo: Good idea! I tried adding `proptest` but created issues with `build-sbf`. I added a simple test instead.

sdk/pinocchio/src/pubkey.rs
```diff
@@ -33,6 +35,24 @@ pub fn log(pubkey: &Pubkey) {
     core::hint::black_box(pubkey);
 }
 
+/// Compare two `Pubkey`s for equality.
+///
+/// The implementation of this function is currently more efficient
+/// than `p1 == p2` since it compares 8 bytes at a time instead of
+/// byte-by-byte.
```
illia-bobyr, line 42: If this is just a more efficient implementation of `==`, it seems better to redefine `PartialEq`:

```rust
// Comparing `Pubkey`s 8 bytes at a time is faster, than comparing byte at a
// time.  One would expect LLVM to do this optimization for us, but it does not
// at the moment (as of platform tools v1.43).  Comparing keys is a pretty
// common operation, so we want to make it faster if we can.
impl PartialEq for Pubkey {
    fn eq(&self, other: &Self) -> bool {
        let p1 = self.as_ptr() as *const u64;
        let p2 = other.as_ptr() as *const u64;

        // SAFETY: `p1` and `p2` both point to `[u8; 32]` bytes of data, which
        // is the same size as `[u64; 4]`.
        unsafe {
            read_unaligned(p1) == read_unaligned(p2)
                && read_unaligned(p1.add(1)) == read_unaligned(p2.add(1))
                && read_unaligned(p1.add(2)) == read_unaligned(p2.add(2))
                && read_unaligned(p1.add(3)) == read_unaligned(p2.add(3))
        }
    }
}
```

**Bikeshedding**

Maybe it is possible to get the same effect if you just help the compiler a bit,
by casting `[u8; 32]` into `[u64; 4]`?

```rust
// Comparing `Pubkey`s 8 bytes at a time is faster than comparing byte at a
// time.  One would expect LLVM to do this optimization for us, but it does not
// at the moment (as of platform tools v1.43).  Comparing keys is a pretty
// common operation, so we want to make it faster if we can.
impl PartialEq for Pubkey {
    fn eq(&self, other: &Self) -> bool {
        let p1 = self.as_ptr() as *const [u64; 4];
        let p2 = other.as_ptr() as *const [u64; 4];

        // SAFETY: `p1` and `p2` both point to `[u8; 32]` bytes of data, which
        // is the same size as `[u64; 4]`.
        unsafe { read_unaligned(p1) == read_unaligned(p2) }
    }
}
```

I would also add `static_assertions`, to be able to encode the `SAFETEY` comment
in code:

```rust
use static_assertions::assert_eq_size;

// Comparing `Pubkey`s 8 bytes at a time is faster than comparing byte at a
// time.  One would expect LLVM to do this optimization for us, but it does not
// at the moment (as of platform tools v1.43).  Comparing keys is a pretty
// common operation, so we want to make it faster if we can.
impl PartialEq for Pubkey {
    fn eq(&self, other: &Self) -> bool {
        let p1 = self.as_ptr() as *const [u64; 4];
        let p2 = other.as_ptr() as *const [u64; 4];

        assert_eq_size!(*self, [u64; 4]);
        assert_eq_size!(*other, [u64; 4]);
        // SAFETY: `self` and `[u64; 4]` are of the same size.
        unsafe { read_unaligned(p1) == read_unaligned(p2) }
    }
}
```

(It is a compile time assertion with no runtime code.)

rustopian: I wanted this, too – at first. It would be great to get the savings everywhere devs use `==`.

But since our `pinocchio::pubkey::Pubkey` is just an alias for `[u8;32]`, we can't `impl PartialEq` on it. Creating a new type is maybe not worth it and probably a separate discussion.

rustopian: I wanted this, too – at first. It would be great to get the savings everywhere devs use `==`.

But since our `pinocchio::pubkey::Pubkey` is just an alias for `[u8;32]`, we can't `impl PartialEq` on it. Creating a new type is maybe not worth it and probably a separate discussion.

Also, [sounds like Lucas](https://github.com/anza-xyz/solana-sdk/pull/318#issuecomment-3287327164) is looking into this, so maybe this discussion will be irrelevant soon.

febo: > If this is just a more efficient implementation of `==`, it seems better to redefine `PartialEq`:

We have done this for the `solana-address` [here](https://github.com/anza-xyz/solana-sdk/pull/318), since `Address` is a type.

> **Bikeshedding**
> 
> Maybe it is possible to get the same effect if you just help the compiler a bit, by casting `[u8; 32]` into `[u64; 4]`?

Unfortunately no, it is even worse. 😅 I think the compiler tries to do something different when it sees an array, which in the end becomes "bloated" and uses more CU. You can use this [simple bench](https://github.com/febo/pubkey-comparison) to see the effect.

As @rustopian mentioned, Lucas will have a look at this to see if we can improve the code generation so the compiler does the "right" thing.

illia-bobyr: > I wanted this, too – at first. It would be great to get the savings everywhere devs use `==`.
> 
> But since our `pinocchio::pubkey::Pubkey` is just an alias for `[u8;32]`, we can't `impl PartialEq` on it. Creating a new type is maybe not worth it and probably a separate discussion.

This is out of scope for this PR obviously, but, considering `Pubkey` is such a fundamental abstraction, I think it should be a type.

febo: It will be very soon: #118 😊

PR #250, opened 2025-09-12
Bump toolchain version

febo:
### Problem

Currently pinocchio packages are build with rust `1.79`, which is quite old.

### Solution

This PR bumps the MSRV to <del>1.84.1</del>`1.84`, which is a version compatible with Solana CLI `2.3+`.

### DIFF
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -13,6 +13,7 @@ pub use alloc::BumpAllocator;
 use core::{
     cmp::min,
     mem::{size_of, MaybeUninit},
+    ptr::with_exposed_provenance_mut,
     slice::from_raw_parts,
 };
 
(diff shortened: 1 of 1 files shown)


grod220 (dismissed):
LGTM

Cargo.toml
```diff
@@ -16,7 +16,7 @@ members = [
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/anza-xyz/pinocchio"
-rust-version = "1.79"
+rust-version = "1.84.1"
```
joncinque: the rust-version declared here is actually the minimum supported rust version: https://doc.rust-lang.org/stable/cargo/reference/rust-version.html

So keeping it low is a good thing! The version of rust used by the repo is updated in rust-toolchain.toml, which already has 1.84.1.

So unless there's something that we need to drop support for, let's keep this at 1.79.

febo: I fixed a miri warning that needed `1.84` (need to use [`with_addr`](https://doc.rust-lang.org/core/primitive.pointer.html#method.with_addr)). I will change it for that.

joncinque: Ok great, that's a good enough reason for me :smile:

joncinque (approved):

rustopian (dismissed):
Looks good on my end.

joncinque (approved):
Looks great to me!

PR #251, opened 2025-09-15
Added `solana-address1` feature

Buzzec:
Optional feature that replaces `Pubkey` with `solana_address::Address`. Also adds a feature for `solana-pubkey2`


Buzzec:
This is nice for frameworks that use standard solana libraries like `star_frame` so that there aren't two different `Pubkey` types being used and converted between.

febo:
> This is nice for frameworks that use standard solana libraries like `star_frame` so that there aren't two different `Pubkey` types being used and converted between.

By coincidence, I am about to finish rebasing #118 which will switch the `Pubkey` for `Address`. That one is based on #112 as part of the integration of the SDK types.

Do you see a need to support solana-pubkey `2.*`? The plan was to go forward with solana-address.

Buzzec:
> > This is nice for frameworks that use standard solana libraries like `star_frame` so that there aren't two different `Pubkey` types being used and converted between.
> 
> By coincidence, I am about to finish rebasing #118 which will switch the `Pubkey` for `Address`. That one is based on #112 as part of the integration of the SDK types.
> 
> Do you see a need to support solana-pubkey `2.*`? The plan was to go forward with solana-address.

The reason I added it was that `star_frame` is stuck on that version. @stegaBOB knows specifically why, but I believe a few dependencies haven't updated their solana versions yet

stegaBOB:
> but I believe a few dependencies haven't updated their solana versions yet

Yup this is the reason. Just waiting for mollusk at the moment. That should be released soon I think though so no need to maintain support here

Buzzec:
Actually we're all updated now so closing this.

PR #252, opened 2025-09-17
log: Add cap to precision value

febo:
### Problem

When a precision value larger than the maximum number of digits of a type is used, an overflow occurs and leads to out-of-bounds writes.

### Solution

Add a maximum cap to the `Argument::Precision` to avoid an overflow. The cap is determined by the maximum number of digits that a type can have minus one.

### DIFF
--- a/sdk/log/crate/src/lib.rs
+++ b/sdk/log/crate/src/lib.rs
@@ -199,9 +199,47 @@ mod tests {
 
         logger.clear();
 
-        // This should have no effect.
+        // This should have no effect since it is a string.
         logger.append_with_args("0123456789", &[Argument::Precision(2)]);
         assert!(&*logger == "0123456789".as_bytes());
+
+        // Setting a precision equal to or greater than the maximum digits should have
+        // no effect. The value used will be capped at the maximum digits - 1.
+        let mut l1 = Logger::<50>::default();
+        let mut l2 = Logger::<50>::default();
+
+        l1.append_with_args(2u8, &[Argument::Precision(u8::MAX)]);
+        assert_eq!(&*l1, "0.02".as_bytes());
+        // same as 2 precision
+        l2.append_with_args(2u8, &[Argument::Precision(2)]);
+        assert_eq!(&*l1, &*l2);
+
+        l1.clear();
+        l2.clear();
+
+        l1.append_with_args(2u16, &[Argument::Precision(u8::MAX)]);
+        assert_eq!(&*l1, "0.0002".as_bytes());
+        // same as 4 precision
+        l2.append_with_args(2u16, &[Argument::Precision(4)]);
+        assert_eq!(&*l1, &*l2);
+
+        l1.clear();
+        l2.clear();
+
+        l1.append_with_args(2u32, &[Argument::Precision(u8::MAX)]);
+        assert_eq!(&*l1, "0.000000002".as_bytes());
+        // same as 9 precision
+        l2.append_with_args(2u32, &[Argument::Precision(9)]);
+        assert_eq!(&*l1, &*l2);
+
+        l1.clear();
+        l2.clear();
+
+        l1.append_with_args(2u64, &[Argument::Precision(u8::MAX)]);
+        assert_eq!(&*l1, "0.0000000000000000002".as_bytes());
+        // same as 19 precision
+        l2.append_with_args(2u64, &[Argument::Precision(19)]);
+        assert_eq!(&*l1, &*l2);
     }
 
     #[test]
--- a/sdk/log/crate/src/logger.rs
+++ b/sdk/log/crate/src/logger.rs
@@ -175,6 +175,10 @@ pub enum Argument {
     /// Number of decimal places to display for numbers.
     ///
     /// This is only applicable for numeric types.
+    ///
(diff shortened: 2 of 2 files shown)


sdk/log/crate/src/lib.rs
```diff
@@ -199,9 +199,47 @@ mod tests {
...
         logger.append_with_args("0123456789", &[Argument::Precision(2)]);
         assert!(&*logger == "0123456789".as_bytes());
+
+        // Setting a precision equal to or greater than the maximum digits should have
+        // no effect. The value used will be capped at the maximum digits - 1.
+        let mut l1 = Logger::<50>::default();
+        let mut l2 = Logger::<50>::default();
+
+        l1.append_with_args(2u8, &[Argument::Precision(u8::MAX)]);
+        assert_eq!(&*l1, "0.02".as_bytes());
```
joncinque: This behavior is a bit surprising, but I'm having a hard time coming up with a better option.

Maybe we prepend some other character to indicate that it's been truncated at the front? So if you set u8::MAX precision, you get "*002" or "?002" or "^002".

Prepending "..." seems excessive, since it requires allocating two extra bytes that will almost never be used.

As a side note, is there a reason to not support a precision of `MAX_DIGITS` and return ".002"?

febo: > This behavior is a bit surprising, but I'm having a hard time coming up with a better option.
> 
> Maybe we prepend some other character to indicate that it's been truncated at the front? So if you set u8::MAX precision, you get "*002" or "?002" or "^002".

Same here. I like your idea of using a character. We already use `@` when the string is truncated at the end. Perhaps `^` is a good choice, since the other kind of imply either any or missing character.

> Prepending "..." seems excessive, since it requires allocating two extra bytes that will almost never be used.
> 
> As a side note, is there a reason to not support a precision of `MAX_DIGITS` and return ".002"?

No reason, we can do that.

joncinque: oh right, I forgot about `@`, we can just use that if you want!

febo: Refactored the logic, so the precision is not restricted to the size of the type. That allows logging an `u8` value with precision `9` for example. This is probably a more predictable behaviour than "truncating" the precision.

The output will still be truncated based on the size of the buffer – e.g.:
```rust
let mut logger = Logger::<4>::default();
append_with_args(u8::MAX, &[Argument::Precision(9)]);
```
will result in`"0.0@"` .

sdk/log/crate/src/lib.rs
joncinque, line 1: Great tests!

sdk/log/crate/src/logger.rs
```diff
@@ -175,6 +185,10 @@ pub enum Argument {
     /// Number of decimal places to display for numbers.
     ///
     /// This is only applicable for numeric types.
+    ///
+    /// The precision cannot be equal to or greater than the maximum number
+    /// of digits for the type. If it is, it will be set to the maximum
```
joncinque: nit: up to you if you want to specify, but it *can* be set to something greater, but the output will get truncated, right?

sdk/log/crate/src/logger.rs
```diff
@@ -267,99 +281,130 @@ macro_rules! impl_log_for_unsigned_integer {
...
+                        // if the precision is greater than zero.
+                        let required = match precision {
+                            0 => written,
+                            // decimal point
+                            _precision if precision < written => written + 1,
+                            // decimal point + one leading zero
+                            _ => precision + 2,
                         };
+                        // Determines if the value will be truncated or not.
+                        let overflow = required > length;
```
joncinque: nit: can we call this `truncated` or `is_truncated`? `overflow` sounds like it should be a number

sdk/log/crate/src/logger.rs
```diff
@@ -267,99 +281,130 @@ macro_rules! impl_log_for_unsigned_integer {
...
+                        // Cap the number of digits to write to the buffer length.
+                        let written = min(MAX_DIGITS - offset, length);
```
joncinque: nit: they're not actually written yet, so this variable was a bit confusing. Maybe `num_to_write` would be clearer?

sdk/log/crate/src/logger.rs
```diff
@@ -267,99 +281,130 @@ macro_rules! impl_log_for_unsigned_integer {
...
                             }
                         }
 
-                        // There might not have been space for all the value.
+                        let written = min(required, length);
+
+                        // There might not have been space.
                         if overflow {
-                            // SAFETY: the buffer is checked to be within `written` bounds.
+                            // SAFETY: `written` is capped to the length of the buffer.
```
joncinque: nit: if you want to mention it, `buffer` is also guaranteed to have a length of at least 1, and `required` is guaranteed to be non-zero, making the `-1` part safe

sdk/log/crate/src/logger.rs
febo: Good catch! I removed the comment since it does not apply anymore. The output will be truncated according to the size of the buffer, so it is independent of the precision.

joncinque (approved):
Looks great!

PR #256, opened 2025-09-22
check account size & type if extensions

skrrb:
when deserializing accounts with extensions, it must be ensured that they don't have the same size as a mutisig and that the account is one of the proper types


febo:
#400 was created based on this PR, so closing this one.

PR #258, opened 2025-09-30
Update README.md

mishalturkane:
added latest version of pinocchio


Neocryptoquant:
I noticed that too. Prolly an oversight from devs

febo:
Fixed in #265

PR #259, opened 2025-09-30
Update lib.rs

Neocryptoquant:
Updated version to stay consistent with development


sonicfromnewyoke (commented):
i believe it's better to keep in the scope of #258

Neocryptoquant:
Oh!
That's great. Thank you.

febo:
Fixed in #265

PR #260, opened 2025-09-30
Update README.md

Neocryptoquant:
Response to @sonicfromnewyoke. Zero dependencies.


febo:
I am not sure if I get the point of this change.

febo:
Fixed in #265

PR #261, opened 2025-10-04
Remove log module

febo:
### Problem

Pinocchio is using common functionality from the SDK, but still replicated helpers for the `sol_log_*` syscalls.

### Solution

Remove the `log` module from pinocchio. Programs that require logging should either use `solana-msg` or the soon to be created `solana-program-log`.

Logging is still used in the custom panic handler implementation – this is replaced by a direct call to `sol_log_` syscall.


febo:
> Looks good to me, but would the examples be clearer if they used solana-program-log? This way people know how to mix and match pinocchio with the sdk

This is a good point. Do you think we need to add to all examples, or just one (the first)?

rustopian (dismissed):
Good add

joncinque (dismissed):
One example is fine with me -- we can always add more later if needed. Looks great!

LStan:
`msg` macro and `sol_log` function are in `solana-msg`. But what about others (`sol_log_64`, `sol_log_data`, etc.)? They are not in `solana-program-log`. Are they not needed?

joncinque (approved):
Looks good to me

febo:
> `msg` macro and `sol_log` function are in `solana-msg`. But what about others (`sol_log_64`, `sol_log_data`, etc.)? They are not in `solana-program-log`. Are they not needed?

This is a good point. They are in `solana-define-syscall`, but we should probably reexport them in `solana-program-log` as well.

PR #262, opened 2025-10-04
Use `solana-define-syscall` crate

febo:
### Problem

Pinocchio is using common functionality from the SDK, but still duplicates the syscall definitions.

### Solution

This PR adds the `solana-define-syscall` and removes the duplicated syscall definitions.

### DIFF
--- a/sdk/pinocchio/src/lib.rs
+++ b/sdk/pinocchio/src/lib.rs
@@ -230,7 +230,7 @@ pub use solana_address as address;
 pub use solana_address::Address;
 
 // Re-export the `solana_define_syscall` for downstream use.
-#[cfg(target_os = "solana")]
+#[cfg(any(target_os = "solana", target_arch = "bpf"))]
 pub use solana_define_syscall::definitions as syscalls;
 
 // Re-export the `solana_instruction_view` for downstream use.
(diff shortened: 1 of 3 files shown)


sdk/pinocchio/src/lib.rs
```diff
@@ -228,6 +227,10 @@ pub use solana_account_view::AccountView;
 pub use solana_address as address;
 pub use solana_address::Address;
 
+// Re-export the `solana_define_syscall` for downstream use.
+#[cfg(target_os = "solana")]
```
joncinque: To go with the other change:
```suggestion
#[cfg(any(target_os = "solana", target_arch = "bpf"))]
```

sdk/pinocchio/src/sysvars/clock.rs
```diff
@@ -1,5 +1,7 @@
 //! Information about the network's clock, ticks, slots, etc.
 
+#![allow(deprecated)]
```
joncinque, line 5: Why is this needed? For the `get_clock` syscall? If so, can you specify to remove the annotation once the generic `get_sysvar` is used?

sdk/pinocchio/src/sysvars/rent.rs
```diff
@@ -2,6 +2,8 @@
 //!
 //! This is required for the rent sysvar implementation.
 
+#![allow(deprecated)]
```
joncinque, line 7: Same here, let's add an explanation

joncinque (commented):
A few small bits, looks great overall!

sdk/pinocchio/src/sysvars/clock.rs
febo: Yes, exactly. Added a comment as suggested.

joncinque (approved):
Looks great!

Issue #111, opened 2025-03-27
Integrating pinocchio and SDK types


febo:
### Problem

Pinocchio re-implements many types from [Solana SDK](https://github.com/anza-xyz/solana-sdk) to avoid non-essential dependencies. Since now the SDK has been divided into separate crates and they are much leaner in terms of external (non-SDK) dependencies, it is possible to use them in pinocchio. The main benefit is to make it easier to write code that works with both SDK and pinocchio once they share the same basic types.

Pinocchio will continue to be a zero-dependency library, with the caveat the "zero-dependency" means zero *external* (non-SDK) dependencies. Therefore, it will continue to mitigate dependency issues. It will also continue to be a zero-copy first library, redefining or creating new types when needed.

The principle of "zero-dependencies" is not to avoid having external dependencies at all, but to have them *opt-in* – it should always be possible to write a full-featured program without external dependencies.

This issue is used to track the different PRs that bring SDK types into pinocchio in preparation for the next release.

febo:
Work completed with `0.10` release.

PR #263, opened 2025-10-06
Replace `std` for `alloc`

febo:
### Problem

Currently, pinocchio offers a `"std"` feature to enable code that requires memory allocation using the `std` library, although the same functionality is available through `alloc`.

### Solution

Replace the `"std"` feature by `"alloc"` and use the `alloc` library instead.

The `"alloc"` feature is enabled by default and can be disabled by adding the dependency with `default-features = false`.

### DIFF
--- a/sdk/pinocchio/Cargo.toml
+++ b/sdk/pinocchio/Cargo.toml
@@ -33,4 +33,5 @@ solana-program-error = { workspace = true }
 solana-define-syscall = { workspace = true }
 
 [dev-dependencies]
-five8_const = { workspace = true }
+solana-address = { workspace = true, features = ["decode"] }
+pinocchio = { path = ".", features = ["alloc"] }
--- a/sdk/pinocchio/src/entrypoint/mod.rs
+++ b/sdk/pinocchio/src/entrypoint/mod.rs
@@ -653,12 +653,10 @@ mod alloc {
     }
 }
 
-#[cfg(all(test, not(target_os = "solana")))]
+#[cfg(test)]
 mod tests {
-    extern crate alloc;
-
-    use alloc::{
-        alloc::{alloc, dealloc},
+    use ::alloc::{
+        alloc::{alloc, dealloc, handle_alloc_error},
         vec,
     };
     use core::{alloc::Layout, ptr::copy_nonoverlapping};
@@ -684,7 +682,7 @@ mod tests {
             unsafe {
                 let ptr = alloc(layout);
                 if ptr.is_null() {
-                    alloc::alloc::handle_alloc_error(layout);
+                    handle_alloc_error(layout);
                 }
                 AlignedMemory { ptr, layout }
             }
--- a/sdk/pinocchio/src/lib.rs
+++ b/sdk/pinocchio/src/lib.rs
@@ -210,7 +210,7 @@
 //! cargo build-sbf --features bpf-entrypoint
 //! ```
 
-#![cfg_attr(not(test), no_std)]
+#![no_std]
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,15 +22,6 @@ dependencies = [
 
 [[package]]
 name = "five8_const"
-version = "0.1.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "26dec3da8bc3ef08f2c04f61eab298c3ab334523e55f076354d6d6f613799a7b"
-dependencies = [
- "five8_core",
-]
-
-[[package]]
-name = "five8_const"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1a0f1728185f277989ca573a402716ae0beaaea3f76a8ff87ef9dd8fb19436c5"
@@ -54,7 +45,7 @@ checksum = "78ca9ab1a0babb1e7d5695e3530886289c18cf2f87ec19a575a0abdce112e3a3"
 name = "pinocchio"
 version = "0.9.2"
 dependencies = [
- "five8_const 0.1.4",
+ "pinocchio",
  "solana-account-view",
  "solana-address",
  "solana-define-syscall",
@@ -190,7 +181,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e37320fd2945c5d654b2c6210624a52d66c3f1f73b653ed211ab91a703b35bdd"
 dependencies = [
  "five8",
- "five8_const 1.0.0",
+ "five8_const",
  "solana-define-syscall",
  "solana-program-error",
 ]
--- a/sdk/pinocchio/src/sysvars/slot_hashes/test.rs
+++ b/sdk/pinocchio/src/sysvars/slot_hashes/test.rs
@@ -9,8 +9,7 @@ use core::{
     ptr,
 };
 
-use std::io::Write;
-use std::vec::Vec;
+use alloc::vec::Vec;
 
 #[test]
 fn test_layout_constants() {
@@ -30,7 +29,10 @@ fn test_layout_constants() {
     );
 
     pub fn check_base58(input_bytes: &[u8], expected_b58: &str) {
-        assert_eq!(five8_const::decode_32_const(expected_b58), input_bytes);
+        assert_eq!(
+            Address::from_str_const(expected_b58).as_array(),
+            input_bytes
+        );
     }
 
     check_base58(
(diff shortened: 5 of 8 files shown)


Aursen:
The main usage of the old std was for the alloc one (all other usages are available in the core). I think it's a good point to force people to build an sdk outside of the program itself

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -5,11 +5,8 @@ pub mod lazy;
 
 pub use lazy::{InstructionContext, MaybeAccount};
 
-#[cfg(not(feature = "std"))]
+#[cfg(not(feature = "alloc"))]
 use core::alloc::{GlobalAlloc, Layout};
```
illia-bobyr: This comment is not directly related to the PR goal.
But as I was trying to understand all the cases how features can be enabled and disabled, I think I found some things to discuss.

It seems that feature flags are used in a way that might cause problems.

In Rust, features need to work in an "additive" way.
A library can not remove functionality in response to a feature being enabled.
As it would potentially break compilation, if a feature is enabled due to a deduplication in the dependency graph.

`cargo` combines all the features requested by all instances of a given library in the dependency tree. [^1]

Let's say we have an application that depend on `pinocchio` and specifies `default-features = false` to disable `alloc`.  It then uses `no_allocator` for whatever reason.
The application code compiles correctly.

Next, the application adds a dependency on some library, say `pinocchio-token-mu`, that in turn also depend on `pinocchio`.
If `pinocchio-token-mu` does not specify `default-features = false` (which is reasonably may do, as it does not care for the allocators), it will effectively have a dependency on `pinocchio` with the `alloc` feature enabled.

`cargo` will combine both instances of `pinocchio` in the dependency graph, using a superset of features.
In this case, it would be `alloc`.

And now, the main application will stop compiling, as `no_allocator` will disappear due to the `alloc` feature being enabled just because of the new dependency on `pinocchio-token-mu`.

I think, it should be fine to just not disable functionality in this particular case.
`alloc` will provide the `BumpAllocator`, but it should not remove `NoAllocator`.

[^1]: To be precise, `cargo` combines all instances that are version compatible.  Version incompatible instances are treated as different libraries.

README.md
```diff
@@ -171,11 +171,12 @@ pub fn process_instruction(
 By default, Pinocchio is a `no_std` crate. This means that it does not use any code from the
 standard (`std`) library. While this does not affect how Pinocchio is used, there is a one
 particular apparent difference. Helpers that need to allocate memory, such as fetching `SlotHashes`
-sysvar data, are not available. To enable these helpers, the `std` feature must be enabled when adding
-Pinocchio as a dependency:
+sysvar data, are not available. To enable these helpers, the `alloc` feature must be enabled when
+adding Pinocchio as a dependency:
 ```
-pinocchio = { version = "0.10.0", features = ["std"] }
+pinocchio = { version = "0.10.0", features = ["alloc"] }
```
illia-bobyr: In `sdk/pinocchio/Cargo.toml` `alloc` is included in the default feature set:

```toml
[features]
alloc = []
# ...
default = ["alloc"]
```

I do not have an opinion either way, but just pointing out a discrepancy.

If we want `alloc` to be enabled by default, `README.md` and a doc-comment in `sdk/pinocchio/src/lib.rs` should be updated.

If we do not want `alloca` to be enabled by default, then `sdk/pinocchio/Cargo.toml` should be updated.

sdk/pinocchio/src/entrypoint/mod.rs
febo: Oh, that is a very good point! And you are right, we can keep the `NoAllocator` since it does not create a problem with `alloc` being enabled.

README.md
febo: Yes, the readme fix is here: https://github.com/anza-xyz/pinocchio/pull/265

sdk/pinocchio/src/entrypoint/mod.rs
febo: Hmmm, having a second thought on this. I think we need to keep the current behaviour. It would be a mistake to allow the `no_allocator!()` be used while there is code that enabled the `"alloc"` feature.

The tricky part is that `"alloc"` is being enabled by default. The ideal option would be not to do this, but I feel that it will make it harder to use since majority of programs will probably use an allocator.

So we kind of have 2 scenarios:

1. The one that you mentioned, where using `no_allocator!` and having another dependency that enables `"alloc"` won't compile;
2. Not enable `"alloc"` by default and most of programs will need to manually enabled it.

I am inclined to go with problem 1, since I would consider this a more "advanced" configuration.

illia-bobyr: > Hmmm, having a second thought on this. I think we need to keep the current behaviour. It would be a mistake to allow the `no_allocator!()` be used while there is code that enabled the `"alloc"` feature.

When you are saying "it would be a mistake" - would it produce undefined behavior?
I do not know the details of the `no_allocator!()` scenario.
Is the expectation that all other libraries then need to expose the `alloc` flag in a similar manner?
Such that if someone wants to use this library with Pinocchio in non-`alloc` mode, they would specify `default-features = false` for all the involved libraries?

`no_allocator!` documentation says that the idea is to prevent the program from allocating anything by providing an allocator that panics.
It is really necessary to break the whole application compilation if any library is using `alloc`?
The code in question may not be used.
It might be just one of the abstractions provided by the library.

README.md
illia-bobyr: This is a draft PR, and I can see that #265 is a pretty substantial update of the docs.
But I would still recommend including the relevant changes in this commit as well.
This way the commit is more self-contained.
In particular, if you cherry-pick it or revert it, or search through the change history, it could be helpful.

Considering we are talking about a documentation update, it is less important.

Relative order of inclusion for this PR and #265 also matters here, I guess.
Depending on which one goes in first, the changes may look different.
But #265 is also effectively incorrect, if I'm being pedantic, if it is merged first.

sdk/pinocchio/src/entrypoint/mod.rs
febo: The `no_allocator` macro sets a global allocator that panics, so anything that tries to allocate memory will panic in runtime.

We can't prevent the program of trying to use a type that needs an allocator (e.g., Vec) but we can prevent enabling the feature "alloc" and using the macro.

We could also allow it and let it panic at runtime. As you mentioned, chances are that no one will use the `no_allocator` macro with the "alloc" feature enabled.

illia-bobyr: > We could also allow it and let it panic at runtime. As you mentioned, chances are that no one will use the `no_allocator` macro with the "alloc" feature enabled.

I think, the more important case is the other way around.
If I use `no_allocator` in my program, I might still want to use functionality from a library that depend on Pinocchio with `alloc`.
I just would need to be careful to avoid using any parts of this library that may allocate.
Ideally, the library could guide me and similarly mark all code that allocates with an `alloc` feature gate of its own.
But the chances are it would not.

So, I think, maybe it is fine to always provide `no_allocator!()`?
`alloc` will still be required for all the functionality that requires allocation.

febo: Yeah, this makes sense. I made the change.

README.md
febo: Agreed, made the changes on this PR.

illia-bobyr:
Maybe someone else needs to look at this change, but it looks good to me.
If it compiles, it probably is correct :)
If you want, I can approve it.

sdk/pinocchio/src/lib.rs
rustopian:
```suggestion
//! ## `alloc` crate feature
```

rustopian (approved):
Looks great. One small comment update

sdk/pinocchio/src/entrypoint/mod.rs
```diff
@@ -653,34 +653,15 @@ mod alloc {
...
-        panic!("** NO ALLOCATOR **");
-    }
-
-    #[inline]
-    unsafe fn dealloc(&self, _: *mut u8, _: Layout) {
-        // I deny all allocations, so I don't need to free.
-    }
-}
-
 #[cfg(all(test, not(target_os = "solana")))]
```
joncinque: How about running the tests with the alloc feature enabled? Just be sure to include pinocchio as a dev-dependency with alloc enabled. I think you can then remove the `#![cfg_attr(not(test), no_std)]` from lib.rs

febo: Nice, this looks better. Removed some references to `std` in the tests as well (https://github.com/anza-xyz/pinocchio/pull/263/commits/dff6623333e0cec395e27664ebab725c84178186).

febo:
> Also, have you already added `target_arch = "bpf"` in one of the later PRs?

Yes, it is the last PR of the list: https://github.com/anza-xyz/pinocchio/issues/111

sdk/pinocchio/Cargo.toml
```diff
@@ -32,4 +33,5 @@ solana-program-error = { workspace = true }
 solana-define-syscall = { workspace = true }
 
 [dev-dependencies]
-five8_const = { workspace = true }
+solana-address = { workspace = true, features = ["decode"] }
```
febo, line 36: We can use the address with the `"decode"` feature to create an `Address` from an `str` (used in a test).

PR #266, opened 2025-10-14
Add Custom Errors for Better Debugging

### LINKED ISSUE #253: Feature request: Add require and error_code equivalent of anchor
### Motivation
After writing some programs in pinocchio, I realised that there's a redundancy to check specific requirements, e.g. checking whether a account is a signer

I think it's best if there's a helper function like require that checks these requirements, and a error_code equivalent of anchor to allow for custom errors

0xzrf:
# Fixes #253

## Overview
This PR introduces **custom error definitions** and **enhanced logging** for program errors.  

Currently, when a custom error occurs, the output is not very informative:

Current error output

This lack of clarity increases debugging time for developers and can be particularly problematic when deployed on-chain.

---

## Problem
- Developers struggle to quickly identify the source of errors in their programs.
- Debugging on-chain contracts becomes time-consuming and error-prone.

---

## Solution
By defining custom errors and leveraging enhanced logging, this PR aims to **reduce debugging time significantly** for contracts built on Pinocchio.

---

## Implementation
1. **Define a custom error** in your program using the `pinocchio_err` crate:

Custom error definition

2. **Use the `require!` macro** from the `pinocchio` crate to return program errors:

Using require macro

---

## Result
When a custom error occurs, the output becomes clear and actionable:

Improved error output

This improvement helps developers quickly identify and resolve issues, making on-chain program debugging far more efficient.

### DIFF
--- a/sdk/err/src/lib.rs
+++ b/sdk/err/src/lib.rs
@@ -1,0 +1,62 @@
+use pinocchio::program_error::ProgramError;
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Data, DeriveInput};
+
+/// Defines a custom error macro to create custom errors in pinocchio framework
+#[proc_macro_derive(ErrorCode, attributes(msg))]
+pub fn error_code(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = &input.ident;
+    let variants = if let Data::Enum(data_enum) = &input.data {
+        &data_enum.variants
+    } else {
+        return syn::Error::new_spanned(name, "ErrorCode can only be derived for enums")
+            .to_compile_error()
+            .into();
+    };
+
+    let match_arms = variants.iter().map(|variant| {
+        let variant_name = &variant.ident;
+        let msg = variant
+            .attrs
+            .iter()
+            .find(|attr| attr.path.is_ident("msg"))
+            .and_then(|attr| attr.parse_meta().ok())
+            .and_then(|meta| {
+                if let syn::Meta::NameValue(nv) = meta {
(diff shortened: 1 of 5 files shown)


sdk/err/src/lib.rs
```diff
@@ -0,0 +1,62 @@
+use pinocchio::program_error::ProgramError;
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Data, DeriveInput};
+
+/// Defines a custom error macro to create custom errors in pinocchio framework
+#[proc_macro_derive(ErrorCode, attributes(msg))]
+pub fn error_code(input: TokenStream) -> TokenStream {
```
febo, line 8: This seems to replicate the logic from [thiserror](https://crates.io/crates/thiserror), so not sure if that is necessary.

sdk/pinocchio/src/program_error.rs
```diff
@@ -284,6 +284,38 @@ impl ToStr for ProgramError {
...
+            {
+                if false {
+                    // just to scope-check type
+                } else {
+                    #[allow(unused_unsafe)]
+                    unsafe {
+                        $crate::msg!($error.message());
+                    }
+                }
+            }
```
febo, line 313: This is super strange. I am not sure why it is not written just like:
```rust
$crate::msg!($error.message());
```
Also, it is inefficient in terms of program size since it will generate a call to the log syscall in every use of `require!`. It would be much better to follow a pattern where any error is printed at the end of the `process_instruction`, similar to p-token: https://github.com/solana-program/token/blob/main/pinocchio/program/src/entrypoint.rs#L258

febo:
Most likely this should be done by a higher-level framework. I am closing this for now.

PR #267, opened 2025-10-27
Add Token2022 Extensions

burhankhaja:
**Problem**
Pinocchio lacks the support for Token2022 extensions.

**Solution**
Added full Token2022 extension support, including instruction handlers and logic for complete functionality. All tests pass, confirming compatibility with the Token2022 standard.


burhankhaja:
**Pending Extensions**

* **Token Metadata**
* **Transfer Fee**
* **Confidential Extensions**

programs/token-2022/src/extension/transfer_hook/state.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub const AUTHORITY_START: usize = 170;
+
+    pub const BASE_LEN: usize = core::mem::size_of::<TransferHook>();
+
+    pub const LEN: usize = Self::AUTHORITY_START + Self::BASE_LEN;
+
+    /// Return a `TransferHook` from the given bytes (unsafe, unchecked).
+    #[inline(always)]
+    pub unsafe fn from_bytes_unchecked(bytes: &[u8]) -> &Self {
+        &*(bytes[Self::AUTHORITY_START as usize..].as_ptr() as *const TransferHook)
```
febo, line 36: It is a bit strange to use a fix offset to where the extension will start since it depends on what other extensions are enabled. This is why the original token-2022 code looks up where the extension data is: https://github.com/solana-program/token-2022/blob/main/interface/src/extension/mod.rs#L148-L183

programs/token-2022/src/extension/transfer_hook/instructions/initialize.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl InitializeTransferHook<'_> {
+    #[inline(always)]
+    pub fn invoke(&self) -> ProgramResult {
+        self.invoke_signed(&[])
+    }
+
+    #[inline(always)]
+    pub fn invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
```
febo, line 33: There are no signers on this instruction, so we should only have `invoke`.

programs/token-2022/src/extension/transfer_hook/instructions/initialize.rs
```diff
@@ -0,0 +1,52 @@
...
+    instruction::{AccountMeta, Instruction, Signer},
+    pubkey::Pubkey,
+    ProgramResult,
+};
+
+pub struct InitializeTransferHook<'a> {
+    /// Mint Account to initialize.
+    pub mint_account: &'a AccountInfo,
+    /// Optional authority that can set the transfer hook program id
+    pub authority: Option<&'a Pubkey>,
```
febo, line 19: Should use a different lifetime than `AccountInfo` references. This applies to all `Pubkey` values.
```suggestion
    pub authority: Option<&'b Pubkey>,
```

programs/token-2022/src/extension/transfer_hook/instructions/initialize.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use pinocchio::{
+    account_info::AccountInfo,
+    cpi::invoke_signed,
+    instruction::{AccountMeta, Instruction, Signer},
+    pubkey::Pubkey,
+    ProgramResult,
+};
+
+pub struct InitializeTransferHook<'a> {
```
febo, line 15:
```suggestion
pub struct InitializeTransferHook<'a, 'b> {
```

cpi-tests/README.md
```diff
@@ -0,0 +1,215 @@
+# Pinocchio Interface Test Setup for SPL Token-2022
```
febo, line 1: The `cpi-tests` should under the `token-2022` folder since it is specific to it.

programs/token-2022/src/extension/cpi_guard/state.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct CpiGuard {
+    /// Lock privileged token operations from happening via CPI
+    pub lock_cpi: u8,
+}
+
+impl CpiGuard {
+    /// The length of the account with `CpiGuard` extension data
+    const LEN: u8 = 171;
+    /// The index where CPI guard data starts in the account with `CpiGuard` extension data
+    const CPI_GUARD_START: u8 = 170;
```
febo, line 24: Can't assume a fixed position for the start of the extension data.

febo:
Thanks for your contribution - it is a good start. Left a few comments that apply to the majority of the extensions. In general:
* Instructions that do not expect signers do not need an `invoke_signer`.
* We cannot have fixed offset for extension state – it all depends on which extensions are enabled on the account.
* Lifetime of values on the instruction builders need revision – values that can come from a different source (e.g., `AccountInfo` and `Pubkey`) should have different lifetimes.

The PR is also quite big so I suggest breaking it in smaller ones. Start adding the instruction builder for each extension under `src/instructions/extensions`. The state is a bit more complicated and should be on its own PR – see #225.

burhankhaja:
Thanks for the detailed feedback, Febo. The point about `fixed offsets` makes complete sense, that is a real issue on our end, we will make sure to address it in our upcoming pull requests. 

We will also address the other recommendations accordingly, separating `lifetimes`, removing unnecessary `invoke_signers`, adding extensions under `src/instructions/extensions,` and moving cpi-tests under `token-2022`

Just to confirm, when we split the extensions into separate PRs, should each extension include its own tests within the same PR, or should we submit the tests in separate PRs afterward?

febo:
Thanks for the initial work, the extension were divided into separate PRs and mostly merged on main.

burhankhaja:
Hasta la vista

God bless

PR #269, opened 2025-11-02
Add MemoTransfer extension

burhankhaja:
**Summary:**

Implements instruction builders for MemoTransfer extension as per previous feebacks, split from the larger Token2022 draft PR 267  to make review incremental and focused.

**Key Updates:**
- Added ExtensionDiscriminator enum for identifying discriminators
- Added MemoTransfer instruction builders: `enable` ,  `disable`


burhankhaja:
Closed for refactoring purposes, so that we keep extension scaffold separate , plus there was an issue with using `invoke_with_bounds`, instead of `invoke_signed_with_bounds` , in cases where the token account is managed by escrows or any type of programs

Issue #270, opened 2025-11-04
AdvanceNonceAccount uses 1-byte discriminator [4] instead of 4-byte [4,0,0,0]


GalPeleg1:
---

### **Description**
The `pinocchio_system::instructions::AdvanceNonceAccount` instruction builds its data as `[4]`, but the Solana System Program expects a 4-byte little-endian discriminator (`[4,0,0,0]`).  
This mismatch causes `invalid instruction data` errors when invoking the System Program.

---

### **Fix**
```rust
// Correct discriminator for SystemInstruction::AdvanceNonceAccount
let data: [u8; 4] = [4, 0, 0, 0];

let instruction = Instruction {
    program_id: &crate::ID,
    accounts: &account_metas,
    data: &data,
};

febo:
Fixed by #271

PR #271, opened 2025-11-09
Update advance_nonce_account.rs

dedeleono:
This update builds the instruction with a const [4, 0, 0, 0] payload, ensuring the System Program receives the 4-byte little-endian discriminator it expects. This prevents the invalid instruction data errors previously occurring.

Ref: Issue #270

### DIFF
--- a/programs/system/src/instructions/advance_nonce_account.rs
+++ b/programs/system/src/instructions/advance_nonce_account.rs
@@ -37,7 +37,7 @@ impl AdvanceNonceAccount<'_> {
             AccountMeta::readonly_signer(self.authority.key()),
         ];
 
-        // instruction data uses 4-byte LE discriminator expected by the system program
+        // instruction
         let instruction = Instruction {
             program_id: &crate::ID,
             accounts: &account_metas,
--- a/programs/system/src/instructions/upgrade_nonce_account.rs
+++ b/programs/system/src/instructions/upgrade_nonce_account.rs
@@ -21,7 +21,7 @@ impl UpgradeNonceAccount<'_> {
         // account metadata
         let account_metas: [AccountMeta; 1] = [AccountMeta::writable(self.account.key())];
 
-        // instruction data uses 4-byte LE discriminator expected by the system program
+        // instruction
         let instruction = Instruction {
             program_id: &crate::ID,
             accounts: &account_metas,


programs/system/src/instructions/advance_nonce_account.rs
```diff
@@ -37,11 +37,12 @@ impl AdvanceNonceAccount<'_> {
...
         ];
 
-        // instruction
+        // instruction data uses 4-byte LE discriminator expected by the system program
+        const DATA: [u8; 4] = [4, 0, 0, 0];
         let instruction = Instruction {
             program_id: &crate::ID,
             accounts: &account_metas,
-            data: &[4],
+            data: &DATA,
```
sonicfromnewyoke: maybe this way will be simpler?
```suggestion
            data: &[4, 0, 0, 0],
```

programs/system/src/instructions/advance_nonce_account.rs
```diff
@@ -37,11 +37,12 @@ impl AdvanceNonceAccount<'_> {
             AccountMeta::readonly_signer(self.authority.key()),
         ];
 
-        // instruction
+        // instruction data uses 4-byte LE discriminator expected by the system program
+        const DATA: [u8; 4] = [4, 0, 0, 0];
```
sonicfromnewyoke: use this array directly
```suggestion
```

programs/system/src/instructions/advance_nonce_account.rs
dedeleono: true haha, just made a new commit with the change and also updated upgrade_nonce_account.rs

programs/system/src/instructions/advance_nonce_account.rs
```diff
@@ -37,11 +37,11 @@ impl AdvanceNonceAccount<'_> {
             AccountMeta::readonly_signer(self.authority.key()),
         ];
 
-        // instruction
+        // instruction data uses 4-byte LE discriminator expected by the system program
```
febo: nit: Could we keep the existing comment? I think it is a bit redundant to have it on every instruction and other don't have it.

programs/system/src/instructions/upgrade_nonce_account.rs
```diff
@@ -21,11 +21,11 @@ impl UpgradeNonceAccount<'_> {
         // account metadata
         let account_metas: [AccountMeta; 1] = [AccountMeta::writable(self.account.key())];
 
-        // instruction
+        // instruction data uses 4-byte LE discriminator expected by the system program
```
febo: nit: Same here, could we keep the existing comment?

febo (commented):
This is great, thank you! I will fix the CI issue so we can get this merged.

programs/system/src/instructions/upgrade_nonce_account.rs
dedeleono: Yup, just pushed the new commit to keep the existing comment

febo:
@dedeleono If you rebase the PR, CI should be able to run.

dedeleono:
> Update

Nice, just did! @febo

PR #274, opened 2025-11-15
Token2022: Add TransferHook extension instruction builders

burhankhaja:
**Key Updates:** 
- Implements instruction builders for TransferHook extension 

**Note:** 
- This PR is split from the larger Token2022 draft PR #267

### DIFF
--- a/programs/token-2022/src/instructions/extensions/transfer_hook/initialize.rs
+++ b/programs/token-2022/src/instructions/extensions/transfer_hook/initialize.rs
@@ -7,19 +7,29 @@ use {
     solana_program_error::ProgramResult,
 };
 
-/// Initialize the Transfer Hook extension on a mint.
+/// Initialize a new mint with a transfer hook program.
 ///
-/// Expected accounts:
+/// Fails if the mint has already been initialized, so must be called before
+/// `InitializeMint`.
 ///
-/// 0. `[writable]` The mint account to initialize the Transfer Hook extension.
+/// The mint must have exactly enough space allocated for the base mint (82
+/// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+/// then space required for this extension, plus any others.
+///
+/// Accounts expected by this instruction:
+///
+///   0. `[writable]` The mint to initialize.
 pub struct InitializeTransferHook<'a, 'b> {
-    /// Mint Account to initialize.
-    pub mint_account: &'a AccountView,
-    /// Optional authority that can set the transfer hook program id
+    /// The token mint.
+    pub mint: &'a AccountView,
+
+    /// The address for the account that can update the program id.
     pub authority: Option<&'b Address>,
-    /// Program that authorizes the transfer
+
+    /// The program id that performs logic during transfers.
     pub program_id: Option<&'b Address>,
-    /// Token Program
+
+    /// The token program.
     pub token_program: &'b Address,
 }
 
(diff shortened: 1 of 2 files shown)


programs/token-2022/src/instructions/extensions/transfer_hook/initialize.rs
```diff
@@ -0,0 +1,65 @@
...
+    #[inline(always)]
+    pub fn invoke(&self) -> ProgramResult {
+        self.invoke_signed(&[])
+    }
+
+    #[inline(always)]
+    pub fn invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
+        let instruction_accounts = [InstructionAccount::writable(self.mint_account.address())];
+
+        let mut data = [0u8; 66];
```
febo: Let's use uninitialized bytes here – you can see an example [here](https://github.com/anza-xyz/pinocchio/blob/main/programs/token-2022/src/instructions/initialize_account_3.rs#L38).

programs/token-2022/src/instructions/extensions/transfer_hook/initialize.rs
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            data[2..34].copy_from_slice(&[0; 32]);
+        }
+
+        // Set program_id at offset [34..66]
+        if let Some(x) = self.program_id {
+            data[34..66].copy_from_slice(x.to_bytes().as_ref());
+        } else {
+            data[34..66].copy_from_slice(&[0; 32]);
+        }
```
febo: This is not quite right. Optional values are represented as `COption`, so there is a 4-bytes prefix to indicate whether the value is present or not. Each value is 36 bytes long:
* 4 bytes for option
* 32 bytes for address

programs/token-2022/src/instructions/extensions/transfer_hook/update.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        for (account_meta, signer) in instruction_accounts[2..]
+            .iter_mut()
+            .zip(account_signers.iter())
+        {
+            account_meta.write(InstructionAccount::readonly_signer(signer.address()));
+        }
+
+        let mut data = [0u8; 34];
```
febo: Same here, use uninitialiazed array.

febo (commented):
There are a few bits to fix.

programs/token-2022/src/instructions/extensions/transfer_hook/initialize.rs
```diff
@@ -0,0 +1,65 @@
...
+impl InitializeTransferHook<'_, '_> {
+    pub const DISCRIMINATOR: u8 = 0;
+
+    #[inline(always)]
+    pub fn invoke(&self) -> ProgramResult {
+        self.invoke_signed(&[])
+    }
+
+    #[inline(always)]
+    pub fn invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
```
febo: There are no signers in this instruction, no need for this one.

programs/token-2022/src/instructions/extensions/transfer_hook/update.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        // Account metadata
+        const UNINIT_INSTRUCTION_ACCOUNTS: MaybeUninit<InstructionAccount> =
+            MaybeUninit::<InstructionAccount>::uninit();
+        let mut instruction_accounts = [UNINIT_INSTRUCTION_ACCOUNTS; 2 + MAX_MULTISIG_SIGNERS];
+
+        unsafe {
+            // SAFETY:
+            // - `instruction_accounts` is sized to 2 + MAX_MULTISIG_SIGNERS
+            // - Index 0 is always present
```
febo: Move safety comment to the top of the `unsafe` block.

programs/token-2022/src/instructions/extensions/transfer_hook/update.rs
```diff
@@ -0,0 +1,128 @@
...
+        };
+
+        // Account info array
+        const UNINIT_INFO: MaybeUninit<&AccountView> = MaybeUninit::uninit();
+        let mut account_views = [UNINIT_INFO; 2 + MAX_MULTISIG_SIGNERS];
+
+        unsafe {
+            // SAFETY:
+            // - `account_views` is sized to 2 + MAX_MULTISIG_SIGNERS
+            // - Index 0 is always present
```
febo: Move safety comment to the top of the `unsafe` block.

programs/token-2022/src/instructions/extensions/transfer_hook/initialize.rs
```diff
@@ -0,0 +1,65 @@
...
+    solana_account_view::AccountView,
+    solana_address::Address,
+    solana_instruction_view::{
+        cpi::{invoke_signed, Signer},
+        InstructionAccount, InstructionView,
+    },
+    solana_program_error::ProgramResult,
+};
+
+pub struct InitializeTransferHook<'a, 'b> {
```
febo, line 22: We should add the comments for the instruction, like the ones in the other instructions.

burhankhaja:
@febo i have resolved all the suggestions in this extensions, similarly in other extension prs too - except for scaled-ui-amounts and default-account-state

programs/token-2022/src/instructions/extensions/transfer_hook/update.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        // Account metadata
+        const UNINIT_INSTRUCTION_ACCOUNTS: MaybeUninit<InstructionAccount> =
+            MaybeUninit::<InstructionAccount>::uninit();
+        let mut accounts = [UNINIT_INSTRUCTION_ACCOUNTS; 2 + MAX_MULTISIG_SIGNERS];
+
+        // SAFETY:
+        // - `accounts` is sized to 2 + MAX_MULTISIG_SIGNERS
+        // - Index 0 and 1 are always present
+
```
febo: nit:  This blank line can be removed.

programs/token-2022/src/instructions/extensions/transfer_hook/initialize.rs
febo: @burhankhaja Still missing the comments on the instructions.

programs/token-2022/src/instructions/extensions/transfer_hook/update.rs
burhankhaja: it is already there

programs/token-2022/src/instructions/extensions/transfer_hook/update.rs
burhankhaja: i think they are already outside the unsafe block

PR #282, opened 2025-11-18
system: Add create account helper

febo:
### Problem

Creating accounts funded by a payer account is a common operation. In most cases, the logic needs to be replicated in different programs.

### Solution

Add a helper to create accounts. The helper supports creating both PDA and non-PDA accounts.

### DIFF
--- a/programs/system/src/lib.rs
+++ b/programs/system/src/lib.rs
@@ -15,6 +15,18 @@ pub mod instructions;
 pinocchio_pubkey::declare_id!("11111111111111111111111111111111");
 
 /// Create an account with a minimum balance to be rent-exempt.
+#[inline(always)]
+pub fn create_account_with_minimum_balance(
+    account: &AccountInfo,
+    space: usize,
+    owner: &Pubkey,
+    payer: &AccountInfo,
+    rent_sysvar: Option<&AccountInfo>,
+) -> ProgramResult {
+    create_account_with_minimum_balance_signed(account, space, owner, payer, rent_sysvar, &[])
+}
+
+/// Create an account with a minimum balance to be rent-exempt.
 ///
 /// When creating a PDA `account`, the PDA signer seeds must be provided
 /// via the `signers`.
@@ -24,13 +36,13 @@ pinocchio_pubkey::declare_id!("11111111111111111111111111111111");
 /// owned by the system program and its signer seeds can be provided
 /// via the `signers`.
 #[inline(always)]
-pub fn create_account_with_minimum_balance(
+pub fn create_account_with_minimum_balance_signed(
     account: &AccountInfo,
     space: usize,
     owner: &Pubkey,
     payer: &AccountInfo,
-    signers: &[Signer],
     rent_sysvar: Option<&AccountInfo>,
+    signers: &[Signer],
 ) -> ProgramResult {
     let lamports = if let Some(rent_sysvar) = rent_sysvar {
         let rent = Rent::from_account_info(rent_sysvar)?;


programs/system/src/lib.rs
```diff
@@ -1,5 +1,71 @@
...
+        // Create the account if it does not exist.
+        CreateAccount {
+            from: payer,
+            to: account,
+            lamports,
+            space: space as u64,
+            owner,
+        }
+        .invoke_signed(signers)
+    } else {
```
BretasArthur1, line 64: Hey @febo! Given this `else` block, I imagine that the case for this would be something like "Account has lamports but needs everything else configured." So, if I'm correct, on the end of the block we always perform the `Assign` cpi call. So for example if I call the  `create_account_with_minimal_balance` function twice on an already configured account it always call `Assign` and call `resize_unchecked`. 

Shouldn't we then add a verification like 

```rust
if account.owner() != owner {
        Assign { account, owner }.invoke_signed(signers)?;
    }

     ...

if account.data_len() != space {
        unsafe { account.resize_unchecked(space)? };
    }
```

So in the case where we call twice we don't perform this operations :)

If the idea of this should only and exclusively be called on brand new or completely unconfigured accounts currently approach it's perfect! And please correct if I'm wrong about something!

febo: At the end of the day it is a trade-off since extra checks consume CUs. Calling it on an already "configured" account is most likely a bug – it would be easier for programs to check that before calling the `create_account_with_minimal_balance` helper.

BretasArthur1: Nice, sounds reasonable! Thanks. The code it's really clean.

programs/system/src/lib.rs
```diff
@@ -1,5 +1,71 @@
...
+        let required_lamports = lamports.saturating_sub(account.lamports());
+
+        // Transfer lamports from `payer` to `account` if needed.
+        if required_lamports > 0 {
+            Transfer {
+                from: payer,
+                to: account,
+                lamports: required_lamports,
+            }
+            .invoke()?;
```
grod220: Should we make this `.invoke_signed(signers)?;` to support the case a PDA acts as the funding account for the transfer?

programs/system/src/lib.rs
```diff
@@ -1,5 +1,71 @@
...
+    let lamports = if let Some(rent_sysvar) = rent_sysvar {
+        let rent = Rent::from_account_info(rent_sysvar)?;
+        rent.minimum_balance(space)
+    } else {
+        Rent::get()?.minimum_balance(space)
+    };
+
+    if account.lamports() == 0 {
+        // Create the account if it does not exist.
+        CreateAccount {
```
rustopian, line 56: Good stopgap. The whole `if`/`else` setup can be replaced with `CreateAccountAllowPrefund` once active

programs/system/src/lib.rs
```diff
@@ -1,5 +1,71 @@
...
+    rent_sysvar: Option<&AccountInfo>,
+) -> ProgramResult {
+    let lamports = if let Some(rent_sysvar) = rent_sysvar {
+        let rent = Rent::from_account_info(rent_sysvar)?;
+        rent.minimum_balance(space)
+    } else {
+        Rent::get()?.minimum_balance(space)
+    };
+
+    if account.lamports() == 0 {
```
sonicfromnewyoke, line 54: what do you think about using hints from `pinocchio::hint` module?

i expect account is not exist, in the majority of cases
```suggestion
    if likely(account.lamports() == 0) {
```

programs/system/src/lib.rs
```diff
@@ -1,5 +1,71 @@
...
+            lamports,
+            space: space as u64,
+            owner,
+        }
+        .invoke_signed(signers)
+    } else {
+        let required_lamports = lamports.saturating_sub(account.lamports());
+
+        // Transfer lamports from `payer` to `account` if needed.
+        if required_lamports > 0 {
```
sonicfromnewyoke, line 68: the same here can be applied, i believe.

programs/system/src/lib.rs
BretasArthur1: nice

febo: > what do you think about using hints from `pinocchio::hint` module?
> 
> i expect account is not exist, in the majority of cases

It made no difference in my tests. Also, I had a look at the asm and it seems the compiler is doing the right thing already. Did you notice a difference?

programs/system/src/lib.rs
febo: Same thing here, no difference.

programs/system/src/lib.rs
sonicfromnewyoke: the same for the lib (i didn't test it in the real app).
As a part of library it compiles in the expected instructions.
Just wanted  to make extra sure this logic won't be miscompiled in the large program.

programs/system/src/lib.rs
febo: Yeah, that is a good point. It would only work for PDAs used as signers (i.e., owned by the system program), but it is a good idea to support this scenario. 

Made the change on [1466eda](https://github.com/anza-xyz/pinocchio/pull/282/commits/1466edae6f381eb436c1c6e2b350a6fa2f16a95e) and updated the docs.

programs/system/src/lib.rs
febo: I wrote a small program and there was no difference. I think we can leave it as it is and revisit later if we see that there is a change in behaviour.

febo:
> LGTM

Something just occurred to me. What if we have a `create_account_with_minimum_balance` and a `create_account_with_minimum_balance_signed` variants? The first one does not take any signer seeds.

grod220:
Curious, what's the benefit of having a specific `_signed()` variant with the same parameters? Are there CU differences? Do consumers prefer control over signing for that transfer?

febo:
> Curious, what's the benefit of having a specific `_signed()` variant with the same parameters? Are there CU differences? Do consumers prefer control over signing for that transfer?

Just convenience – instead of writing:
```rust
create_account_with_minimum_balance(
   to,
   10,
   program_id,
   from,
   &[], 
   None
)
```
you will write:
```rust
create_account_with_minimum_balance(
   to,
   10,
   program_id,
   from,
   None
)
```
Then when you are signing with a PDA, you use:
```rust
create_account_with_minimum_balance_signed(
   to, 
   10,
   program_id,
   from,
   &[pda_seeds],
   None
)
```

grod220:
Ah! Indeed that scheme is consistent with what we do elsewhere. Supportive 👍

They can just re-use each other:

```rust
pub fn create_account_with_minimum_balance(
    to: &AccountInfo,
    space: usize,
    owner: &Pubkey,
    from: &AccountInfo,
    rent_sysvar: Option<&AccountInfo>,
) -> ProgramResult {
    create_account_with_minimum_balance_signed(
        to,
        space,
        owner,
        from,
        &[],
        rent_sysvar,
    )
}

pub fn create_account_with_minimum_balance_signed(
    to: &AccountInfo,
    space: usize,
    owner: &Pubkey,
    from: &AccountInfo,
    signers: &[Signer],
    rent_sysvar: Option<&AccountInfo>,
) -> ProgramResult
```

joncinque (commented):
Sorry for the late review, but looks great to me!

cavemanloverboy:
good work

PR #284, opened 2025-11-21
pinocchio-stake

L0STE:
A no-std efficient client to interact with the stake-program

### DIFF
--- a/programs/stake/src/instructions/authorize.rs
+++ b/programs/stake/src/instructions/authorize.rs
@@ -76,22 +76,22 @@ impl Authorize<'_, '_> {
         };
 
         // Instruction data
-        // -  [0]   : instruction discriminator (1 byte, u8)
-        // -  [1..33]: new_authority (32 bytes, Pubkey)
-        // -  [33]  : authority_type (1 byte, u8)
-        let mut instruction_data = [UNINIT_BYTE; 34];
+        // -  [0..4]  : instruction discriminator (4 bytes, u32)
+        // -  [4..36]: new_authority (32 bytes, Pubkey)
+        // -  [36]  : authority_type (1 byte, u8)
+        let mut instruction_data = [UNINIT_BYTE; 37];
 
-        // Set discriminator as u8 at offset [0]
-        write_bytes(&mut instruction_data, &[1]);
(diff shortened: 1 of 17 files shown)


programs/stake/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+
+pub mod instructions;
+pub mod state;
+
+pinocchio_pubkey::declare_id!("Stake11111111111111111111111111111111111111");
+
+use core::mem::MaybeUninit;
+use pinocchio::{account_info::AccountInfo, instruction::AccountMeta};
+
+const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
```
sonicfromnewyoke, line 11: this constant is already duplicated in `pinocchio-token`, `pinocchio--token-2022`,  `pinocchio-logger`, and `pinocchio` crates.

i guess we are ready to extract it into either `utils/helpers` or import from `pinocchio` directly.

no need to do it in the scope of this MR, I believe, and better to make it as a separate one.

L0STE: Really good point, but I have a counter argument to this:

Afaik Febo said that pinocchio should be as barebone as possible (meaning that probably this wouldn't fit there) and pinocchio should be the only import for any clients (meaning we can't put it anywhere else)

I don't mind it having it plastered across all clients, but at the same time having it in pinocchio would be great since it's always used across all clients

sonicfromnewyoke: makes sense.

anyways, i have already made a small refactoring MR to show how it can be used, and taking into account this crate already dependent on `pinocchio` we can achieve the next usage:

```diff
-use core::mem::MaybeUninit;
-use pinocchio::{account_info::AccountInfo, instruction::AccountMeta};
+use pinocchio::{account_info::AccountInfo, bytes::UNINIT_BYTE, instruction::AccountMeta};

-const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();

LStan:
`GetMinimumDelegation`, `Initialize`, `DeactivateDelinquent` don't have signers, so `invoke_signed` is not needed.

PR #285, opened 2025-11-21
refactor: extract byte operations into dedicated crate

sonicfromnewyoke:
### Problem

there is a part of functionality that is copy/pasted across Pinocchio crates

### Summary of Changes

- crate `bytes` module under `pinnochio` main crate
- extract 
```rust
const UNINIT_BYTE: core::mem::MaybeUninit<u8> // ...

fn write_bytes(destination: &mut [core::mem::MaybeUninit<u8>], source: &[u8]) {
  // ...
}
``` 
into this module
- import this functionality in `token` and `token-2022` crates instead of having duplicated logic

also will be used in one more crate - #284

cc @febo

### DIFF
--- a/programs/token-2022/src/instructions/approve.rs
+++ b/programs/token-2022/src/instructions/approve.rs
@@ -2,13 +2,12 @@ use core::slice::from_raw_parts;
 
 use pinocchio::{
     account_info::AccountInfo,
+    bytes::{write_bytes, UNINIT_BYTE},
     instruction::{AccountMeta, Instruction, Signer},
     program::invoke_signed,
     pubkey::Pubkey,
     ProgramResult,
 };
-
-use crate::{write_bytes, UNINIT_BYTE};
 
 /// Approves a delegate.
 ///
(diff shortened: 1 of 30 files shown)


Aursen:
It's nice to have it right there in the Pinocchio crate

Aursen:
Currently, it should not be in the main crate. Rather, it should be in another crate

sonicfromnewyoke:
> Currently, it should not be in the main crate. Rather, it should be in another crate

im thinking maybe it worth to get rid of those helpers at all and use `core::ptr::copy_nonoverlapping`

Aursen:
> > Currently, it should not be in the main crate. Rather, it should be in another crate
> 
> im thinking maybe it worth to get rid of those helpers at all and use `core::ptr::copy_nonoverlapping`

Yeah make sense. I don't know the opinion of @febo on it

febo:
Closing this as we should be using `core::ptr::copy_nonoverlapping` instead.

PR #286, opened 2025-11-22
ci: Use install action

febo:
### Problem

`taiki-e/cache-cargo-install-action` requires building the tools in CI, which could fail when new updates are published and the requirements change – e.g., the rust version of the crate is higher than the version used in the workspace.

### Solution

Update CI to use `taiki-e/install-action`, which install the binary instead.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -69,27 +69,27 @@ runs:
 
     - name: Install 'cargo-audit'
       if: ${{ contains(inputs.components, 'audit') }}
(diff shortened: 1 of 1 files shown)


febo:
> LGTM
> 
> Also, not sure if relevant for this repo, but an increasing amount of repos are moving to a Makefile workflow that is used with [solana-program/actions](https://github.com/solana-program/actions) in CI/CD ([stake program example](https://github.com/solana-program/stake/blob/main/.github/workflows/main.yml)). May help simplify the maintenance of these actions.

Yeah, we should switch to use a makefile.

PR #287, opened 2025-11-24
Add Compute Budget Program CPI helpers

tanmay4l:
## Problem
Pinocchio lacks CPI helpers for the Compute Budget program. Developers must manually construct instruction bytes with discriminators, which is error-prone.

this provides the same convenience for priority fees that `pinocchio-system` provides for transfers.
cc @febo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,6 +48,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "pinocchio-compute-budget"
+version = "0.1.0"
+dependencies = [
+ "pinocchio",
+ "pinocchio-pubkey",
+]
+
+[[package]]
 name = "pinocchio-log"
 version = "0.5.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 resolver = "2"
 members = [
     "programs/associated-token-account",
+    "programs/compute-budget",
     "programs/memo",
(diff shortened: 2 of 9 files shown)


tanmay4l:
> i afraid those instructions should be used on the top level of txn but not via CPI

oh so Compute budget = transaction-level, not CPI got it thanks

sonicfromnewyoke:
> > i afraid those instructions should be used on the top level of txn but not via CPI
> 
> oh so Compute budget = transaction-level, not CPI got it thanks

if i remember correctly, [fn check_transactions() {}](https://github.com/anza-xyz/agave/blob/master/runtime/src/bank/check_transactions.rs#L53) checks budget that was computed and put into `compute_budget_instruction_details` of `TransactionMeta` **before** the execution. So, that's why i think CPI call won't give any effect (or even doesn't work)

tanmay4l:
> > > i afraid those instructions should be used on the top level of txn but not via CPI
> > 
> > 
> > oh so Compute budget = transaction-level, not CPI got it thanks
> 
> if i remember correctly, [fn check_transactions() {}](https://github.com/anza-xyz/agave/blob/master/runtime/src/bank/check_transactions.rs#L53) checks budget that was computed and put into `compute_budget_instruction_details` of `TransactionMeta` **before** the execution. So, that's why i think CPI call won't dive any effect (or even doesn't work)

got it

PR #290, opened 2025-12-02
v0.9.x: Use `sol_get_sysvar` in sysvars

febo:
### Problem

There are two ways to retrieve sysvar data. Many sysvars have an individual syscall (e.g., `Rent` has a `sol_get_rent_sysvar`) and there is also a generic `sol_get_sysvar`. Apart from the base syscall base cost, the genetic syscall charges`max(size_of::<T>() / 250, 10)`, while sysvar-specific ones charges `size_of::<T>()`. For any sysvar longer than 10 bytes, the generic syscall is cheaper in terms of compute units.

### Solution

Add an implementation of `Sysvar::get` that uses `sol_get_sysvar` and updates the sysvars. As a result, instructions that use `Rent::get` are `~9` CUs cheaper.

### DIFF
--- a/sdk/pinocchio/src/sysvars/fees.rs
+++ b/sdk/pinocchio/src/sysvars/fees.rs
@@ -1,7 +1,12 @@
 //! Calculation of transaction fees.
 
 use super::{clock::DEFAULT_MS_PER_SLOT, Sysvar};
-use crate::impl_sysvar_get;
+use crate::{impl_sysvar_get, program_error::ProgramError, pubkey::Pubkey};
+
+pub const FEES_ID: Pubkey = [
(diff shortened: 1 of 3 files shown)


sdk/pinocchio/src/sysvars/fees.rs
```diff
@@ -93,5 +98,5 @@ impl Fees {
 }
 
 impl Sysvar for Fees {
-    impl_sysvar_get!(sol_get_fees_sysvar);
+    impl_sysvar_get!(FEES_ID, 7);
```
LStan, line 101: Why not 0?

febo: The `FeeRateGovernor` struct, which is part of the `Fees` sysvar, has an `u8` field and therefore the struct has 7 bytes padding at the end.

PR #294, opened 2025-12-05
sdk: Refactor `Rent`sysvar

febo:
### Problem

The concept of rent no longer exists, only rent-exemption, and upstream BPF does not support `f64` values but currently the `Rent` sysvar implementation uses them.

### Solution

Refactor the sysvar implementation and remove all references to `f64`.

### DIFF
--- a/sdk/src/sysvars/rent.rs
+++ b/sdk/src/sysvars/rent.rs
@@ -260,6 +260,9 @@ impl Rent {
     /// Returns `ProgramError::InvalidArgument` if `data_len` exceeds the maximum
     /// permitted data length or if the `lamports_per_byte` is too large based on
     /// the `exemption_threshold`, which would cause an overflow.
+    //
+    // Note: Clippy suggests collapsing the `if` statements, but they are kept
+    // separate since it is more CU-efficient this way.
     #[allow(clippy::collapsible_if)]
     #[inline(always)]
     pub fn try_minimum_balance(&self, data_len: usize) -> Result<u64, ProgramError> {


sdk/src/sysvars/rent.rs
```diff
@@ -195,22 +176,26 @@ impl Rent {
...
             _ => {
-                (((ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte_year) as f64
-                    * self.exemption_threshold) as u64
+                #[cfg(any(target_os = "solana", not(target_arch = "bpf")))]
+                {
+                    (((ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte) as f64
+                        * f64::from_le_bytes(self.exemption_threshold)) as u64
+                }
+                #[cfg(target_arch = "bpf")]
+                panic!("Floating-point operations are not supported on BPF targets");
```
febo: I am not sure about this, but there is not much we can do if the `exemption_threshold` value does not match one of the known ones.

sdk/src/sysvars/rent.rs
```diff
@@ -28,16 +28,27 @@ pub const RENT_ID: Address = Address::new_from_array([
...
 pub const DEFAULT_LAMPORTS_PER_BYTE_YEAR: u64 = 1_000_000_000 / 100 * 365 / (1024 * 1024);
 
-/// SIMD-0194
+/// Default rental rate in lamports/byte.
 ///
-/// This equates to the integer value of 3480. To account for 2 years of rent
-/// exemption, we multiply this value by 2 to make it 6960.
+/// This calculation is based on:
+/// - `10^9` lamports per SOL
+/// - `$1` per SOL
```
joncinque, line 47: I hope you appreciated that I *had* to keep this comment around, it's pretty legendary 😁

sdk/src/sysvars/rent.rs
```diff
@@ -195,22 +176,26 @@ impl Rent {
...
             }
             F64_EXEMPTION_THRESHOLD_AS_U64 => {
-                ((ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte_year)
+                ((ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte)
                     * DEFAULT_EXEMPTION_THRESHOLD_AS_U64
             }
             _ => {
-                (((ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte_year) as f64
-                    * self.exemption_threshold) as u64
+                #[cfg(any(target_os = "solana", not(target_arch = "bpf")))]
```
joncinque: Maybe I'm being dumb, but why isn't this just `#[cfg(not(target_arch = "bpf"))]`?

sdk/src/sysvars/rent.rs
joncinque: I don't think we really have an alternative unfortunately, this is probably the best option

sdk/src/sysvars/rent.rs
```diff
@@ -302,24 +259,24 @@ mod tests {
     #[test]
     pub fn test_minimum_balance_simd0194() {
         let mut rent = super::Rent {
-            lamports_per_byte_year: DEFAULT_LAMPORTS_PER_BYTE,
-            exemption_threshold: 1.0, // SIMD-0194 default
+            lamports_per_byte: DEFAULT_LAMPORTS_PER_BYTE,
+            exemption_threshold: 1.0f64.to_le_bytes(), // SIMD-0194 default
```
joncinque: nit: how about using `F64_SIMD0194_EXEMPTION_THRESHOLD_AS_U64.to_le_bytes()` instead?

joncinque (commented):
Just one question about the `target_os = solana` part, the rest looks great!

sdk/src/sysvars/rent.rs
febo: Followed your approach in [#487](https://github.com/anza-xyz/solana-sdk/pull/487/files) and used a `[u8; 8]` directly.

sdk/src/sysvars/rent.rs
febo: You are right – fixed!

sdk/src/sysvars/rent.rs
febo: Definitely! 😀

sdk/src/sysvars/rent.rs
```diff
@@ -178,18 +174,16 @@ impl Rent {
     #[inline]
     pub fn minimum_balance(&self, data_len: usize) -> u64 {
         let bytes = data_len as u64;
-        let exemption_threshold_as_u64 = u64::from_le_bytes(self.exemption_threshold);
 
-        match exemption_threshold_as_u64 {
-            F64_SIMD0194_EXEMPTION_THRESHOLD_AS_U64 => {
+        match self.exemption_threshold {
+            SIMD0194_EXEMPTION_THRESHOLD => {
                 (ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte
```
joncinque: I know this code has been here for some time, but I'm just realizing that we now have an opportunity to overflow this calculation, which wasn't the case when it used floats. Should we eventually use some saturating math here to be safe?

sdk/src/sysvars/rent.rs
febo: @joncinque Reverted to use `u64` values since it is more efficient in terms of CUs - there was a `12` CUs difference in a small test.

sdk/src/sysvars/rent.rs
febo: This is a tricky one. Adding saturated math increased CUs by `~70`, which is about `50%`. It will probably be cheaper to validate the `data_len` and `lamports_per_byte` values so that we can determine that it won't overflow.

sdk/src/sysvars/rent.rs
```diff
@@ -28,38 +28,52 @@ pub const RENT_ID: Address = Address::new_from_array([
...
-/// This equates to the integer value of 3480. To account for 2 years of rent
-/// exemption, we multiply this value by 2 to make it 6960.
+/// This calculation is based on:
+/// - `10^9` lamports per SOL
+/// - `$1` per SOL
+/// - `$0.01` per megabyte day
+/// - `$7.30` per megabyte
 pub const DEFAULT_LAMPORTS_PER_BYTE: u64 = 6960;
 
+#[cfg(not(target_arch = "bpf"))]
```
febo, line 52: We need to "hide" this `f64` value for upstream BPF.

sdk/src/sysvars/rent.rs
joncinque: Ah ok, let's make the same in change in the sdk then

sdk/src/sysvars/rent.rs
joncinque: Wow, I didn't think it would be so bad! Adding checks makes sense to me, and if it's costly, we could *only* check `data_len` and add a warning specifying the range of valid values for `lamports_per_byte`

sdk/src/sysvars/rent.rs
```diff
@@ -73,23 +87,18 @@ pub const ACCOUNT_STORAGE_OVERHEAD: u64 = 128;
...
-    /// SIMD-0194 will deprecate this value. The current implementation checks
-    /// the value against known defaults to avoid performing a floating-point
-    /// operation on-chain.
-    #[deprecated(since = "0.9.2", note = "SIMD-0194 will deprecate this value")]
-    pub exemption_threshold: f64,
-
-    /// Burn percentage
-    pub burn_percent: u8,
+    /// The concept of rent no longer exists, only rent-exemption.
+    exemption_threshold: [u8; 8],
```
joncinque, line 108: Should we just store this as a u64 to make things simpler?

febo: We can everything as a `[u8; 8]` if we use an `if` statements instead of a `match` – no difference in CUs (https://github.com/anza-xyz/pinocchio/pull/294/commits/eb6d8da730d37c1d63d108466384cf5832b82089).

sdk/src/sysvars/rent.rs
febo: Adding a test on both `data_len` and `lamports_per_byte` takes `6` extra CUs. The caveat is that currently `minimum_exempt` is not a fallible function, so either we would need to panic or change the return type.

We could probably have a `minimum_exempt` and an unsafe `minimim_exempt_unchecked` (without the validation).

joncinque: That makes sense to me -- performance is getting huge gains anyway from not using floats, so the extra checks are not such a big deal

sdk/src/sysvars/rent.rs
joncinque: Ah interesting, good to know! Do we know if that's a problem with codegen? It seems like the two should be equivalent, no?

sdk/src/sysvars/rent.rs
febo: Added a fallible `try_minimum_balance` and unsafe `minimum_balance_unchecked`; also deprecated the `minimum_balance`, which now would panic if `data_len` exceeds the maximum allowed.

It is a bit tricky to have a good check for `lamports_per_byte`, since it depends on both `exempt_threshold` and maximum `data_len`. I was thinking that we can assume that the runtime guarantees the value of `lamports_per_byte` is valid in relation to the maximum `data_len`.

sdk/src/sysvars/rent.rs
febo: I checked this and the compiler/llvm won't be able to infer all the information to optimize when used in a match statement, so this is the expected behaviour.

sdk/src/sysvars/rent.rs
```diff
@@ -174,27 +177,88 @@ impl Rent {
...
+    /// Returns `ProgramError::InvalidArgument` if `data_len` exceeds the maximum permitted
+    /// data length.
+    #[inline(always)]
+    pub fn try_minimum_balance(&self, data_len: usize) -> Result<u64, ProgramError> {
+        if data_len as u64 > MAX_PERMITTED_DATA_LENGTH {
+            Err(ProgramError::InvalidArgument)
+        } else {
+            // SAFETY: The `data_len` is validated to be lower than the maximum permitted
+            // data length.
+            Ok(unsafe { self.minimum_balance_unchecked(data_len) })
```
joncinque: Do we want to worry about the case where `lamports_per_byte` is huge? Because this can still overflow

sdk/src/sysvars/rent.rs
```diff
@@ -21,45 +24,62 @@ pub const RENT_ID: Address = Address::new_from_array([
...
+    note = "The concept of rent no longer exists, only rent-exemption"
+)]
 pub const DEFAULT_EXEMPTION_THRESHOLD: f64 = 2.0;
 
-/// Default amount of time (in years) the balance has to include rent for the
-/// account to be rent exempt as a `u64`.
-const DEFAULT_EXEMPTION_THRESHOLD_AS_U64: u64 = 2;
-
-/// The `u64` representation of the default exemption threshold.
+/// The `u64::to_le_bytes` representation of the default exemption threshold.
```
joncinque: nit
```suggestion
/// The `f64::to_le_bytes` representation of the default exemption threshold.
```

sdk/src/sysvars/rent.rs
```diff
@@ -21,45 +24,62 @@ pub const RENT_ID: Address = Address::new_from_array([
...
 ///
-/// This is used to check whether the `f64` value can be safely cast to a `u64`.
-const F64_EXEMPTION_THRESHOLD_AS_U64: u64 = 4611686018427387904;
+/// This value is equivalent to `2.0f64`. It is only used to check whether
+/// the exemption threshold is the default value to avoid performing
+/// floating-point operations on-chain.
+const CURRENT_EXEMPTION_THRESHOLD: [u8; 8] = [0, 0, 0, 0, 0, 0, 0, 64];
 
-/// The `u64` representation of the deprecated exemption threshold.
+/// The `u64::to_le_bytes` representation of the SIMD-0194 exemption threshold.
```
joncinque: nit
```suggestion
/// The `f64::to_le_bytes` representation of the SIMD-0194 exemption threshold.
```

sdk/src/sysvars/rent.rs
```diff
@@ -153,35 +165,75 @@ impl Rent {
...
+    ///
+    /// * `data_len` - The number of bytes in the account
+    ///
+    /// # Returns
+    ///
+    /// The minimum balance in lamports for rent exemption.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `data_len` exceeds the maximum permitted data length.
```
joncinque: Wanna add "or if `lamports_per_byte` is greater than `1759218604441`"? That'll cause an overflow with `MAX_PERMITTED_DATA_SIZE` at 10 * 1024 * 1024 and exemption threshold at 1.

sdk/src/sysvars/rent.rs
```diff
@@ -153,35 +165,75 @@ impl Rent {
...
+    /// * `data_len` - The number of bytes in the account
+    ///
+    /// # Returns
+    ///
+    /// The minimum balance in lamports for rent exemption.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `data_len` is within permitted the permitted
+    /// limit to avoid overflow.
```
joncinque: Same here, wanna add something about `lamports_per_byte`?
```suggestion
    /// The caller must ensure that `data_len` and `lamports_per_byte` are within the permitted
    /// limits to avoid overflow.
```

sdk/src/sysvars/rent.rs
febo: Yeah, I mentioned on a comment above:

_It is a bit tricky to have a good check for lamports_per_byte, since it depends on both exempt_threshold and maximum data_len. I was thinking that we can assume that the runtime guarantees the value of lamports_per_byte is valid in relation to the maximum data_len._

sdk/src/sysvars/rent.rs
febo: > Wanna add "or if `lamports_per_byte` is greater than `1759218604441`"? That'll cause an overflow with `MAX_PERMITTED_DATA_SIZE` at 10 * 1024 * 1024 and exemption threshold at 1.

Should be `1_759_197_129_867` right? We need to add `128` bytes to the `MAX_PERMITTED_DATA_SIZE` to account for the account storage overhead.

febo: But actually we are not panicking in this case, since we are not validating the value. There are multiple limits and it depends on the `exempt_threshold` – e.g.:
* `1.0`: limit is `1_759_197_129_867`
* `2.0`: limit is `879_598_564_933`

It also depends on what the maximum `data_len` is.

sdk/src/sysvars/rent.rs
febo: Similar comment here. Perhaps we can assume that the runtime is set with the correct combination of `exempt_threshold`, `lamports_per_byte` and maximum data len limit. So the only "variable" that the client needs to be careful is the `data_len`.

sdk/src/sysvars/rent.rs
```diff
@@ -153,35 +165,75 @@ impl Rent {
...
+    #[inline(always)]
+    pub unsafe fn minimum_balance_unchecked(&self, data_len: usize) -> u64 {
```
joncinque: Is this actually unsafe? If not, I'd say `unchecked` makes it clear enough and remove `unsafe`

sdk/src/sysvars/rent.rs
joncinque: I thought about this some more, and if we're going to have a `try_` version and an `_unchecked` version, let's just have the `try_` version use checked math. We can't go wrong in that case, since people can always circumvent it if they want with `_unchecked`.

What do you think?

sdk/src/sysvars/rent.rs
joncinque: Ah you're right, I forgot about the 128 bytes, sorry about that.

joncinque (commented):
A few more ideas, let me know what you think! Maybe I'm too focused on silly usage of `Rent`, but it doesn't feel good to provide a `try_` function in an SDK that could overflow

sdk/src/sysvars/rent.rs
febo: We can add the `if` statements for each "known" value of the exempt threshold. For the float fallback there is no way to do the checked math, so we can leave as it is. The `if` statements will probably be cheaper than the checked math.

sdk/src/sysvars/rent.rs
febo: Good point, we should remove the `unsafe`.

sdk/src/sysvars/rent.rs
joncinque: Do you mean for known values of the `lamports_per_byte`? If so, that works for me!

sdk/src/sysvars/rent.rs
```diff
@@ -194,25 +252,32 @@ impl Rent {
...
-    #[allow(deprecated)]
-    pub fn minimum_balance(&self, data_len: usize) -> u64 {
-        let bytes = data_len as u64;
-        let exemption_threshold_as_u64 = u64::from_le_bytes(self.exemption_threshold.to_le_bytes());
+    ///
+    /// # Errors
+    ///
+    /// Returns `ProgramError::InvalidArgument` if `data_len` exceeds the maximum permitted
+    /// data length.
+    #[allow(clippy::collapsible_if)]
```
febo, line 266: Clippy complains about the `if` statement chain, but it was the most CU-wise efficient way to write it.

joncinque: Can you add that as a comment? Otherwise I can see someone removing this and changing the code, without realizing that there was a reason for it

joncinque (commented):
One last thing, and we can land this!

joncinque (approved):
Looks great, thanks for all the revisions!

PR #295, opened 2025-12-06
sdk: Improve IntrospectedInstruction

deanmlittle:
Made several improvements to the `IntrospectedInstruction` API, including:

1. Adding `num_account_metas(&self) -> usize` helper function
2. Created a private `new_unchecked(raw: *const u8) -> IntrospectedInstruction` function that is consumed internally to avoid having to expose `PhantomData` as a public field
3. Made `num_instructions` refer `usize` and removed all inline usize casts

### DIFF
--- a/sdk/src/sysvars/instructions.rs
+++ b/sdk/src/sysvars/instructions.rs
@@ -134,7 +134,7 @@ pub struct IntrospectedInstruction<'a> {
 }
 
 impl IntrospectedInstruction<'_> {
-    /// Create a new IntrospectedInstruction.
+    /// Create a new `IntrospectedInstruction`.
     ///
     /// # Safety
     ///


sdk/src/sysvars/instructions.rs
```diff
@@ -180,24 +201,23 @@ impl IntrospectedInstruction<'_> {
...
     }
 
     /// Get the instruction data of the `Instruction`.
     #[inline(always)]
     pub fn get_instruction_data(&self) -> &[u8] {
         // SAFETY: The first 2 bytes represent the number of accounts in the instruction.
-        let offset = u16::from_le_bytes(unsafe { *(self.raw as *const [u8; 2]) }) as usize
-            * size_of::<IntrospectedInstructionAccount>()
-            + ADDRESS_BYTES;
+        let offset =
```
febo, line 219: This bit in particular looks much nicer!

sdk/src/sysvars/instructions.rs
```diff
@@ -133,10 +130,34 @@ impl<'a> TryFrom<&'a AccountView> for Instructions<Ref<'a, [u8]>> {
...
+    /// who are sure that the index is in bounds, we have exposed it as an unsafe function.
+    #[inline(always)]
+    unsafe fn new_unchecked(raw: *const u8) -> Self {
+        Self {
+            raw,
+            marker: PhantomData,
+        }
+    }
+
+    /// Load the number of instructions in the currently executing `Transaction`.
```
febo: Here it should be `Instruction` right? Something like:
```suggestion
    /// Get the number of accounts of the `Instruction`.
```

sdk/src/sysvars/instructions.rs
```diff
@@ -133,10 +130,34 @@ impl<'a> TryFrom<&'a AccountView> for Instructions<Ref<'a, [u8]>> {
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct IntrospectedInstruction<'a> {
     pub raw: *const u8,
-    pub marker: PhantomData<&'a [u8]>,
+    marker: PhantomData<&'a [u8]>,
 }
 
 impl IntrospectedInstruction<'_> {
+    /// Create a new IntrospectedInstruction.
```
febo: nit: The spellcheck probably will complain if we don't put between quotes.
```suggestion
    /// Create a new `IntrospectedInstruction`.
```

sdk/src/sysvars/instructions.rs
```diff
@@ -133,10 +130,34 @@ impl<'a> TryFrom<&'a AccountView> for Instructions<Ref<'a, [u8]>> {
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct IntrospectedInstruction<'a> {
     pub raw: *const u8,
-    pub marker: PhantomData<&'a [u8]>,
+    marker: PhantomData<&'a [u8]>,
```
febo, line 133: Nice!

sdk/src/sysvars/instructions.rs
```diff
@@ -133,10 +130,34 @@ impl<'a> TryFrom<&'a AccountView> for Instructions<Ref<'a, [u8]>> {
...
 impl IntrospectedInstruction<'_> {
+    /// Create a new `IntrospectedInstruction`.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe because it does not verify anything about the pointer.
+    ///
+    /// It is private and used internally within the `get_instruction_account_at` function, which
+    /// performs the necessary index verification. However, to optimize performance for users
+    /// who are sure that the index is in bounds, we have exposed it as an unsafe function.
```
LStan, line 145: I guess, this comment was copied from `get_instruction_account_at_unchecked` and should not be here

PR #296, opened 2025-12-07
sdk: Use `sol_get_sysvar` for sysvars

febo:
### Problem

As described in #290,. `sol_get_sysvar` is more CU-efficient compared to using the individual sysvar syscalls.

### Solution

Add the implementation based on `sol_get_sysvar` from #290.

### DIFF
--- a/sdk/src/sysvars/rent.rs
+++ b/sdk/src/sysvars/rent.rs
@@ -108,18 +108,17 @@ pub struct Rent {
     exemption_threshold: [u8; 8],
     // Burn percentage.
     //
-    // The concept of rent no longer exists, only rent-exemption. The
-    // field is not omitted to preserve the struct 8-byte layout.
+    // The concept of rent no longer exists, only rent-exemption.
     //
     // The field (`burn_percent: u8`) is omitted to maintain the struct
     // aligned to 8-bytes, which improves the performance of loading the
     // sysvar.
 }
 
-// Assert that the size of the `Rent` struct is as expected (17 bytes).
+// Assert that the size of the `Rent` struct is as expected (16 bytes).
 const _ASSERT_STRUCT_LEN: () = assert!(size_of::<Rent>() == 16);
 
-// Assert that the alignment of the `Rent` struct is as expected (1 byte).
+// Assert that the alignment of the `Rent` struct is as expected (8 byte).
 const _ASSERT_ACCOUNT_ALIGN: () = assert!(align_of::<Rent>() == 8);
 
 impl Rent {


sdk/src/sysvars/mod.rs
```diff
@@ -65,6 +63,56 @@ macro_rules! impl_sysvar_get {
...
+                    Ok(unsafe { var.assume_init() })
+                }
+                $crate::sysvars::OFFSET_LENGTH_EXCEEDS_SYSVAR => {
+                    Err($crate::error::ProgramError::InvalidArgument)
+                }
+                $crate::sysvars::SYSVAR_NOT_FOUND => {
+                    Err($crate::error::ProgramError::UnsupportedSysvar)
+                }
+                // Unexpected errors are folded into `UnsupportedSysvar`.
+                _ => Err($crate::error::ProgramError::UnsupportedSysvar),
```
nagisa, line 113: Thought: If these are the only failure modes, then it might be worthwhile to have `get` return a get-specific error code (that possibly has a `impl From<SysvarGetError> for ProgramError` so that `?` continues working.)

sdk/src/sysvars/fees.rs
```diff
@@ -96,5 +101,5 @@ impl Fees {
 }
 
 impl Sysvar for Fees {
-    impl_sysvar_get!(sol_get_fees_sysvar);
+    impl_sysvar_get!(FEES_ID, 7);
```
joncinque: The fees sysvar is not supported by `sol_get_sysvar`, so this should stay as it is: https://github.com/anza-xyz/agave/blob/8fe17e2a7cecbbc59a5954510d2abdf179fb6700/program-runtime/src/sysvar_cache.rs#L109

joncinque (commented):
Looks good to me, just the comment on the fees sysvar

sdk/src/sysvars/fees.rs
febo: Done in https://github.com/anza-xyz/pinocchio/pull/296/commits/41c52fcbff5f710755a48a64102a49f2ee82885e.

sdk/src/sysvars/mod.rs
febo: @joncinque The change of alignment of the `Rent` struct (from `8` to `1`) makes the code slightly inefficient since the generated code has to perform an extra copy to align the return of the `Sysvar::get()`: `ProgramError` is 4-bytes aligned, while `Rent` is 1.

Creating a different `SysvarError` aligned to `1` improves a bit (2 CUs) and we can incluce an `impl From<SysvarError> for ProgramError` as @nagisa suggested. There is a bit of inefficiency on the conversion, so it is not the same as the ideal case.

Another alternative that offers the ideal improvement is to not include the `burn_percent` in the struct, which is not used in any of the code, and align the `Rent` struct to 8-bytes again:
```rust
pub struct Rent {
    /// Rental rate in lamports per byte.
    lamports_per_byte: u64,

    /// Exemption threshold in years.
    ///
    /// The concept of rent no longer exists, only rent-exemption.
    exemption_threshold: [u8; 8],

    // Burn percentage.
    //
    // The concept of rent no longer exists, only rent-exemption. The
    // field is not omitted to preserve the struct 8-byte layout.
    //omitted_burn_percent: u8,
}
```
This prevents any extra copy and provides the best performance – an extra gain of `8` CUs.

So the options are:
1. Leave as it is.
2. Create a custom `SysvarError` aligned to `1`.
3. Omit `burn_percent` from the struct.

sdk/src/sysvars/fees.rs
```diff
@@ -1,7 +1,12 @@
 //! Calculation of transaction fees.
 
 use super::{clock::DEFAULT_MS_PER_SLOT, Sysvar};
-use crate::impl_sysvar_get;
+use crate::{impl_sysvar_get, Address};
+
+pub const FEES_ID: Address = Address::new_from_array([
+    6, 167, 213, 23, 24, 226, 90, 141, 131, 80, 60, 37, 26, 122, 240, 113, 38, 253, 114, 0, 223,
+    111, 196, 237, 82, 106, 156, 144, 0, 0, 0, 0,
+]);
```
joncinque: Is this still needed?

sdk/src/sysvars/mod.rs
joncinque: Omitting `burn_percent` is a neat idea, but let's be sure -- do we want to have separate types between Pinocchio and the SDK?

If we're ok with _not_ using the same type, then we can go with a separate program-only version of `Rent` that's more CU-efficient. Given the benefits, I'm ok with a pinocchio-only `Rent` type.

sdk/src/sysvars/fees.rs
febo: Nope, removed.

sdk/src/sysvars/rent.rs
```diff
@@ -100,19 +100,28 @@ const CURRENT_MAX_LAMPORTS_PER_BYTE: u64 = 879_598_564_933;
...
+    // The field (`burn_percent: u8`) is omitted to maintain the struct
+    // aligned to 8-bytes, which improves the performance of loading the
+    // sysvar.
 }
 
+// Assert that the size of the `Rent` struct is as expected (17 bytes).
+const _ASSERT_STRUCT_LEN: () = assert!(size_of::<Rent>() == 16);
+
+// Assert that the alignment of the `Rent` struct is as expected (1 byte).
+const _ASSERT_ACCOUNT_ALIGN: () = assert!(align_of::<Rent>() == 8);
```
joncinque: nit: the comments are no longer correct
```suggestion
// Assert that the size of the `Rent` struct is as expected (16 bytes).
const _ASSERT_STRUCT_LEN: () = assert!(size_of::<Rent>() == 16);

// Assert that the alignment of the `Rent` struct is as expected (8 bytes).
const _ASSERT_ACCOUNT_ALIGN: () = assert!(align_of::<Rent>() == 8);
```

sdk/src/sysvars/rent.rs
```diff
@@ -100,19 +100,28 @@ const CURRENT_MAX_LAMPORTS_PER_BYTE: u64 = 879_598_564_933;
...
-    /// The concept of rent no longer exists, only rent-exemption.
-    burn_percent: u8,
+    // Burn percentage.
+    //
+    // The concept of rent no longer exists, only rent-exemption. The
+    // field is not omitted to preserve the struct 8-byte layout.
+    //
+    // The field (`burn_percent: u8`) is omitted to maintain the struct
+    // aligned to 8-bytes, which improves the performance of loading the
+    // sysvar.
```
LStan, line 115: There is something wrong with the comment: the field is not omitted and omitted simultaneously.

Issue #142, opened 2025-04-23
`HEAP_LENGTH` is the minimum length


publicqi:
```rust
pub const HEAP_LENGTH: usize = 32 * 1024;
```

currently `HEAP_LENGTH` is 0x8000 which is the minimum and default heap length. Maximum can go up to 0x40000.

https://github.com/anza-xyz/agave/blob/de672d2a81175a866e3a947b7dbede21fe2c054d/program-runtime/src/execution_budget.rs#L24-L25

we should release the limit. also for bump allocator, instead of allocating from the end of the heap, we can grow increasingly. when there's no enough space to use, an runtime access violation will be thrown and execution will halt.

we should provide user a macro to specify heap length. the downside of setting it directly to max is that the compiler cannot generate compile-time optimization, so giving the user power to change is a better solution.

@febo since you're cleaning up the allocator can you take this one?

febo:
Fixed in #301

PR #298, opened 2025-12-11
Make heap size configurable and bump allocator grow forward

rajgoesout:
Closes https://github.com/anza-xyz/pinocchio/issues/142

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -24,8 +24,24 @@ pub use alloc::BumpAllocator;
 /// Start address of the memory region used for program heap.
 pub const HEAP_START_ADDRESS: u64 = 0x300000000;
 
-/// Length of the heap memory region used for program heap.
-pub const HEAP_LENGTH: usize = 32 * 1024;
+/// Minimum heap length supported by the runtime.
+pub const MIN_HEAP_LENGTH: usize = 0x8000;
+
+/// Maximum heap length supported by the runtime.
+pub const MAX_HEAP_LENGTH: usize = 0x40000;
+
+/// Default heap length used by the runtime.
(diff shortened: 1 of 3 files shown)


sdk/src/entrypoint/mod.rs
```diff
@@ -623,15 +646,18 @@ mod alloc {
 
             let mut pos = *pos_ptr;
             if pos == 0 {
-                // First time, set starting position.
-                pos = self.start + self.len;
+                // First allocation starts after the cursor word.
+                pos = self.start + size_of::<usize>();
             }
-            pos = pos.saturating_sub(layout.size());
+            pos = pos.wrapping_add(layout.align().wrapping_sub(1));
```
febo, line 652: It is a bit strange to use `wrapping` operations here.

sdk/src/entrypoint/mod.rs
```diff
@@ -493,11 +509,15 @@ macro_rules! nostd_panic_handler {
 #[macro_export]
 macro_rules! default_allocator {
     () => {
+        $crate::default_allocator!($crate::entrypoint::DEFAULT_HEAP_LENGTH);
+    };
+    ( $len:expr ) => {
```
febo, line 514: Unfortunately this is not how the heap size is set. The heap size is set by the runtime using a `ComputerBudget` instruction ([here](https://github.com/solana-labs/solana/blob/master/sdk/src/compute_budget.rs#L31)). Given that we can rely on the runtime to enforce the limit, we can assume the maximum limit so that program do not need to be modified when a different limit is used.

febo:
Thank you for your contribution. It is almost there, but there are a few bits we could improve. I prepared a PR #301 with an alternative. The main differences is that it uses the maximum limit that can be requested and does not use checked/saturated math, which saves some CUs, but at the same time it avoids overflows.

rajgoesout:
Thank you for the review and the explanation!

PR #301, opened 2025-12-16
sdk: Use forward allocation

febo:
### Problem

Currently, the `BumpAllocator` allocates memory backwards and it has a fixed limit of `32KiB` size. Even if a program requests a larger heap size, this won't be used unless a custom allocator is implemented.

### Solution

Refactor the `BumpAllocator` to allocate memory forward. This has 2 benefits:

1) It is slightly more efficient, saving `~7` CUs per allocation;
2) Removes the need to implement a custom allocator when a different heap size is requested.

The implementation relies on the runtime to zero initialize the heap region and to enforce the available heap length. The execution of a program is sand-boxed and the runtime prevents memory access outside the program memory space.

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -18,10 +18,7 @@ use crate::{
     Address, BPF_ALIGN_OF_U128, MAX_TX_ACCOUNTS,
 };
 
-#[cfg(all(
-    any(test, target_os = "solana", target_arch = "bpf"),
-    feature = "alloc"
-))]
+#[cfg(feature = "alloc")]
 pub use alloc::BumpAllocator;
 
 /// Start address of the memory region used for program heap.
@@ -503,7 +500,7 @@ macro_rules! default_allocator {
         #[cfg(any(target_os = "solana", target_arch = "bpf"))]
         #[global_allocator]
         static A: $crate::entrypoint::BumpAllocator = unsafe {
-            $crate::entrypoint::BumpAllocator::new(
+            $crate::entrypoint::BumpAllocator::new_unchecked(
                 $crate::entrypoint::HEAP_START_ADDRESS as usize,
                 // Use the maximum heap length allowed. Programs can request heap sizes up
                 // to this value using the `ComputeBudget`.
@@ -606,13 +603,10 @@ unsafe impl GlobalAlloc for NoAllocator {
     }
 }
 
-#[cfg(all(
-    any(test, target_os = "solana", target_arch = "bpf"),
-    feature = "alloc"
-))]
+#[cfg(feature = "alloc")]
 mod alloc {
     use {
-        crate::hint::unlikely,
+        crate::{entrypoint::MAX_HEAP_LENGTH, hint::unlikely},
         core::{
             alloc::{GlobalAlloc, Layout},
             mem::size_of,
@@ -639,16 +633,13 @@ mod alloc {
     impl BumpAllocator {
         /// Creates the allocator tied to specific range of addresses.
         ///
-        /// The start address must be aligned to `usize` and the length must be
-        /// at least `size_of::<usize>()` bytes.
-        ///
         /// # Safety
         ///
         /// This is unsafe in most situations, unless you are totally sure that
         /// the provided start address and length can be written to by the allocator,
         /// and that the memory will be usable for the lifespan of the allocator.
-        /// The reserved memory region must be at least `size_of::<usize>()` bytes
-        /// long.
+        /// The start address must be aligned to `usize` and the length must be
+        /// at least `size_of::<usize>()` bytes.
         ///
         /// For Solana on-chain programs, a certain address range is reserved, so
         /// the allocator can be given those addresses. In general, the `len` is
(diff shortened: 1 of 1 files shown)


sdk/src/entrypoint/mod.rs
```diff
@@ -595,62 +606,137 @@ unsafe impl GlobalAlloc for NoAllocator {
...
+                end: start + len,
+            }
+        }
     }
 
-    /// Integer arithmetic in this global allocator implementation is safe when operating on the
-    /// prescribed [`HEAP_START_ADDRESS`] and [`HEAP_LENGTH`]. Any other use may overflow and is
-    /// thus unsupported and at one's own risk.
+    // Integer arithmetic in this global allocator implementation is safe when operating on the
+    // prescribed `GlobalAlloc::start` and `GlobalAlloc::end`. Any other use may overflow and
```
joncinque: nit: shouldn't this refer to `BumpAllocator`'s fields?
```suggestion
    // prescribed `BumpAllocator::start` and `BumpAllocator::end`. Any other use may overflow and
```

sdk/src/entrypoint/mod.rs
```diff
@@ -595,62 +606,137 @@ unsafe impl GlobalAlloc for NoAllocator {
...
+        /// The start address must be aligned to `usize` and the length must be
+        /// at least `size_of::<usize>()` bytes.
+        ///
+        /// # Safety
+        ///
+        /// This is unsafe in most situations, unless you are totally sure that
+        /// the provided start address and length can be written to by the allocator,
+        /// and that the memory will be usable for the lifespan of the allocator.
+        /// The reserved memory region must be at least `size_of::<usize>()` bytes
+        /// long.
```
joncinque: nit: Up to you, but should we specify that the start address must align to `usize`?

Lichtso (dismissed):

sdk/src/entrypoint/mod.rs
```diff
@@ -595,62 +606,137 @@ unsafe impl GlobalAlloc for NoAllocator {
...
     unsafe impl GlobalAlloc for BumpAllocator {
-        /// Allocates memory as a bump allocator.
+        /// Allocates memory as described by the given `layout` using a forward bump allocator.
+        ///
+        /// Returns a pointer to newly-allocated memory, or `null` to indicate allocation failure.
+        ///
+        /// # Safety
+        ///
+        /// `layout` must have non-zero size. Attempting to allocate for a zero-sized layout will
+        /// result in undefined behavior.
```
joncinque, line 668: I believe this is ok, since most users of allocators no-op if the type is zero-sized, ie https://doc.rust-lang.org/src/alloc/boxed.rs.html#571

sdk/src/entrypoint/mod.rs
```diff
@@ -595,62 +606,137 @@ unsafe impl GlobalAlloc for NoAllocator {
...
+            if unlikely(pos == 0) {
                 // First time, set starting position.
-                pos = self.start + self.len;
+                pos = self.start + size_of::<usize>();
+            }
+
+            let padding = layout.align() - 1;
+
+            if unlikely(self.end - pos < padding) {
+                return null_mut();
```
joncinque: Maybe I'm being dense, but is this check necessary? `allocation` finds the start address, so we should just need to later check that `self.end > allocation + layout.size()`

sdk/src/entrypoint/mod.rs
```diff
@@ -595,62 +606,137 @@ unsafe impl GlobalAlloc for NoAllocator {
...
 
+        /// Behaves like `alloc`, but also ensures that the contents are set to zero before being returned.
+        ///
+        /// This method relies on the runtime to zero out the memory when reserving the heap region,
+        /// so it simply calls `alloc`.
         #[inline]
-        unsafe fn dealloc(&self, _: *mut u8, _: Layout) {
-            // I'm a bump allocator, I don't free.
+        unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
+            self.alloc(layout)
```
joncinque, line 705: hehe, very cool!

sdk/src/entrypoint/mod.rs
```diff
@@ -18,15 +18,22 @@ use crate::{
     Address, BPF_ALIGN_OF_U128, MAX_TX_ACCOUNTS,
 };
 
-#[cfg(all(any(target_os = "solana", target_arch = "bpf"), feature = "alloc"))]
+#[cfg(all(
+    any(test, target_os = "solana", target_arch = "bpf"),
```
joncinque: nit: up to you, but if we gate the tests on the `alloc` feature, then we don't need the `#[cfg(...test...)` part, since the `alloc` feature is enabled on pinocchio as a dev dependency

sdk/src/entrypoint/mod.rs
febo: I was avoiding doing the `allocation + layout.size()` operation without checks since it can overflow if you try to allocate `u64::MAX` for example. And `allocation` might be bigger than `self.end` if there is no "space" left for the alignment padding.

We could probably save `1` CU by testing that `layout.size()` is not greater than `MAX_HEAP_SIZE` and then do the check `self.end > allocation + layout.size()` instead. I will try that.

sdk/src/entrypoint/mod.rs
febo: Good point!

sdk/src/entrypoint/mod.rs
febo: It does save `1` extra CU. 😊

joncinque (approved):
Perfect!

PR #304, opened 2025-12-30
Optimize `write_bytes` function

cryptopapi997:
We can save quite a few CUs spent on `write_bytes` by using `core::ptr::copy_nonoverlapping` instead of the previous looping approach. This is safe, as 

- we are aligned due to both pointers being `u8`
- both pointers have enough capacity by just picking the shorter one as the length
- we are non-overlapping thanks to rust's aliasing rules preventing this (unless the caller messes this up by using unsafe, however in that case it's on them)

So far, this func is used in the spl-token and token2022 crates. There, it gives the following benches for different CPIs (loop being how it's currently being done and copy being what this PR introduces):

## Token

| Operation | Loop | Copy | Saved | % |
|-----------|------|--------|-------|---|
| Transfer | 5851 | 5850 | **1** | 0.0% |
| TransferChecked | 7458 | 7457 | **1** | 0.0% |
| MintTo | 5745 | 5744 | **1** | 0.0% |
| Burn | 5961 | 5960 | **1** | 0.0% |
| Approve | 4112 | 4111 | **1** | 0.0% |
| Revoke | 3834 | 3834 | **0** | 0.0% |
| FreezeAccount | 5471 | 5470 | **1** | 0.0% |
| ThawAccount | 5471 | 5470 | **1** | 0.0% |
| CloseAccount | 4121 | 4120 | **1** | 0.0% |
| InitializeMint | 4366 | 4177 | **189** | 4.3% |
| InitializeMint2 | 4144 | 3993 | **151** | 3.6% |
| InitializeAccount | 5780 | 5779 | **1** | 0.0% |
| InitializeAccount2 | 5781 | 5615 | **166** | 2.9% |
| InitializeAccount3 | 5551 | 5417 | **134** | 2.4% |
| SetAuthority | 4331 | 4200 | **131** | 3.0% |

## Token-2022

| Operation | Loop | Copy | Saved | % |
|-----------|------|--------|-------|---|
| Transfer | 2503 | 2503 | **0** | 0.0% |
| TransferChecked | 2843 | 2844 | **-1** | -0.0% |
| MintTo | 2296 | 2296 | **0** | 0.0% |
| Burn | 2355 | 2355 | **0** | 0.0% |
| Approve | 2099 | 2100 | **-1** | -0.0% |
| Revoke | 1853 | 1853 | **0** | 0.0% |
| FreezeAccount | 2096 | 2096 | **0** | 0.0% |
| ThawAccount | 2094 | 2094 | **0** | 0.0% |
| CloseAccount | 2312 | 2311 | **1** | 0.0% |
| InitializeMint | 2498 | 2309 | **189** | 7.6% |
| InitializeMint2 | 2355 | 2202 | **153** | 6.5% |
| InitializeAccount | 2759 | 2760 | **-1** | -0.0% |
| InitializeAccount2 | 2749 | 2581 | **168** | 6.1% |
| InitializeAccount3 | 2600 | 2466 | **134** | 5.2% |
| SetAuthority | 2101 | 1970 | **131** | 6.2% |

The code for generating these benches can be found [here](https://github.com/cryptopapi997/pinocchio-write-bytes-cu-bench) , just run it once with `main` and once with this branch to get the two benches.

### DIFF
--- a/programs/token-2022/src/lib.rs
+++ b/programs/token-2022/src/lib.rs
@@ -12,15 +12,11 @@ const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
 #[inline(always)]
 fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
     let len = destination.len().min(source.len());
-    // SAFETY: 
+    // SAFETY:
     // - Both pointers have alignment 1.
     // - For valid (non-UB) references, the borrow checker guarantees no overlap.
     // - `len` is bounded by both slice lengths.
     unsafe {
-        core::ptr::copy_nonoverlapping(
-            source.as_ptr(),
-            destination.as_mut_ptr() as *mut u8,
-            len,
-        );
+        core::ptr::copy_nonoverlapping(source.as_ptr(), destination.as_mut_ptr() as *mut u8, len);
     }
 }
--- a/programs/token/src/lib.rs
+++ b/programs/token/src/lib.rs
@@ -12,15 +12,11 @@ const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
 #[inline(always)]
 fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
     let len = destination.len().min(source.len());
-    // SAFETY: 
+    // SAFETY:
     // - Both pointers have alignment 1.
     // - For valid (non-UB) references, the borrow checker guarantees no overlap.
     // - `len` is bounded by both slice lengths.
     unsafe {
-        core::ptr::copy_nonoverlapping(
-            source.as_ptr(),
-            destination.as_mut_ptr() as *mut u8,
-            len,
-        );
+        core::ptr::copy_nonoverlapping(source.as_ptr(), destination.as_mut_ptr() as *mut u8, len);
     }
 }


Aursen:
In most cases, this tends to increase the CUs consumption. I tested it out on different programs (not very accurate because many factors for that). How do you benchmark?

cryptopapi997:
Appreciate the quick feedback!

I benched by looking at the copy patterns w the largest data size (67 bytes, from initalize_mint) and the smallest (9 bytes, from transfer) in this repo and measured the total those two take, which gives:

```
  Loop CU      Copy CU   Saved CU    Saved %
----------------------------------------------------------------
160           38        122      76.2%
 ```

After your comment I checked how much the individual ones do on each impl of write_bytes, the 9 byte version does 15 on both, while the 67 byte version does 153 vs 31, so my measurement of the total isn't as representative as it could be. Either way, from my benches it implies we either do equally good or better (but not worse).

I uploaded the bench code [here] (https://github.com/cryptopapi997/pinocchio-write-bytes-cu-bench), feel free to have a look what I might be doing different (although it's a bit ugly ;) )

**Comment by cryptopapi997** [comment]
^ but of course performance in actual programs is what matters, so if you had tested it on a couple of those and it clearly leads to worse performance feel free to close this PR

**Comment by Aursen** [comment]
The asm generated is smaller, I think you right. I maybe made a mistake on my benchmark

**Comment by publicqi** [comment]
can we try benchmarking token/token-2022 cpi as well? lots of other factors from llvm can affect cu, like register assignment, copy len etc. since the changes are in token and token-2022, i think real invocations would be better for reference

**Comment by cryptopapi997** [comment]
Added the bench for this to the PR description

File: programs/token-2022/src/lib.rs [hunk 1]
```diff
@@ -11,7 +11,16 @@ const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
 
 #[inline(always)]
 fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
-    for (d, s) in destination.iter_mut().zip(source.iter()) {
-        d.write(*s);
+    let len = destination.len().min(source.len());
+    // SAFETY:
```
febo:
```suggestion
    // SAFETY:
```

programs/token-2022/src/lib.rs
```diff
@@ -11,7 +11,16 @@ const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
...
+    // SAFETY: 
+    // - Both pointers have alignment 1.
+    // - For valid (non-UB) references, the borrow checker guarantees no overlap.
+    // - `len` is bounded by both slice lengths.
+    unsafe {
+        core::ptr::copy_nonoverlapping(
+            source.as_ptr(),
+            destination.as_mut_ptr() as *mut u8,
+            len,
+        );
```
febo:
```suggestion
        core::ptr::copy_nonoverlapping(source.as_ptr(), destination.as_mut_ptr() as *mut u8, len);
```

programs/token/src/lib.rs
```diff
@@ -11,7 +11,16 @@ const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
 
 #[inline(always)]
 fn write_bytes(destination: &mut [MaybeUninit<u8>], source: &[u8]) {
-    for (d, s) in destination.iter_mut().zip(source.iter()) {
-        d.write(*s);
+    let len = destination.len().min(source.len());
+    // SAFETY:
```
febo:
```suggestion
    // SAFETY:
```

programs/token/src/lib.rs
```diff
@@ -11,7 +11,16 @@ const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
...
+    // SAFETY: 
+    // - Both pointers have alignment 1.
+    // - For valid (non-UB) references, the borrow checker guarantees no overlap.
+    // - `len` is bounded by both slice lengths.
+    unsafe {
+        core::ptr::copy_nonoverlapping(
+            source.as_ptr(),
+            destination.as_mut_ptr() as *mut u8,
+            len,
+        );
```
febo:
```suggestion
        core::ptr::copy_nonoverlapping(source.as_ptr(), destination.as_mut_ptr() as *mut u8, len);
```

cryptopapi997:
Should be fixed now @febo

PR #307, opened 2026-01-05
add CreateAccountAllowPrefund builder

rustopian:
`CreateAccountAllowPrefund` is now available in `solana-system-interface` and `agave`. This adds a builder in pinocchio analogous to the existing system instruction builders.

### DIFF
--- a/programs/system/src/instructions/create_account_allow_prefund.rs
+++ b/programs/system/src/instructions/create_account_allow_prefund.rs
@@ -27,9 +27,11 @@ pub struct Funding<'a> {
 ///
 /// # Important
 ///
-/// This instruction does not warn if the account has more than
-/// enough lamports. Special care must be taken to ensure that
-/// the account being created has a lamports balance.
+/// Special care should be taken not to accidentally use a wallet
+/// account as the new account. This instruction allocates space
+/// and assigns the account to a new program owner, which makes
+/// the account unusable as a wallet. No warning is given if the
+/// account has more than enough lamports.
 ///
 /// ### Accounts:
 ///   0. `[WRITE, SIGNER]` New account
@@ -77,7 +79,11 @@ impl<'a, 'b> CreateAccountAllowPrefund<'a, 'b> {
             to,
             space,
             owner,
-            funding: Some(Funding { from, lamports }),
+            funding: if lamports == 0 {
+                None
+            } else {
+                Some(Funding { from, lamports })
+            },
         })
     }
 


programs/system/src/instructions/create_account_allow_prefund.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub fn with_minimal_balance(
+        from: &'a AccountView,
+        to: &'a AccountView,
+        rent_sysvar: &'a AccountView,
+        space: u64,
+        owner: &'a Address,
+    ) -> Result<Self, ProgramError> {
+        let rent = Rent::from_account_view(rent_sysvar)?;
+        let required_lamports = rent.try_minimum_balance(space as usize)?;
+        let lamports = required_lamports.saturating_sub(to.lamports());
```
rustopian, line 76: This *does* allow a wallet to be bricked with a large lamport balance, which is explicitly the risk developers must avoid when using `CreateAccountAllowPrefund`.

But measures such as `checked_sub` here seem inappropriate, since the lamports pre-funded may be a little bit more than required rent, and any defensive check for an arbitrary amount of overage in lamports adds compute.

programs/system/src/instructions/create_account_allow_prefund.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub to: &'a AccountView,
+
+    /// Number of bytes of memory to allocate.
+    pub space: u64,
+
+    /// Address of program that will own the new account.
+    pub owner: &'b Address,
+
+    /// Funding account and lamports to transfer to the new account.
+    pub payer_and_lamports: Option<(&'a AccountView, u64)>,
```
febo: Wondering whether would be better to keep the same "interface" as `CreateAccount` or not – i.e., `from` and `lamports` as separate fields.

programs/system/src/instructions/create_account_allow_prefund.rs
febo: Could we document this "risk"?

programs/system/src/instructions/create_account_allow_prefund.rs
rustopian: I thought about this, but it mirrors the same change elsewhere, and semantically keeps the two together

Up to you tho; will change if you prefer

febo: Make sense. What do you think if we are more explicit about the role of these values?
```rust
/// Funding lamports to transfer into a newly created account.
pub struct Funding<'a> {
    /// Funding account.
    pub from: &'a AccountView,

    /// Number of lamports to transfer to the new account.
    pub lamports: u64,
}
```
Then the field would be:
```rust
pub funding: Option<Funding<'a>>,
```
And you would use as:
```rust
CreateAccountPrefund {
    funding: Funding {
        from,
        lamports: 1_000_000_000,
    },
    ...
}
```

programs/system/src/instructions/create_account_allow_prefund.rs
```diff
@@ -0,0 +1,100 @@
+use pinocchio::{
+    cpi::{invoke_signed, Signer},
+    error::ProgramError,
+    instruction::{InstructionAccount, InstructionView},
+    sysvars::{rent::Rent, Sysvar},
+    AccountView, Address, ProgramResult,
+};
+
+/// Create a new account without the `lamports==0` assertion.
```
febo: nit: Can we expand a little bit the description?
```suggestion
/// Create a new account, which can be prefunded.
///
/// While [`super::CreateAccount`] can only be used when the account being created has
/// no lamports, `CreatePrefund` relaxes this requirement.
```

programs/system/src/instructions/create_account_allow_prefund.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl<'a, 'b> CreateAccountAllowPrefund<'a, 'b> {
+    #[inline(always)]
+    /// Creates a new `CreateAccountAllowPrefund` instruction with the minimum balance required
+    /// for the account. The caller must provide a `payer` if the account needs lamports;
+    /// otherwise, the resulting instruction will fail (downstream) when invoked.
+    ///
+    /// This instruction does not warn if the account has more than enough lamports; large
+    /// lamport balances can be frozen by `CreateAccountAllowPrefund` if used incorrectly.
```
febo: nit: Maybe we should put this under a "`# Important`" section and expand a bit on the risk of "bricking" a wallet account if it is mistakenly passed as the account to create.

programs/system/src/instructions/create_account_allow_prefund.rs
```diff
@@ -0,0 +1,148 @@
...
+            Rent::get()?.try_minimum_balance(space as usize)?
+        };
+
+        let lamports = required_lamports.saturating_sub(to.lamports());
+
+        Ok(Self {
+            to,
+            space,
+            owner,
+            funding: Some(Funding { from, lamports }),
```
grod220: If lamports == 0, this probably should return `funding: None`. Without it, I think this would make the callers supply an unnecessary writeable signer. Probably also needs a test case for this path.

programs/system/src/instructions/create_account_allow_prefund.rs
```diff
@@ -0,0 +1,148 @@
...
+
+impl<'a, 'b> CreateAccountAllowPrefund<'a, 'b> {
+    /// Creates a new `CreateAccountAllowPrefund` instruction with the minimum
+    /// balance required for the account.
+    ///
+    /// If the account already has lamports to cover the minimum balance, then
+    /// no lamports will be transferred.
+    #[inline(always)]
+    pub fn with_minimum_balance(
+        from: &'a AccountView,
```
grod220, line 64: Should we have a method variant that doesn't require the `from` if a caller knows the account is fully prefunded and just needs allocate + assign?

programs/system/src/instructions/create_account_allow_prefund.rs
```diff
@@ -0,0 +1,148 @@
...
+///
+/// This instruction is identical to `CreateAccount` except
+/// that it allows the account being created to already have
+/// lamports in it.
+///
+/// # Important
+///
+/// This instruction does not warn if the account has more than
+/// enough lamports. Special care must be taken to ensure that
+/// the account being created has a lamports balance.
```
grod220: The second sentence is a bit confusing to me. It sounds like having lamports is the thing you want. Is this perhaps a better wording?

> Special care must be taken not to pass an existing wallet or account that should retain its current owner. This instruction will allocate space and assign the account to a new program owner, permanently locking any lamports it holds.

programs/system/src/instructions/create_account_allow_prefund.rs
febo: <del>In that case they don't need to use the constructor and can set the fields directly with `funding: None`.</del>

In this case, they probably don't need to use this helper since this one is used only for calling `CreateAccountAllowPrefund`.

programs/system/src/instructions/create_account_allow_prefund.rs
febo: <del>Good point. It is not accurate to say "permanently locking any lamports it holds". I will update it.</del>

update: Sorry, now I understood your comment. I updated the message.

programs/system/src/instructions/create_account_allow_prefund.rs
febo: Good point!

grod220:
Looks good!

PR #308, opened 2026-01-05
system: Make rent sysvar optional

febo:
### Problem

Currently the create account CPI helpers `*with_minimal_balance` always require the rent `AccountView`.

### Solution

Make the argument optional and use `Sysvar::get` when the `AccountView` is not passed in. The PR also renames `with_minimal_balance` to `with_minimum_balance` to better match the existing SDK naming.


rustopian (approved):
Looks good all around.

I've also adopted these changes + the more flexible lifetimes in the related #307.

PR #310, opened 2026-01-07
readme: Clarify use of `nostd_panic_handler!` macro

febo:
### Problem

Not much information about the `nostd_panic_handler!` macro and when it should be used is included in the readme.

### Solution

Expand the description for the macro and signal when it should be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ program's `Cargo.toml`:
 solana-define-syscall = { version = "4.0.1", features = ["unstable-static-syscalls"] }
 ```
 
-When compiling your program with the upstream BPF target, the `std` library is no available. Therefore, the program crate must include the `#![no_std]` crate-level attribute and use the `nostd_panic_handler!` macro. An allocator may be used as long as `alloc` is used.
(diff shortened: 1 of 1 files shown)


sdk/src/lib.rs
```diff
@@ -249,6 +308,10 @@
 //! # Enable static syscalls for BPF target
 //! solana-define-syscall = { version = "4.0.1", features = ["unstable-static-syscalls"] }
 //! ```
+//!
+//! When compiling your program with the upstream BPF target, the `std` library is no available.
```
joncinque: nit typo
```suggestion
//! When compiling your program with the upstream BPF target, the `std` library is not available.
```

README.md
```diff
@@ -284,6 +291,8 @@ program's `Cargo.toml`:
 solana-define-syscall = { version = "4.0.1", features = ["unstable-static-syscalls"] }
 ```
 
+When compiling your program with the upstream BPF target, the `std` library is no available. Therefore, the program crate must include the `#![no_std]` crate-level attribute and use the `nostd_panic_handler!` macro. An allocator may be used as long as `alloc` is used.
```
joncinque:
```suggestion
When compiling your program with the upstream BPF target, the `std` library is not available. Therefore, the program crate must include the `#![no_std]` crate-level attribute and use the `nostd_panic_handler!` macro. An allocator may be used as long as `alloc` is used.
```

joncinque (commented):
Looks great! Just noticed a typo

PR #311, opened 2026-01-13
ci: integrate trusted publishing

yihau:
####  added `environment: prod` to semver_check

the pipeline flow is semver_check -> publish_release. if we add environments to both steps, we will need to approve twice

(let me know if you have any preference for the environment name)

#### remove `CARGO_REGISTRY_TOKEN` check

remove the token validation

#### switch to trusted publishing

use `rust-lang/crates-io-auth-action@v1` to get an ephemeral token and use it for the later publishing step

(btw, the environment isn't part of trusted publishing. we only use its approval mechanism to make the publishing process safer.)

---

Note:

set `Trusted Publishing` hook
- [x] pinocchio
- [x] pinocchio-token-2022
- [x] pinocchio-token
- [x] pinocchio-system
- [x] pinocchio-memo
- [x] pinocchio-associated-token-account

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,7 +51,6 @@ jobs:
   semver_check:
     name: Check Semver
     runs-on: ubuntu-latest
-    environment: prod
     steps:
       - name: Git checkout
         uses: actions/checkout@v4
@@ -84,6 +83,7 @@ jobs:
   publish_release:
     name: Publish
     runs-on: ubuntu-latest
+    environment: prod
     needs: semver_check
     permissions:
       contents: write


.github/workflows/publish.yml
```diff
@@ -86,16 +87,8 @@ jobs:
...
-        env:
-          token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        if: ${{ env.token == '' }}
-        run: |
-          echo "The CARGO_REGISTRY_TOKEN secret variable is not set"
-          echo "Go to \"Settings\" -> \"Secrets and variables\" -> \"Actions\" -> \"New repository secret\"."
-          exit 1
-
       - name: Git checkout
         uses: actions/checkout@v4
```
nbelenkov: there is also checkout@v6 now, https://github.com/actions/checkout/releases

.github/workflows/publish.yml
```diff
@@ -51,6 +51,7 @@ jobs:
   semver_check:
     name: Check Semver
     runs-on: ubuntu-latest
+    environment: prod
```
nbelenkov: Maybe what we should do is allow this run without prod and then move prod to the publish step. Which would mean

run semver checks (they pass) (doesn't have prod) -> approval -> publish (in prod env)

So you only need to approve when the checks actually pass

nbelenkov:
For context, @febo @joncinque, the reason we need `environment: prod` is so we can authorise only certain users to publish releases by triggering the action. What the environment also allows to configure is that it requires 2 approvals to run the actual flow, so one to propose the action and another user to approve (we can configure who the approvers are in the environment), which makes the whole process much safer.

Speaking of this @yihau, we should also have this only run on protected branches

.github/workflows/publish.yml
nbelenkov: So this gets parsed literally, so you can actually inject commands into it like `version = "1.2.3; bad things"`. Low risk here, as you need to have a 2nd approver, but still better to first parse it properly

```
  env:
    INPUT_VERSION: ${{ inputs.version }}
    INPUT_LEVEL: ${{ inputs.level }}

- run: |
    if [ "${{ inputs.level }}" = "version" ]; then
      LEVEL="$INPUT_VERSION"
    else
      LEVEL="$INPUT_LEVEL"
    fi
```

.github/workflows/publish.yml
yihau: ~I added the environment here because I thought we wanna support approval publishing as well. I can remove it if we don't need it.~

---

updated:

I think this should happen at the beginning. if the semver check passes and we approve like 6h later, the check becomes ineffective

nbelenkov:
ok I think this is everyhing, otherwise looks good to me. Moved the other ideas out of this PR to keep it on topic, will move them to an issue

.github/workflows/publish.yml
nbelenkov: Sorry I am not sure if you saw this before I changed it, i misunderstood the flow initially. Basically if we have on publish and not semvar check it would still not work until permission is granted and only run the checks before triggering approval process

nbelenkov: True, that makes sense, let's stick with semver being the prod

febo: Wonder if we should have a specific job that runs the approval at the start or not. It will make it more explicit where the approval is happening. I am thinking more about how we will roll out this change to other repos in a consistent way.

yihau: there will be a pending action that requires approval. I can add you to my playground now so you can try it out 🫡

yihau: sorry for the misleading @nbelenkov. when we trigger a pipeline, it will checkout the ref by default. the check is still valid in that point! just moved it to the second step: [c655ed8](https://github.com/anza-xyz/pinocchio/pull/311/commits/c655ed8296605038044261a645cf13068238fcf8)

febo:
Do we need to add `pinocchio-associated-token-account` to the "Trusted Publishing" hook? It is one of the crates under the `programs` folder.

yihau:
> Do we need to add pinocchio-associated-token-account to the "Trusted Publishing" hook? It is one of the crates under the programs folder

good catch! I missed it. just added 🫡

nbelenkov:
I actually ended up inventoring all our crates today, and if I am not mistaken, there are 9 Pinocchio-associated crates:

```
pinocchio
pinocchio-token-2022
pinocchio-token
pinocchio-memo
pinocchio-associated-token-account
pinocchio-system
pinocchio-log
pinocchio-pubkey
pinocchio-log-macro
```

These are at least ones we have registered, do we also need to add the last 3 or maybe they have migrated to sdk?

febo:
> I actually ended up inventoring all our crates today, and if I am not mistaken, there are 9 Pinocchio-associated crates:
> 
> ```
> pinocchio
> pinocchio-token-2022
> pinocchio-token
> pinocchio-memo
> pinocchio-associated-token-account
> pinocchio-system
> pinocchio-log
> pinocchio-pubkey
> pinocchio-log-macro
> ```
> 
> These are at least ones we have registered, do we also need to add the last 3 or maybe they have migrated to sdk?

No need for the last 3, these migrated to the SDK already.

febo (approved):
Looks great!

yihau:
please let me know if you encounter any questions during the next release!

joncinque:
Still catching up on things from last week -- this looks great to me! Very simple and easy to understand.

PR #315, opened 2026-01-20
refactor(sdk): add macro to simplify logic to advance non dup accounts

BretasArthur1:
# Problem

The entrypoint contained duplicated logic for advancing and aligning the serialized input pointer when handling non-duplicated accounts. 

# Solution

Extract the pointer advancement and alignment logic into new helper macro, `advance_input_with_account`, and reuse it across the code path, preserving DRY.

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -213,8 +213,11 @@ macro_rules! align_pointer {
     };
 }
 
-/// Advance the serialized input pointer past a non-duplicated account and align it for the next one.
-macro_rules! advance_non_dup_account_input {
+/// Advance the input pointer in relation to a non-duplicated account.
+///
+/// The macro will add `STATIC_ACCOUNT_DATA` and the account length to
+/// the input pointer and align its address using [`align_pointer`].
+macro_rules! advance_input_with_account {
     ($input:ident, $account:expr) => {{
         $input = $input.add(STATIC_ACCOUNT_DATA);
         $input = $input.add((*$account).data_len as usize);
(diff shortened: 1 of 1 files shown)


sdk/src/entrypoint/mod.rs
```diff
@@ -206,13 +206,22 @@ macro_rules! align_pointer {
...
         // back to a pointer. The resulting pointer is guaranteed to be valid
-        // becauseit follows the layout serialized by the runtime.
+        // because it follows the layout serialized by the runtime.
         with_exposed_provenance_mut(
             ($ptr.expose_provenance() + (BPF_ALIGN_OF_U128 - 1)) & !(BPF_ALIGN_OF_U128 - 1),
         )
     };
 }
 
+/// Advance the serialized input pointer past a non-duplicated account and align it for the next one.
```
febo: nit: We can be a bit more precise here.
```suggestion
/// Advance the input pointer in relation to a non-duplicated account.
///
/// The macro will add `STATIC_ACCOUNT_DATA` and the account length to
/// the input pointer and align its address using [`align_pointer`].
```

sdk/src/entrypoint/mod.rs
```diff
@@ -206,13 +206,22 @@ macro_rules! align_pointer {
...
+macro_rules! advance_non_dup_account_input {
```
febo: nit: Maybe this conveys the idea that we are advancing the input with an account.
```suggestion
macro_rules! advance_input_with_account {
```

febo:
@BretasArthur1 If you agree with the name change for the macro, we need to replace all occurrences of `advance_non_dup_account_input`. Could we also please add a description for the PR?

PR #318, opened 2026-01-24
Adding SECURITY.MD

nbelenkov:
Adding a bug bounty program, based on `solana-sdk`

### DIFF
--- a/SECURITY.MD
+++ b/SECURITY.MD
@@ -1,0 +1,94 @@
+# Security Policy
+
+1. [Reporting security problems](#reporting)
+2. [Incident Response Process](#process)
+3. [Security Bug Bounties](#bounty)
+
+<a name="reporting"></a>
+## Reporting security problems in Pinocchio
+
+**DO NOT CREATE A GITHUB ISSUE** to report a security problem.
+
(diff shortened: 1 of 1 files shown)


SECURITY.MD
```diff
@@ -0,0 +1,94 @@
...
+For the affected packages, prepare a fix for the issue and push them to the private repository associated with the draft security advisory.
+There is no CI available in the private repository so you must build from source and manually verify fixes.
+Code review from the reporter is ideal, as well as from multiple members of the core development team.
+
+### 4. Notify Security Group Validators
+Once an ETA is available for the fix, a member of the anza-xyz/security-incident-response group should notify the validators so they can prepare for an update using the "Solana Red Alert" notification system.
+The teams are all over the world and it's critical to provide actionable information at the right time. Don't be the person that wakes everybody up at 2am when a fix won't be available for hours.
+
+### 5. Ship the patch
+Once the fix is accepted it may be distributed directly to validators as a patch, depending on the vulnerability.
```
febo, line 68: Is this standard text that we always include? It does not really apply to pinocchio since issues won't need a validator patch.

nbelenkov: Yeah, this is 1:1 copy from solana-sdk bb. We need to update wording on them in general, but I feel like we should have this one in place first and then update them all at once in the new future

PR #319, opened 2026-01-24
docs(entrypoint): document inlining risk for entrypoint

BretasArthur1:
# Problem

LTO can inline instruction handlers into entrypoint, inflating its stack frame and causing BPF stack overflows with large dispatch/CPI account arrays.

# Solution

Add documentation guidance recommending `#[inline(never)]` on instruction handlers to keep them out of the entrypoint stack frame if the case requires.

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -115,7 +115,7 @@ const STATIC_ACCOUNT_DATA: usize = size_of::<RuntimeAccount>() + MAX_PERMITTED_D
 /// In this case, it is not possible to use the `entrypoint` macro. Use the
 /// [`crate::program_entrypoint!`] macro instead and set up the allocator and panic handler
 /// manually.
-/// 
+///
 /// The compiler may inline the instruction handler (and its call tree) into the
 /// generated `entrypoint`, which can significantly increase the entrypoint stack frame. If your
 /// program has large instruction dispatch logic or builds sizable CPI account arrays, consider


sdk/src/entrypoint/mod.rs
```diff
@@ -110,6 +110,12 @@ const STATIC_ACCOUNT_DATA: usize = size_of::<RuntimeAccount>() + MAX_PERMITTED_D
 /// `#[panic_handler]` set by the `std`. Therefore, this macro should be used when the program or
 /// any of its dependencies are dependent on the `std` library.
 ///
+/// When using LTO(Link Time Optimization), the compiler may inline the instruction handler (and its call tree) into the
```
febo: nit: I don't think we need to mention LTO. There are other reasons where the compiler might decide to inline the `process_instruction` – e.g., adding an explicit `#[inline(always)]`.
```suggestion
/// The compiler may inline the instruction handler (and its call tree) into the
```

sdk/src/entrypoint/mod.rs
```diff
@@ -110,6 +110,12 @@ const STATIC_ACCOUNT_DATA: usize = size_of::<RuntimeAccount>() + MAX_PERMITTED_D
...
+/// generated `entrypoint`, which can significantly increase the entrypoint stack frame. If your
+/// program has large instruction dispatch logic or builds sizeable CPI account arrays, consider
```
febo:
```suggestion
/// program has large instruction dispatch logic or builds sizable CPI account arrays, consider
```

BretasArthur1:
ci is red, let me fix it

PR #321, opened 2026-02-02
refactor: migrate workspace automation from ts/zx to just

AkliluYirgalem:
Justfile CLI: Centralized all tasks into just. just -l provides an instant command menu.
Native development workflow: Replaced TypeScript scripts with justfile, removing Node.js from the pipeline.
Clean CI: Simplified workflows to single-line just calls, improving maintainability.


febo:
This is interesting, but we are moving to use "plain" `Makefiles` instead. You can see an example in the Token [repository](https://github.com/solana-program/token).

PR #325, opened 2026-02-04
docs: correct `STATIC_ACCOUNT_DATA` doc comment

alnoki:
Currently, the `STATIC_ACCOUNT_DATA` doc comment is technically wrong since the static size of an account in the input buffer *also* includes the `u64` rent epoch field.

Correct the existing description, and add a clarifying comment regarding the nature of serialization order

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -33,10 +33,11 @@ pub const MAX_HEAP_LENGTH: u32 = 256 * 1024;
 /// Value used to indicate that a serialized account is not a duplicate.
 pub const NON_DUP_MARKER: u8 = u8::MAX;
 
-/// The "static" size of an account in the input buffer.
+/// The "static" size of an account in the input buffer, less rent epoch field.
 ///
 /// This is the size of the account header plus the maximum permitted data
-/// increase.
+/// increase, accounting for static length before the variable-length data,
+/// which is followed by the rent epoch field.
 const STATIC_ACCOUNT_DATA: usize = size_of::<RuntimeAccount>() + MAX_PERMITTED_DATA_INCREASE;
 
 /// Declare the program entrypoint and set up global handlers.


alnoki:
@febo

febo (commented):
Not sure about this  change. A [`RuntimeAccount`](https://github.com/anza-xyz/solana-sdk/blob/master/account-view/src/lib.rs#L37-L78) does not have a rent epoch field, so the comment is accurate in that sense. Also, rent epoch is [deprecated](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0267-rent-epoch-in-vm-is-a-constant.md).

alnoki:
> A [`RuntimeAccount`](https://github.com/anza-xyz/solana-sdk/blob/master/account-view/src/lib.rs#L37-L78) does not have a rent epoch field

@febo I am confused by your comment, [since rent epoch is indeed serialized into the input buffer](https://github.com/anza-xyz/agave/blob/67963d68232074824fe0deb2476e9fd209d88f1d/program-runtime/src/serialization.rs#L551) and the doc comment says `The "static" size of an account in the input buffer`

Since rent epoch is included in the input buffer, it's required for deriving accurate offsets like in the below example (if rent epoch is excluded, ASM programs using offsets derived from the input buffer exhibit undefined behavior): 


```rust
use pinocchio::RuntimeAccount as RuntimeAccountHeader;

#[repr(C)]
pub struct RuntimeAccount<const DATA_SIZE: usize> {
    pub header: RuntimeAccountHeader,
    pub data: [u8; DATA_SIZE],
    pub rent_epoch: u64,
}

type EmptyRuntimeAccount = RuntimeAccount<{ runtime_data_size(data::DATA_LEN_ZERO as usize) }>;

#[repr(C, packed)]
/// Input buffer header for all instructions.
pub struct InputBufferHeader {
    pub n_accounts: u64,
    pub user: EmptyRuntimeAccount,
    pub tree_header: RuntimeAccountHeader,
}
```

febo:
> Since rent epoch is included in the input buffer, it's required for deriving accurate offsets like in the below example (if rent epoch is excluded, ASM programs using offsets derived from the input buffer exhibit undefined behavior):

I meant that the constant only refers to the account size and rent epoch is not part of it. Additionally, before rent epoch there are also padding bytes. Since rent epoch is deprecated, it can be considered padding. I think it will be confusing to refer to rent epoch going forwards, it is a concept that does not exist.

Another [PR](https://github.com/anza-xyz/pinocchio/blob/8781d6847c3ae5990dfa79fecbe90bb97d1848b0/sdk/src/entrypoint/mod.rs) is adding a detailed description of what is in the input buffer.

alnoki:
Closing per https://github.com/anza-xyz/pinocchio/blob/8781d6847c3ae5990dfa79fecbe90bb97d1848b0/sdk/src/entrypoint/mod.rs in https://github.com/anza-xyz/pinocchio/pull/325#issuecomment-3953577202

PR #327, opened 2026-02-06
token-2022: Add transfer fee instructions

febo:
### Problem

Currently there are no instruction helpers for `TranferFee` extension.

### Solution

Add `TransferFee` extension discriminator and instructions.

### DIFF
--- a/programs/token-2022/src/instructions/extensions/transfer_fee/initialize_transfer_fee_config.rs
+++ b/programs/token-2022/src/instructions/extensions/transfer_fee/initialize_transfer_fee_config.rs
@@ -49,12 +49,9 @@ impl InitializeTransferFeeConfig<'_, '_> {
         // Instruction data.
 
         let mut instruction_data = [UNINIT_BYTE; 78];
-        // Fixed part of the instruction data:
-        // - discriminators
-        // - transfer_fee_basis_points
-        // - maximum_fee
-        // - 2 bytes for each optional authority
-        let mut data_len = 2 + 2 + 8 + 2;
+        // At the start, data length is 2 bytes for the instruction and extension
+        // discriminators, plus 1 byte for the optional transfer fee authority.
+        let mut data_len = 2 + 1;
 
         instruction_data[0].write(ExtensionDiscriminator::TransferFee as u8);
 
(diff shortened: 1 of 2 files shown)


programs/token-2022/src/instructions/extensions/transfer_fee/withdraw_withheld_tokens_from_accounts.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    #[inline(always)]
+    pub fn invoke_signed(&self, signers: &[Signer]) -> ProgramResult {
+        if self.signers.len() > MAX_MULTISIG_SIGNERS {
+            return Err(ProgramError::InvalidArgument);
+        }
+
+        let expected_accounts = 3 + self.signers.len() + self.sources.len();
+
+        if expected_accounts > MAX_STATIC_CPI_ACCOUNTS {
```
febo, line 101: We could add an `"alloc"` feature and support more accounts.

febo:
@joncinque Rebased and ready for another look.

programs/token-2022/src/instructions/extensions/transfer_fee/initialize_transfer_fee_config.rs
```diff
@@ -0,0 +1,100 @@
...
+        if let Some(authority) = self.transfer_fee_config_authority {
+            instruction_data[2].write(1);
+            write_bytes(&mut instruction_data[3..35], authority.as_ref());
+            // Add 32 bytes for the authority.
+            data_len += size_of::<Address>();
+        } else {
+            instruction_data[2].write(0);
+        }
+        if let Some(authority) = self.withdraw_withheld_authority {
+            instruction_data[35].write(1);
```
joncinque: All of these after the first `COption` should write at `data_len` instead, since COption is actually packed in these instructions, meaning `None` comes out to just `[0]`, and `Some(pubkey)` comes out to `[1, ...pubkey_bytes]`

programs/token-2022/src/instructions/extensions/transfer_fee/withdraw_withheld_tokens_from_accounts.rs
joncinque: Nah that's fine, the max should be good

programs/token-2022/src/instructions/extensions/transfer_fee/set_transfer_fee.rs
```diff
@@ -0,0 +1,154 @@
...
+            &InstructionView {
+                program_id: self.token_program,
+                // SAFETY: instruction accounts has `expected_accounts` initialized.
+                accounts: unsafe {
+                    from_raw_parts(instruction_accounts.as_ptr() as _, expected_accounts)
+                },
+                data: &[
+                    ExtensionDiscriminator::TransferFee as u8,
+                    Self::DISCRIMINATOR,
+                ],
```
joncinque: This one needs to write the instruction data of the transfer_fee_basis_points and maximum_fee

PR #329, opened 2026-02-06
Token2022: Add ImmutableOwner token extension instruction builders

burhankhaja:
**Key Updates:** 
- Implements instruction builders for ImmutableOwner extension 

**Note:** 
- This PR is split from the larger Token2022 draft PR #267


febo:
Closed in favour of #335 since the immutable owner is also a token instruction.

PR #331, opened 2026-02-06
TEMP :  Pausable . Cpi Guard token Extensions  (PRs open in Turbin3 repo)

burhankhaja:
**These extensions are at the brink of being merged into anza, it just needs 2-3 little nit fixes, that is why we left them open in Turbin3 repo before opening Pr into anza**
- Cpi-guard
- Pausable

Here : *https://github.com/Turbin3/pinocchio-extensions/pulls*

**Just wanted to let you know that we already have worked on these extensions**


febo:
Thanks, created individual PRs for these extensions: #333 and #334

PR #333, opened 2026-02-08
token-2022: Add cpi guard instructions

febo:
### Problem

Currently there are no instruction helpers for `CpiGuard` extension.

### Solution

Add `CpiGuard` extension discriminator and instructions.


joncinque (dismissed):
LGTM -- just the `signers` field / args piece, I think `account` is ok here

febo:
@joncinque Ready for a second review.

PR #334, opened 2026-02-08
token-2022: Add pausable extension instructions

febo:
### Problem

Currently there are no instruction helpers for `Pausable` extension.

### Solution

Add `Pausable` extension discriminator and instructions.


febo:
Ready for a second review.

PR #335, opened 2026-02-08
token/token-2022: Add missing token instructions

febo:
### Problem

There are 4 missing instructions in both token and token-2022 helpers.

### Solution

Add the missing instructions:

* `GetAccountDataSize`
* `InitializeImmutableOwner`
* `AmountToUiAmount`
* `UiAmountToAmount`

### DIFF
--- a/programs/token-2022/src/instructions/ui_amount_to_amount.rs
+++ b/programs/token-2022/src/instructions/ui_amount_to_amount.rs
@@ -36,6 +36,8 @@ pub struct UiAmountToAmount<'a, 'b, 'c, const LENGTH: usize> {
 }
 
 impl<const LENGTH: usize> UiAmountToAmount<'_, '_, '_, LENGTH> {
+    pub const DISCRIMINATOR: u8 = 24;
+
     #[inline(always)]
     pub fn invoke(&self) -> ProgramResult {
         // Reserve 1 byte for the instruction discriminator, and the rest for
@@ -50,9 +52,8 @@ impl<const LENGTH: usize> UiAmountToAmount<'_, '_, '_, LENGTH> {
 
         let mut instruction_data = [UNINIT_BYTE; LENGTH];
 
-        // discriminator
-        instruction_data[0].write(24);
-        // amount
+        instruction_data[0].write(Self::DISCRIMINATOR);
+
         write_bytes(&mut instruction_data[1..], self.amount.as_bytes());
 
         invoke(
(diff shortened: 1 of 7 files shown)


programs/token-2022/src/instructions/mod.rs
joncinque, line 1: Just a consideration, but we could save on a lot of code if we just had the token-2022 crate depend on the token crate, and just re-export all the instructions and state types

programs/token-2022/src/instructions/get_account_data_size.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl GetAccountDataSize<'_, '_> {
+    #[inline(always)]
+    pub fn invoke(&self) -> ProgramResult {
+        invoke(
+            &InstructionView {
+                program_id: self.token_program,
+                accounts: &[InstructionAccount::readonly(self.mint.address())],
+                data: &[21],
```
joncinque: This isn't correct unfortunately -- the instruction takes `21` followed by each extension discriminant as a little-endian `u16`: https://github.com/solana-program/token-2022/blob/0033c60f8fb641e07fb00a576f7428c8b52cbe54/interface/src/instruction.rs#L999

I'm not sure if we want to support the actual extensions just yet

programs/token-2022/src/instructions/ui_amount_to_amount.rs
```diff
@@ -0,0 +1,68 @@
...
+    // The token program.
+    pub token_program: &'b Address,
+}
+
+impl<const LENGTH: usize> UiAmountToAmount<'_, '_, '_, LENGTH> {
+    #[inline(always)]
+    pub fn invoke(&self) -> ProgramResult {
+        // Reserve 1 byte for the instruction discriminator, and the rest for
+        // the amount string.
+        let expected_data = 1 + self.amount.len();
```
joncinque, line 45: I got worried that this could potentially cause an overrun, but there's defensive checks in `write_bytes`, and `str.len()` actually gives the byte length, not character length, so we're all good!

programs/token-2022/src/instructions/mod.rs
febo: There is a tiny difference between them. The token ones don't take a `token_program` address parameter (they are fixed to use the token address), while the token-2022 have an extra `token_program` parameter.

joncinque: Ah darn

programs/token-2022/src/instructions/get_account_data_size.rs
febo: Got it! I completed the implementation. It should support all extensions when they are available.

febo:
Ready for a second look.

PR #336, opened 2026-02-08
token-2022: Add mint close authority instruction

febo:
### Problem

Currently there are no instruction helpers for `MintCloseAuthority` extension.

### Solution

Add `MintCloseAuthority` extension discriminator and instruction.

### DIFF
--- a/programs/token-2022/src/instructions/extensions/mint_close_authority.rs
+++ b/programs/token-2022/src/instructions/extensions/mint_close_authority.rs
@@ -25,6 +25,9 @@ pub struct InitializeMintCloseAuthority<'a, 'b> {
 
     /// Authority that must sign the `CloseAccount` instruction on a mint.
     pub close_authority: Option<&'b Address>,
+
+    // The token program.
+    pub token_program: &'b Address,
 }
 
 impl InitializeMintCloseAuthority<'_, '_> {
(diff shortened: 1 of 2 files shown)


programs/token-2022/src/instructions/extensions/mod.rs
```diff
@@ -1,10 +1,12 @@
 pub mod default_account_state;
 pub mod memo_transfer;
+pub mod mint_close_authority;
 pub mod transfer_hook;
 
 #[repr(u8)]
 #[non_exhaustive]
 pub enum ExtensionDiscriminator {
+    MintCloseAuthority = 26,
```
joncinque: This should be `25`
```suggestion
    MintCloseAuthority = 25,
```

programs/token-2022/src/instructions/extensions/mint_close_authority.rs
```diff
@@ -0,0 +1,59 @@
...
+        // discriminator
+        instruction_data[0].write(ExtensionDiscriminator::MintCloseAuthority as u8);
+        // close_authority
+        if let Some(authority) = self.close_authority {
+            instruction_data[1].write(1);
+            write_bytes(&mut instruction_data[2..34], authority.as_ref());
+            expected_data += 32;
+        } else {
+            instruction_data[1].write(0);
+        }
```
joncinque: Should these write use the `get_mut_unchecked` accesses? Or does the compiler realize that these are safe and remove the bounds checks?

joncinque (commented):
Just a discriminator to fixup

programs/token-2022/src/instructions/extensions/mint_close_authority.rs
febo: Replaced the `get_mut_unchecked` since it does not make a difference in this context.

febo:
Ready for a second look.

joncinque (approved):
Looks great!

PR #337, opened 2026-02-08
token-2022: Add `CreateNativeMint` instruction

febo:
### Problem

Currently there is no `CreateNativeMint` instruction.

### Solution

Add `CreateNativeMint` instruction.

### DIFF
--- a/programs/token-2022/src/instructions/mod.rs
+++ b/programs/token-2022/src/instructions/mod.rs
@@ -26,10 +26,10 @@ mod unwrap_lamports;
 mod withdraw_excess_lamports;
 
 pub use {
-    approve::*, approve_checked::*, burn::*, burn_checked::*, close_account::*, create_native_mint,
-    extensions::*, freeze_account::*, initialize_account::*, initialize_account_2::*,
-    initialize_account_3::*, initialize_mint::*, initialize_mint_2::*, initialize_multisig::*,
-    initialize_multisig_2::*, initialize_non_transferable_mint::*, mint_to::*, mint_to_checked::*,
-    revoke::*, set_authority::*, sync_native::*, thaw_account::*, transfer::*, transfer_checked::*,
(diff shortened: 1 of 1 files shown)


Cargo.lock
joncinque, line 1: So many more deps for one function, sorry 😅

programs/token-2022/src/instructions/create_native_mint.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Creates the native mint.
+///
+/// This instruction only needs to be invoked once after deployment and is
+/// permissionless, Wrapped SOL ([`crate::native_mint::id()`]) will not be
+/// available until this instruction is successfully executed.
+///
+/// Accounts expected by this instruction:
+///
+///   0. `[writeable,signer]` Funding account (must be a system account).
```
joncinque: nit typo
```suggestion
///   0. `[writable,signer]` Funding account (must be a system account).
```

programs/token-2022/src/instructions/create_native_mint.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// System program for mint account funding.
+    pub system_program: &'a AccountView,
+
+    /// The token program.
+    pub token_program: &'b Address,
+}
+
+impl CreateNativeMint<'_, '_> {
+    pub const DISCRIMINATOR: u8 = 32;
```
joncinque: This should be 31 instead
```suggestion
    pub const DISCRIMINATOR: u8 = 31;
```

programs/token-2022/src/instructions/create_native_mint.rs
febo: This one is here as well. 😊

https://github.com/solana-program/token-2022/blob/main/interface/src/instruction.rs#L632

febo:
This is ready for another look.

joncinque (approved):
Ship it!

PR #339, opened 2026-02-08
token-2022: Add `InitializeNonTransferableMint` instruction

febo:
### Problem

Currently there is no `InitializeNonTransferableMint` instruction.

### Solution

Add `InitializeNonTransferableMint` instruction.


joncinque (approved):
Perfect!

PR #341, opened 2026-02-08
token-2022: Add `WithdrawExcessLamports` instruction

febo:
### Problem

Currently there is no `WithdrawExcessLamports` instruction.

### Solution

Add `WithdrawExcessLamports` instruction.


joncinque (approved):
Looks good! Same comments as #342

febo:
> Looks good! Same comments as #342

Fixed the review comments.

PR #342, opened 2026-02-08
token-2022: Add `UnwrapLamports` instruction

febo:
### Problem

Currently there is no `UnwrapLamports` instruction.

### Solution

Add `UnwrapLamports` instruction.

### DIFF
--- a/programs/token-2022/src/instructions/unwrap_lamports.rs
+++ b/programs/token-2022/src/instructions/unwrap_lamports.rs
@@ -98,57 +98,39 @@ impl<'a, 'b, 'c> UnwrapLamports<'a, 'b, 'c> {
 
         // Instruction accounts.
 
-        const UNINIT_INSTRUCTION_ACCOUNTS: MaybeUninit<InstructionAccount> =
-            MaybeUninit::<InstructionAccount>::uninit();
-        let mut instruction_accounts = [UNINIT_INSTRUCTION_ACCOUNTS; 3 + MAX_MULTISIG_SIGNERS];
+        let mut instruction_accounts =
+            [const { MaybeUninit::<InstructionAccount>::uninit() }; 3 + MAX_MULTISIG_SIGNERS];
 
-        // SAFETY: The allocation is valid to the maximum number of accounts.
-        unsafe {
-            instruction_accounts
-                .get_unchecked_mut(0)
-                .write(InstructionAccount::writable(self.source.address()));
+        instruction_accounts[0].write(InstructionAccount::writable(self.source.address()));
 
-            instruction_accounts
-                .get_unchecked_mut(1)
-                .write(InstructionAccount::writable(self.destination.address()));
+        instruction_accounts[1].write(InstructionAccount::writable(self.destination.address()));
 
-            instruction_accounts
-                .get_unchecked_mut(2)
-                .write(InstructionAccount::new(
-                    self.authority.address(),
-                    false,
-                    self.multisig_signers.is_empty(),
-                ));
+        instruction_accounts[2].write(InstructionAccount::new(
+            self.authority.address(),
+            false,
+            self.multisig_signers.is_empty(),
+        ));
 
-            for (account, signer) in instruction_accounts
-                .get_unchecked_mut(3..)
-                .iter_mut()
-                .zip(self.multisig_signers.iter())
-            {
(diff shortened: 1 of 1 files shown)


programs/token-2022/src/instructions/unwrap_lamports.rs
```diff
@@ -0,0 +1,185 @@
...
+        authority: &'a AccountView,
+        amount: Option<u64>,
+    ) -> Self {
+        Self::with_signers(token_program, account, destination, authority, amount, &[])
+    }
+
+    /// Creates a new `UnwrapLamports` instruction with a
+    /// multisignature owner/delegate authority and signer accounts.
+    #[inline(always)]
+    pub fn with_signers(
```
joncinque: nit: maybe we can update this once all the PRs land, but the name `with_signers` will almost certainly confuse people. Since the feature is not widely used, we can go with a longer and more annoying name, like `with_multisig_authority` or `with_multisig_signers` or `with_multisig`

programs/token-2022/src/instructions/unwrap_lamports.rs
```diff
@@ -0,0 +1,185 @@
...
+///   2. `[signer]` The source account's owner/delegate.
+///
+///   * Multisignature owner/delegate
+///   0. `[writable]` The source account.
+///   1. `[writable]` The destination account.
+///   2. `[]` The source account's multisignature owner/delegate.
+///   3. `..+M` `[signer]` M signer accounts.
+pub struct UnwrapLamports<'a, 'b, 'c> {
+    /// The source account.
+    pub account: &'a AccountView,
```
joncinque: nit: `account` is a very generic name, and confused me while reading the implementation, why not go with `source`?

programs/token-2022/src/instructions/unwrap_lamports.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    /// The destination account.
+    pub destination: &'a AccountView,
+
+    /// The owner/delegate account.
+    pub authority: &'a AccountView,
+
+    /// Multisignature owner/delegate.
+    pub signers: &'c [&'a AccountView],
```
joncinque: nit: similar to the other comment, let's give this a more descriptive name, like `multisig_signers`, so that people don't accidentally use it

programs/token-2022/src/instructions/unwrap_lamports.rs
```diff
@@ -0,0 +1,185 @@
...
+        // discriminator
+        instruction_data[0].write(Self::DISCRIMINATOR);
+        // amount
+        if let Some(amount) = self.amount {
+            instruction_data[1].write(1);
+            write_bytes(&mut instruction_data[2..10], &amount.to_le_bytes());
+            expected_data += 8;
+        } else {
+            instruction_data[1].write(0);
+        }
```
joncinque: Is there any benefit to use `get_unchecked_mut` for the array accesses?

joncinque (approved):
Since there are so many of these PRs, and I just have nits, it might be easier to make all of the changes afterwards.

Looks great overall!

programs/token-2022/src/instructions/unwrap_lamports.rs
febo: I was using `with_multisig` first, but then decided to change to `with_signers`. Since it is better to use `multisig_signers` for the field, we can use `with_multisig_signers`.

programs/token-2022/src/instructions/unwrap_lamports.rs
febo: I am almost convinced that it should not have a difference, but since we are using this in other parts, we might as well use it here.

joncinque: I realized in a later PR that the compiler should do the right thing and remove the bounds checks.

Check out this godbolt example -- with optimizations, there's no difference if the change is to a known index: https://rust.godbolt.org/z/rx11c1Tve

febo: Should we remove all the `get_unchecked_mut`? These are all static-sized arrays.

joncinque: Yeah sure, but only if the index is known at compile-time

joncinque (approved):
Perfect!

PR #343, opened 2026-02-08
token-2022: Add `PermissionedBurn` instructions

febo:
### Problem

Currently there are no instruction helpers for `PermissionedBurn` extension.

### Solution

Add `PermissionedBurn` extension discriminator and instructions.

### DIFF
--- a/programs/token-2022/src/instructions/extensions/permissioned_burn/burn.rs
+++ b/programs/token-2022/src/instructions/extensions/permissioned_burn/burn.rs
@@ -36,7 +36,7 @@ use {
 ///   4. `..4+M` `[signer]` M signer accounts for the multisig.
 pub struct Burn<'a, 'b, 'c> {
     /// The source account to burn from.
-    pub source: &'a AccountView,
+    pub account: &'a AccountView,
 
     /// The token mint.
     pub mint: &'a AccountView,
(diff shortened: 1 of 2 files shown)


programs/token-2022/src/instructions/extensions/permissioned_burn/burn.rs
```diff
@@ -0,0 +1,191 @@
...
+///   * Multisignature authority
+///   0. `[writable]` The source account to burn from.
+///   1. `[writable]` The token mint.
+///   2. `[signer]` The permissioned burn authority configured on the mint, if
+///      any.
+///   3. `[]` The source account's multisignature owner/delegate.
+///   4. `..4+M` `[signer]` M signer accounts for the multisig.
+pub struct Burn<'a, 'b, 'c> {
+    /// The source account to burn from.
+    pub source: &'a AccountView,
```
febo: Changed this to `source`, but I wonder if in this case `account` works better since it is a token account. Maybe `source` should be for cases where the account could be any of the program accounts.

joncinque: Both can work, I don't feel as strongly here as with the others.

febo: Ok, let's try to use `account` for when it is a token account; and `source` when could be multiple account types.

PR #338, opened 2026-02-08
token-2022: Add `Reallocate` instruction

febo:
### Problem

Currently there is no `Reallocate` instruction.

### Solution

Add `Reallocate` instruction.

PR #346, opened 2026-02-11
token2022: add reallocate token extension

ChiefWoods:


burhankhaja:
@febo we need your thoughts on this PR,

we had already done the work for reallocate while you were doing some nit changes in Turbin3 repo, 

we were waiting for your suggestions before opening its PR here, but later on we noticed you already had created your own PR for reallocate (which ofcourse is in draft right now )

febo:
> @febo we need your thoughts on this PR,
> 
> we had already done the work for reallocate while you were doing some nit changes in Turbin3 repo,
> 
> we were waiting for your suggestions before opening its PR here, but later on we noticed you already had created your own PR for reallocate (which ofcourse is in draft right now )

Yeah, we have #338 already implementing the instruction. It is on draft since I would like to merge it last.

febo:
#338 already implements the extension.

PR #347, opened 2026-02-11
sdk: Add `resize` feature

febo:
### Problem

As described in https://github.com/anza-xyz/solana-sdk/pull/569, account resize cannot be safely performed using a delta value. Therefore, entrypoints need to store the original data size to reliably track differences. But doing so consume CUs and not every program needs to resize accounts.

Additionally, `AccountView::resize` was removed since the resize logic is dependent on what the entrypoint does.

### Solution

This PR enables account resizing in pinocchio by:
* Adding a `"resize"` feature to indicate to the entrypoint that the original data size needs to be stored. This consumes `2` CUs per account.
* When the feature is enabled, an `impl` of the new `Resize` trait is provided for`AccountView`, offering `resize` and `resize_unchecked`.

Note that enabling the `"resize"` feature makes the entrypoint use 2 CUs extra per account.

### DIFF
--- a/programs/system/Cargo.toml
+++ b/programs/system/Cargo.toml
@@ -11,6 +11,10 @@ rust-version = { workspace = true }
 [lib]
 crate-type = ["rlib"]
 
+[features]
+account-resize = ["pinocchio/account-resize"]
+unsafe-account-resize = ["pinocchio/unsafe-account-resize"]
+
 [dependencies]
-pinocchio = { workspace = true, features = ["cpi", "resize"] }
+pinocchio = { workspace = true, features = ["cpi"] }
 solana-address = { workspace = true, features = ["decode"] }
--- a/programs/system/src/lib.rs
+++ b/programs/system/src/lib.rs
@@ -6,7 +6,7 @@ use {
         address::declare_id,
         cpi::Signer,
         sysvars::{rent::Rent, Sysvar},
-        AccountView, Address, ProgramResult, Resize,
+        AccountView, Address, ProgramResult,
     },
 };
 
@@ -74,13 +74,30 @@ pub fn create_account_with_minimum_balance_signed(
             .invoke_signed(signers)?;
         }
 
+        #[cfg(not(any(feature = "account-resize", feature = "unsafe-account-resize")))]
+        // When account resize is not enabled, allocate the required space for the account
+        // using the `Allocate` instruction.
+        crate::instructions::Allocate {
+            account,
+            space: space as u64,
+        }
+        .invoke_signed(signers)?;
+
         // Assign the account to the specified owner.
         Assign { account, owner }.invoke_signed(signers)?;
 
-        // Allocate the required space for the account.
+        // Allocate the required space for the account using the `AccountView::resize`.
         //
         // SAFETY: There are no active borrows of the `account`.
         // This was checked by the `Assign` CPI above.
-        unsafe { account.resize_unchecked(space) }
+        unsafe {
+            #[cfg(feature = "account-resize")]
+            <AccountView as pinocchio::Resize>::resize_unchecked(account, space)?;
+
+            #[cfg(all(feature = "unsafe-account-resize", not(feature = "account-resize")))]
+            <AccountView as pinocchio::UnsafeResize>::resize(account, space);
+        }
+
+        Ok(())
     }
 }


programs/system/Cargo.toml
```diff
@@ -12,5 +12,5 @@ rust-version = { workspace = true }
 crate-type = ["rlib"]
 
 [dependencies]
-pinocchio = { workspace = true, features = ["cpi"] }
+pinocchio = { workspace = true, features = ["cpi", "resize"] }
```
febo: Not sure if enable the `"resize"` feature by default or add a feature to this crate for that. Tempted to add a feature, otherwise every time you are using it, your entrypoint will use more CUs.

sdk/src/lib.rs
```diff
@@ -363,6 +369,144 @@ const BPF_ALIGN_OF_U128: usize = 8;
...
+    /// # Safety
+    ///
+    /// This method is unsafe because it does not check if the account data is
+    /// already borrowed. The caller must guarantee that there are no active
+    /// borrows to the account data.
+    unsafe fn resize_unchecked(&mut self, new_len: usize) -> Result<(), ProgramError>;
+}
+
+#[cfg(all(feature = "unsafe-resize", not(feature = "resize")))]
+pub trait UnsafeResize: sealed::Sealed {
```
febo, line 409: The idea of this trait is to offer `resize` without extra CU consumption. It is called "unsafe" because it does not do any checks, use it at your own risk sort of thing. We could still do the `MAX_PERMITTED_DATA_INCREASE` with the delta, as the previous implementation, but I thought that if we can't guarantee that it does not go over the limit, no point in trying to validate.

sdk/Cargo.toml
```diff
@@ -26,6 +26,8 @@ alloc = ["solana-instruction-view?/slice-cpi"]
 copy = ["solana-account-view/copy", "solana-address/copy"]
 cpi = ["dep:solana-instruction-view"]
 default = ["alloc"]
+resize = []
+unsafe-resize = []
```
febo: I am not sure about the name of the features, maybe `"account-resize"` is better?

programs/system/Cargo.toml
joncinque: A crate feature makes sense.

If a program uses the system program just for Assign, and never calls CreateProgram or Allocate, then they shouldn't need to incur the cost of the resize feature. That'll also encourage people to just use `Assign` and resize accounts themselves however they want.

sdk/Cargo.toml
joncinque: `account-resize` is better to me. For the other one, I'm torn between `account-resize-unsafe` and `unsafe-account-resize` -- they both sound good to me!

sdk/src/entrypoint/mod.rs
```diff
@@ -1,5 +1,61 @@
...
+//! |   │
+//! │   └─ If the account is not a duplicate:
+//! |         |
+//! │         ├─ 1 byte boolean, true if account is a signer
+//! |         |
+//! │         ├─ 1 byte boolean, true if account is writable
+//! |         |
+//! |         ├─ 1 byte boolean, true if account is executable
+//! |         |
+//! │         ├─ 4 bytes of padding (account data length should be stored here)
```
joncinque: nit: let's be pedantic and clear here
```suggestion
//! │         ├─ 4 bytes of padding (account data length stored here with `account-resize` feature)
```

sdk/src/lib.rs
joncinque: I agree with you, let's let people go wild with this one!

sdk/src/lib.rs
```diff
@@ -363,6 +369,144 @@ const BPF_ALIGN_OF_U128: usize = 8;
...
+            }
+        }
+
+        unsafe { (*account).data_len = new_len as u64 };
+
+        Ok(())
+    }
+}
+
+#[cfg(all(feature = "unsafe-resize", not(feature = "resize")))]
```
joncinque: Can we have the traits just living side-by-side rather than exclusive? That way, someone can potentially mix and match doing safe and unsafe stuff.

Unless that creates too much of a footgun

febo: Hmmm, might be a footgun. For example, if `UnsafeResize` is already imported, then any use of `resize` will just work with<del>out</del> warnings.

I imagine that they will be used exclusively. If you are "paying" the 2 CU overhead is because you want the extra checks; otherwise, you are happy to go the "unsafe" route.

update: There will be warnings, since `resize` on the `UnsafeResize` is unsafe. But there will be a name clash.

sdk/Cargo.toml
febo: I think `account-resize` and `unsafe-account-resize` work well.

programs/system/Cargo.toml
febo: Added features in https://github.com/anza-xyz/pinocchio/pull/347/commits/75eddb8faa84940fd34875d38d8a2806e7b2e22a

Currently it has the same feature names as in pinocchio. If no feature is enabled, it uses the `Allocate` instructions; otherwise it uses the corresponding `resize` method.

This only affects the `create_account_with_minimum_balance` helpers. This will be deprecated once `CreateAccountAllowPrefund` is enabled.

programs/system/src/lib.rs
```diff
@@ -74,13 +74,30 @@ pub fn create_account_with_minimum_balance_signed(
...
         }
 
+        #[cfg(not(any(feature = "account-resize", feature = "unsafe-account-resize")))]
+        // When account resize is not enabled, allocate the required space for the account
+        // using the `Allocate` instruction.
+        crate::instructions::Allocate {
+            account,
+            space: space as u64,
+        }
+        .invoke_signed(signers)?;
```
joncinque, line 84: Clever!

PR #348, opened 2026-02-12
token-2022: Nits of token-2022 extensions

febo:
### Problem

Some of the extensions are not following the same pattern.

### Solution

Update the extensions to make them consistent.

### DIFF
--- a/programs/token-2022/src/instructions/extensions/default_account_state/initialize.rs
+++ b/programs/token-2022/src/instructions/extensions/default_account_state/initialize.rs
@@ -1,5 +1,5 @@
 use {
-    crate::instructions::extensions::ExtensionDiscriminator,
+    crate::{instructions::extensions::ExtensionDiscriminator, state::AccountState},
     solana_account_view::AccountView,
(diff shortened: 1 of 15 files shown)


LStan:
Why did you change `impl SomeStruct<'_, '_, '_>` to `impl<'a, 'b, 'c> SomeStruct<'a, 'b, 'c>`, but didn't change `impl SomeStruct<'_, '_>` to `impl<'a, 'b, 'c> SomeStruct<'a, 'b>`?

febo:
> Why did you change `impl SomeStruct<'_, '_, '_>` to `impl<'a, 'b, 'c> SomeStruct<'a, 'b, 'c>`, but didn't change `impl SomeStruct<'_, '_>` to `impl<'a, 'b, 'c> SomeStruct<'a, 'b>`?

It all depends whether there is a constructor helper (`new` and `with_multisig_signers`) or not. They use the lifetimes, so we need to have them explicit in that case.

joncinque (approved):
Looks great!

PR #349, opened 2026-02-13
token-2022: Add TLV extension state infrastructure and initial extension structs

dev-jodee:


programs/token-2022/src/state/account_state.rs
```diff
@@ -14,13 +16,15 @@ pub enum AccountState {
...
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
         match value {
-            0 => AccountState::Uninitialized,
-            1 => AccountState::Initialized,
-            2 => AccountState::Frozen,
-            _ => panic!("invalid account state value: {value}"),
+            0 => Ok(AccountState::Uninitialized),
+            1 => Ok(AccountState::Initialized),
+            2 => Ok(AccountState::Frozen),
```
febo:
```suggestion
            // SAFETY: `value` is guaranteed to be in the range of the enum variants.
            0..=2 => Ok(unsafe { core::mem::transmute::<u8, AccountState>(value) }),
```

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+
+/// Marker trait for plain extension payload values that are safe to
+/// reinterpret from bytes.
+///
+/// # Safety
+///
+/// Implementors must be plain data with no padding and no invalid bit-patterns.
+pub unsafe trait Pod: sealed::SealedPod + Copy + 'static {}
+
+#[inline(always)]
```
febo: nit: We probably want to make error paths "cold", so they don't optimized.
```suggestion
#[cold]
```

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    pub trait SealedPod {}
+}
+
+/// Marker trait for plain extension payload values that are safe to
+/// reinterpret from bytes.
+///
+/// # Safety
+///
+/// Implementors must be plain data with no padding and no invalid bit-patterns.
+pub unsafe trait Pod: sealed::SealedPod + Copy + 'static {}
```
febo: Maybe we should name this something different, otherwise it gives the impression that it is related to `Pod` from bytemuck.

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+        18 => Ok(ExtensionType::MetadataPointer),
+        19 => Ok(ExtensionType::TokenMetadata),
+        20 => Ok(ExtensionType::GroupPointer),
+        21 => Ok(ExtensionType::TokenGroup),
+        22 => Ok(ExtensionType::GroupMemberPointer),
+        23 => Ok(ExtensionType::TokenGroupMember),
+        24 => Ok(ExtensionType::ConfidentialMintBurn),
+        25 => Ok(ExtensionType::ScaledUiAmount),
+        26 => Ok(ExtensionType::Pausable),
+        27 => Ok(ExtensionType::PausableAccount),
```
febo:
```suggestion
        // SAFETY: `extension_type` is guaranteed to be in the range of the enum variants.
        0..=27 => Ok(unsafe { core::mem::transmute::<u16, ExtensionType>(extension_type) }),
```

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    Ok(())
+}
+
+#[inline]
+fn validate_mint_extensions_data(data: &[u8]) -> Result<(), ProgramError> {
+    if data.len() <= ACCOUNT_TYPE_INDEX {
+        return Err(ProgramError::InvalidAccountData);
+    }
+
+    if data[Mint::BASE_LEN..BASE_ACCOUNT_LEN] != ZERO_MINT_PADDING {
```
febo, line 219: Is this a sanity check? The padding should always be zero.

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+
+#[inline(always)]
+unsafe fn mint_from_bytes_unchecked_mut(bytes: &mut [u8]) -> &mut Mint {
+    &mut *(bytes[..Mint::BASE_LEN].as_mut_ptr() as *mut Mint)
+}
+
+#[inline(always)]
+unsafe fn token_account_from_bytes_unchecked_mut(bytes: &mut [u8]) -> &mut TokenAccount {
+    &mut *(bytes[..TokenAccount::BASE_LEN].as_mut_ptr() as *mut TokenAccount)
+}
```
febo: Should these be on their respective type instead of stand-alone functions?

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+        let extension_type = extension_types[i];
+        validate_extension_account_type(extension_type, B::ACCOUNT_TYPE)?;
+
+        let mut j = 0;
+        let mut duplicate = false;
+        while j < i {
+            if extension_types[j] == extension_type {
+                duplicate = true;
+                break;
+            }
```
febo, line 283: Something to double-check, but I don't think duplicated extensions are accepted. If that is the case, we can fail here.

programs/token-2022/src/state/extension/permanent_delegate.rs
```diff
@@ -0,0 +1,39 @@
...
+};
+
+/// Permanent delegate extension data for mints (32 bytes).
+///
+/// When set on a mint, the delegate has unrestricted transfer and
+/// burn authority over all token accounts for the mint.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct PermanentDelegateExtension {
+    delegate: [u8; 32],
```
febo: Shouldn't this be an `Address`?

programs/token-2022/src/state/extension/transfer_hook.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Transfer hook extension data for mints (64 bytes).
+///
+/// Configures a custom program to execute additional logic on every
+/// transfer involving this mint.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct TransferHookExtension {
+    authority: [u8; 32],
+    program_id: [u8; 32],
```
febo: Same here, shouldn't this be addresses?

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1261 @@
...
+        let bytes = self.get_extension_bytes(V::TYPE)?;
+        extension_from_bytes(bytes)
+    }
+}
+
+/// A base state with TLV extension data backed by a checked borrow.
+///
+/// This type holds a [`Ref`] guard that keeps the account data borrow alive.
+/// Use this when you need safe, checked borrowing of the account data.
+pub struct RefStateWithExtensions<'a, B: ExtensionBaseState> {
```
febo: I think we don't need this. We can map the `Ref` given by `AccountView::try_borrow` to a `StateWithExtensions`.

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1261 @@
...
+
+    #[inline]
+    pub fn get_extension<V: ExtensionValue>(&self) -> Result<&V, ProgramError> {
+        let bytes = self.get_extension_bytes(V::TYPE)?;
+        extension_from_bytes(bytes)
+    }
+}
+
+/// A base state with TLV extension data backed by a checked mutable borrow.
+pub struct RefMutStateWithExtensions<'a, B: ExtensionBaseState> {
```
febo: Same thing here, we can map the `RefMut` given by `AccountView::try_borrow_mut` to a `MutStateWithExtensions`.

febo (commented):
This is a great start, sorry it took long to review it. Left a few comments for discussion.

febo (commented):
This is a great start, sorry it took long to review it. Left a few comments for discussion.

programs/token-2022/src/state/extension/mod.rs
dev-jodee: Yeah makes sense renamed to ExtensionPod

programs/token-2022/src/state/extension/mod.rs
dev-jodee: Yeah should always be 0, and if not then there's an issue and that's the sanity check for it. 

Has something similar

https://github.com/solana-program/token-2022/blob/efd0c957fefbd79882d77df5fb2dac88c001249c/program/src/extension/mod.rs#L306-L324

programs/token-2022/src/state/extension/mod.rs
dev-jodee: Good idea, moved them to the specific types

dev-jodee:
Updated with required changes / answered the questions :) thanks for the review

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1106 @@
...
+}
+
+#[inline(always)]
+pub(super) fn extension_from_bytes<T: ExtensionPod>(bytes: &[u8]) -> Result<&T, ProgramError> {
+    if bytes.len() != core::mem::size_of::<T>() {
+        return Err(ProgramError::InvalidAccountData);
+    }
+    if (bytes.as_ptr() as usize) % core::mem::align_of::<T>() != 0 {
+        return Err(ProgramError::InvalidAccountData);
+    }
```
febo: Something for a separate PR: it would be great to have a path to get extensions without having to do the alignment check when loading them from an `AccountView`.

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    TokenMetadata = 19,
+    GroupPointer = 20,
+    TokenGroup = 21,
+    GroupMemberPointer = 22,
+    TokenGroupMember = 23,
+    ConfidentialMintBurn = 24,
+    ScaledUiAmount = 25,
+    Pausable = 26,
+    PausableAccount = 27,
+}
```
grod220, line 125: It may be out of scope of this PR, but it would be nice to get support in `programs/token-2022/src/instructions/get_account_data_size.rs` and replaced the other type there:

```diff
pub struct GetAccountDataSize<'a, 'b, 'c> {
    pub mint: &'a AccountView,
-    pub extensions: &'c [ExtensionDiscriminator],
+    pub extensions: &'c [ExtensionType],
    pub token_program: &'b Address,
}
```

and serialize each value as a little-endian. Otherwise, `GetAccountDataSize` can't help with extensions like `ImmutableOwner`. 

@febo thoughts?

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    fn from(error: TokenError) -> Self {
+        ProgramError::Custom(error as u32)
+    }
+}
+
+/// SPL Token-2022 `TokenError::ExtensionNotFound` discriminant.
+pub const EXTENSION_NOT_FOUND_ERROR_CODE: u32 = TokenError::ExtensionNotFound as u32;
+
+mod sealed {
+    pub trait SealedExtensionPod {}
```
febo: nit: Maybe we can drop the `Pod` suffix since this is not related to bytemuck.

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/// Marker trait for plain extension payload values that are safe to
+/// reinterpret from bytes.
+///
+/// # Safety
+///
+/// Implementers must be plain data with no padding and no invalid bit-patterns.
+pub unsafe trait ExtensionPod: sealed::SealedExtensionPod + 'static {}
```
febo: nit: Same here, drop `Pod` suffix?

programs/token-2022/src/state/extension/mod.rs
febo: Yeah, we can fix that in a separate PR.

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+        | ExtensionType::CpiGuard
+        | ExtensionType::NonTransferableAccount
+        | ExtensionType::TransferHookAccount
+        | ExtensionType::ConfidentialTransferFeeAmount
+        | ExtensionType::PausableAccount => AccountType::Account,
+    }
+}
+
+#[inline(always)]
+fn extension_type_from_u16(extension_type: u16) -> Result<ExtensionType, ProgramError> {
```
febo: nit: Should we this be implemented as `impl TryFrom<u16> for ExtensionType`?

programs/token-2022/src/state/extension/permanent_delegate.rs
```diff
@@ -0,0 +1,37 @@
...
+};
+
+/// Permanent delegate extension data for mints (32 bytes).
+///
+/// When set on a mint, the delegate has unrestricted transfer and
+/// burn authority over all token accounts for the mint.
+#[repr(C)]
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct PermanentDelegateExtension {
+    delegate: Address,
```
febo: nit: This value is optional. Maybe we should wait for https://github.com/anza-xyz/solana-sdk/pull/605 and use `MaybeNull<Address>` here.

programs/token-2022/src/state/extension/permanent_delegate.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    #[inline(always)]
+    pub fn delegate(&self) -> &Address {
+        &self.delegate
+    }
+
+    #[inline(always)]
+    pub fn set_delegate(&mut self, delegate: &Address) {
+        self.delegate = delegate.clone();
+    }
```
febo: We could make the `delegate` field `pub` so we don't need accessors.

programs/token-2022/src/state/extension/transfer_hook.rs
```diff
@@ -0,0 +1,48 @@
...
+
+/// Transfer hook extension data for mints (64 bytes).
+///
+/// Configures a custom program to execute additional logic on every
+/// transfer involving this mint.
+#[repr(C)]
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct TransferHookExtension {
+    authority: Address,
+    program_id: Address,
```
febo: Same here, these are optional and can be made public so we don't need accessors.  Maybe we should wait for https://github.com/anza-xyz/solana-sdk/pull/605 and use `MaybeNull<Address>` here.

programs/token-2022/src/state/extension/transfer_hook_account.rs
```diff
@@ -0,0 +1,31 @@
+use super::{sealed, ExtensionPod, ExtensionType, ExtensionValue};
+
+/// Transfer hook account extension data (1 byte).
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct TransferHookAccountExtension {
+    transferring: u8,
```
febo: We could wait for https://github.com/anza-xyz/solana-sdk/pull/605 to land and use solana-zero-copy `Bool` type here.

febo (commented):
Few more small nits.

dev-jodee:
Fixed the nits, kept track of the two "future work" comment as well. Fixed the nits, kept track of the two "future work" comment as well.

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    TokenMetadata = 19,
+    GroupPointer = 20,
+    TokenGroup = 21,
+    GroupMemberPointer = 22,
+    TokenGroupMember = 23,
+    ConfidentialMintBurn = 24,
+    ScaledUiAmount = 25,
+    Pausable = 26,
+    PausableAccount = 27,
+}
```
grod220, line 110: Can we add `ExtensionType::PermissionedBurn` here too? 🙏

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+/// `collect_extension_types_from_tlv` or the wrapper `get_extension_types`
+/// methods.
+pub const MAX_EXTENSIONS: usize = 28;
+
+const BASE_ACCOUNT_LEN: usize = 165;
+const MINT_PADDING_LEN: usize = BASE_ACCOUNT_LEN - Mint::BASE_LEN;
+const ZERO_MINT_PADDING: [u8; MINT_PADDING_LEN] = [0u8; MINT_PADDING_LEN];
+const ACCOUNT_TYPE_INDEX: usize = BASE_ACCOUNT_LEN;
+const TLV_START_INDEX: usize = ACCOUNT_TYPE_INDEX + 1;
+const TLV_HEADER_LEN: usize = 4;
```
febo, line 39: nit: We could add comments to this constants as well.

programs/token-2022/src/state/extension/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+impl From<TokenError> for ProgramError {
+    #[inline(always)]
+    fn from(error: TokenError) -> Self {
+        ProgramError::Custom(error as u32)
+    }
+}
+
+/// SPL Token-2022 `TokenError::ExtensionNotFound` discriminant.
+pub const EXTENSION_NOT_FOUND_ERROR_CODE: u32 = TokenError::ExtensionNotFound as u32;
```
febo: nit: We could move this constant closer to the other ones.

scripts/setup/solana.dic
```diff
@@ -76,3 +76,7 @@ pausable
 permissioned
 endian
 prefunded
+TLV
+discriminant
+discriminants
```
febo:
```suggestion
discriminant/S
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1113 @@
...
+
+        let value_start = offset + TLV_HEADER_LEN;
+        let value_end = value_start + length;
+
+        if value_end > tlv_data.len() {
+            return Err(ProgramError::InvalidAccountData);
+        }
+        if count == out.len() {
+            return Err(ProgramError::AccountDataTooSmall);
+        }
```
febo, line 127: This is not necessarily an account data too small since the `out` is allocated by the caller.

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1113 @@
...
+    ///
+    /// The caller must ensure that it is safe to borrow the account
+    /// data (e.g., there are no mutable borrows of the account data).
+    #[inline]
+    pub unsafe fn from_account_view_unchecked(
+        account_view: &'a AccountView,
+    ) -> Result<Self, ProgramError> {
+        if account_view.data_len() < B::BASE_LEN {
+            return Err(ProgramError::InvalidAccountData);
+        }
```
febo: This is already covered by the `validate_state_with_extensions_data` below, right?

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1113 @@
...
+            return Err(ProgramError::InvalidAccountOwner);
+        }
+
+        let data = account_view.borrow_unchecked();
+        validate_state_with_extensions_data::<B>(data)?;
+
+        let base = B::from_bytes_unchecked(data);
+        let tlv_data = &data[extension_data_start::<B>(data.len())..];
+
+        Ok(Self { base, tlv_data })
```
febo: This can be replaced by a call to `from_bytes`.
```suggestion
        Self::from_bytes(account_view.borrow_unchecked())
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1113 @@
...
+            return Err(ProgramError::InvalidAccountOwner);
+        }
+
+        let data = account_view.borrow_unchecked_mut();
+        validate_state_with_extensions_data::<B>(data)?;
+
+        let (base_and_type, tlv_data) = data.split_at_mut(extension_data_start::<B>(data.len()));
+        let base = B::from_bytes_unchecked_mut(base_and_type);
+
+        Ok(Self { base, tlv_data })
```
febo: This can be replaced by a call to `from_bytes_mut`.
```suggestion
        Self::from_bytes_mut(account_view.borrow_unchecked_mut())
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1113 @@
...
+
+    #[inline(always)]
+    pub fn base(&self) -> &B {
+        self.base
+    }
+
+    #[inline(always)]
+    pub fn base_mut(&mut self) -> &mut B {
+        self.base
+    }
```
febo: Could we make the field `pub` instead of having these?

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1113 @@
...
+        let base = B::from_bytes_unchecked(data);
+        let tlv_data = &data[extension_data_start::<B>(data.len())..];
+
+        Ok(Self { base, tlv_data })
+    }
+
+    #[inline(always)]
+    pub fn base(&self) -> &B {
+        self.base
+    }
```
febo, line 291: Could we make the field `pub` instead of having this?

febo (commented):
Looks good – just a few more nits.

programs/token-2022/src/state/extension/state.rs
dev-jodee: Changed to InvalidArgument

programs/token-2022/src/state/extension/state.rs
dev-jodee: Yeah you're right, removed

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+        validate_state_with_extensions_data::<B>(data)?;
+        Ok(unsafe { Self::from_bytes_mut_unchecked(data) })
+    }
+
+    /// Return a `StateWithExtensionsMut` from the given account view.
+    ///
+    /// This method performs owner and length validation on `AccountView` and
+    /// checks the account data borrow state.
+    #[inline]
+    pub fn from_account_view(
```
febo: nit: Since this is returning a `RefMut`, let's add the `_mut` qualifier.
```suggestion
    pub fn from_account_view_mut(
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+    ///
+    /// This method performs owner and length validation on `AccountView`,
+    /// but does not perform the borrow check.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that it is safe to mutably borrow the account
+    /// data (e.g., there are no active borrows of the account data).
+    #[inline]
+    pub unsafe fn from_account_view_unchecked(
```
febo: nit: Same here, returning a `mut` reference.
```suggestion
    pub unsafe fn from_account_view_unchecked_mut(
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+    } else {
+        Err(ProgramError::InvalidAccountData)
+    }
+}
+
+/// Collect extension types from TLV data in encounter order.
+///
+/// Returns the number of written entries on success.
+#[inline]
+pub(super) fn collect_extension_types_from_tlv(
```
febo: nit: the `collect_*` prefix is a bit misleading here since we are not "producing a collection" but rather writing to a pre-allocated buffer. Maybe we could consider:
```suggestion
pub(super) fn write_extension_types_from_tlv(
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+    }
+}
+
+#[inline]
+fn validate_state_with_extensions_data<B: ExtensionBaseState>(
+    data: &[u8],
+) -> Result<(), ProgramError> {
+    if data.len() < B::BASE_LEN {
+        return Err(ProgramError::InvalidAccountData);
+    }
```
febo: Is this needed, given that we have `if data.len() <= BASE_ACCOUNT_LEN` afterwards?

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+    #[inline(always)]
+    fn tlv_data(&self) -> &[u8] {
+        let data_start = extension_data_start::<B>(B::BASE_LEN + self.data.len()) - B::BASE_LEN;
+        &self.data[data_start..]
+    }
+
+    /// Find the value bytes for a given extension type via
+    /// linear TLV walk.
+    #[inline]
+    pub fn get_extension_bytes(&self, target: ExtensionType) -> Result<&[u8], ProgramError> {
```
febo: Does this need to be public?

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+        validate_extension_account_type(target, B::ACCOUNT_TYPE)?;
+        get_extension_bytes_from_tlv(self.tlv_data(), target)
+    }
+
+    #[inline]
+    pub fn get_extension_types(&self, out: &mut [ExtensionType]) -> Result<usize, ProgramError> {
```
febo: nit: same here about the `get_`, it is more like a `write_`. We could also have a `get_extension_types` that requires `alloc::vec::Vec`.

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+            return Err(ProgramError::InvalidAccountOwner);
+        }
+
+        Self::from_bytes_mut(account_view.borrow_unchecked_mut())
+    }
+
+    /// Find the value bytes for a given extension type via
+    /// linear TLV walk.
+    #[inline]
+    pub fn get_extension_bytes(&self, target: ExtensionType) -> Result<&[u8], ProgramError> {
```
febo: nit: same comment here, does this need to be public?

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+        validate_extension_account_type(target, B::ACCOUNT_TYPE)?;
+        get_extension_bytes_from_tlv(self.tlv_data(), target)
+    }
+
+    /// Find the mutable value bytes for a given extension type via
+    /// linear TLV walk.
+    #[inline]
+    pub fn get_extension_bytes_mut(
```
febo: nit: same comment here, does this need to be public?

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+        &mut self,
+        target: ExtensionType,
+    ) -> Result<&mut [u8], ProgramError> {
+        validate_extension_account_type(target, B::ACCOUNT_TYPE)?;
+        get_extension_bytes_from_tlv_mut(self.tlv_data_mut(), target)
+    }
+
+    #[inline]
+    pub fn get_extension_types(&self, out: &mut [ExtensionType]) -> Result<usize, ProgramError> {
```
febo: nit: same here about the get_, it is more like a write_. We could also have a get_extension_types that requires alloc::vec::Vec.

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+    }
+}
+
+#[inline(always)]
+fn extension_data_start<B: ExtensionBaseState>(data_len: usize) -> usize {
+    if data_len == B::BASE_LEN {
+        data_len
+    } else {
+        TLV_START_INDEX
+    }
```
febo: I found the use of this function a bit confusing:
```
let data_start = extension_data_start::<B>(B::BASE_LEN + self.data.len()) - B::BASE_LEN;
```
So you add `B::BASE_LEN` and then subtract from the result.

I wonder if we could simplify the implementation to do:
```suggestion
    core::cmp::min(data_len, TLV_START_INDEX - B::BASE_LEN)
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+    #[inline(always)]
+    unsafe fn from_bytes_unchecked(data: &[u8]) -> &Self {
+        debug_assert!(data.len() >= B::BASE_LEN);
+        let ptr = core::ptr::slice_from_raw_parts(data.as_ptr(), data.len() - B::BASE_LEN);
+        &*(ptr as *const Self)
+    }
+
+    #[inline(always)]
+    fn tlv_data(&self) -> &[u8] {
+        let data_start = extension_data_start::<B>(B::BASE_LEN + self.data.len()) - B::BASE_LEN;
```
febo: If we are going with https://github.com/anza-xyz/pinocchio/pull/349/changes#r3307828280, then this needs to be:
```suggestion
        let data_start = extension_data_start::<B>(self.data.len());
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+    #[inline(always)]
+    unsafe fn from_bytes_mut_unchecked(data: &mut [u8]) -> &mut Self {
+        debug_assert!(data.len() >= B::BASE_LEN);
+        let ptr = core::ptr::slice_from_raw_parts_mut(data.as_mut_ptr(), data.len() - B::BASE_LEN);
+        &mut *(ptr as *mut Self)
+    }
+
+    #[inline(always)]
+    fn tlv_data(&self) -> &[u8] {
+        let data_start = extension_data_start::<B>(B::BASE_LEN + self.data.len()) - B::BASE_LEN;
```
febo: If we are going with https://github.com/anza-xyz/pinocchio/pull/349/changes#r3307828280, then this needs to be:
```suggestion
        let data_start = extension_data_start::<B>(self.data.len());
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+        &self.data[data_start..]
+    }
+
+    #[inline]
+    fn tlv_data_mut(&mut self) -> &mut [u8] {
+        let data_start = extension_data_start::<B>(B::BASE_LEN + self.data.len()) - B::BASE_LEN;
```
febo: If we are going with https://github.com/anza-xyz/pinocchio/pull/349/changes#r3307828280, then this needs to be:
```suggestion
        let data_start = extension_data_start::<B>(self.data.len());
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+        extension_from_bytes_mut(bytes)
+    }
+
+    /// # Safety
+    ///
+    /// `data` must have passed `validate_state_with_extensions_data::<B>`.
+    #[inline(always)]
+    unsafe fn from_bytes_mut_unchecked(data: &mut [u8]) -> &mut Self {
+        debug_assert!(data.len() >= B::BASE_LEN);
+        let ptr = core::ptr::slice_from_raw_parts_mut(data.as_mut_ptr(), data.len() - B::BASE_LEN);
```
febo, line 410: As discussed on slack, let's be a bit defensive here:
```suggestion
        let ptr = core::ptr::slice_from_raw_parts_mut(data.as_mut_ptr(), data.len().saturating_sub(B::BASE_LEN));
```

programs/token-2022/src/state/extension/state.rs
```diff
@@ -0,0 +1,1285 @@
...
+        Self::from_bytes(account_view.borrow_unchecked())
+    }
+
+    /// # Safety
+    ///
+    /// `data` must have passed `validate_state_with_extensions_data::<B>`.
+    #[inline(always)]
+    unsafe fn from_bytes_unchecked(data: &[u8]) -> &Self {
+        debug_assert!(data.len() >= B::BASE_LEN);
+        let ptr = core::ptr::slice_from_raw_parts(data.as_ptr(), data.len() - B::BASE_LEN);
```
febo, line 283: As discussed on slack, let's be a bit defensive here:
```suggestion
        let ptr = core::ptr::slice_from_raw_parts(data.as_ptr(), data.len().saturating_sub(B::BASE_LEN));
```

febo (commented):
I think these are the final comments/suggestions.

programs/token-2022/src/state/extension/state.rs
dev-jodee: Yeah you're right, mint base len < base acc len and acc base len == base acc len so can remove

programs/token-2022/src/state/extension/state.rs
dev-jodee: no reason for it to be pub

programs/token-2022/src/state/extension/state.rs
dev-jodee: no

programs/token-2022/src/state/extension/state.rs
dev-jodee: I'd prefer not to have to use alloc, wdyt

febo: Yeah, we could leave it out for now.

programs/token-2022/src/state/extension/state.rs
febo: Given that this function in private and always used after checking the length, let's keep it as it is.

programs/token-2022/src/state/extension/state.rs
febo: Same here, let's keep it as it is.

febo (approved):
Looks great – thanks!

grod220:
Thanks for this! This work will be used directly in p-ATA.

programs/token-2022/src/state/extension/mod.rs
dev-jodee: made a fast follow to fix this before I forget https://github.com/anza-xyz/pinocchio/pull/418

PR #360, opened 2026-03-05
sdk: Improve `SlotHashes` entry count validation

febo:
### Problem

The safe constructor `SlotHashes::new()` calculates the require length using unchecked arithmetic, which could overflow when the number of entries written into the byte slice is incorrectly greater than `MAX_ENTRIES`.

### Solution

Add an explicit check that a valid number of entries was read:
```rust
if num_entries > MAX_ENTRIES {
}
```

cc: @nbelenkov

### DIFF
--- a/sdk/src/sysvars/slot_hashes/mod.rs
+++ b/sdk/src/sysvars/slot_hashes/mod.rs
@@ -14,7 +14,7 @@ mod test_raw;
 mod test_utils;
 
 #[cfg(feature = "alloc")]
-use alloc::{boxed::Box, vec::Vec};
+use alloc::boxed::Box;
 use {
     crate::{
         account::{AccountView, Ref},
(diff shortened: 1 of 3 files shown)


sdk/src/sysvars/slot_hashes/mod.rs
```diff
@@ -151,11 +113,27 @@ impl<T: Deref<Target = [u8]>> SlotHashes<T> {
...
+        if data.len() < NUM_ENTRIES_SIZE {
+            return Err(ProgramError::AccountDataTooSmall);
+        }
+
+        // SAFETY: `data` is guaranteed to have at least `NUM_ENTRIES_SIZE` bytes.
+        let num_entries = unsafe { get_entry_count(data.as_ref()) };
+
+        if num_entries > MAX_ENTRIES {
+            return Err(ProgramError::InvalidArgument);
+        }
```
febo, line 125: This is the main addition. With this assert, the unchecked arithmetic cannot overflow.

sdk/src/sysvars/slot_hashes/mod.rs
```diff
@@ -304,55 +276,34 @@ impl<'a> SlotHashes<Ref<'a, [u8]>> {
...
+        let sysvar_data = account_view.try_borrow()?;
 
         // SAFETY: The account was validated to be the `SlotHashes` sysvar.
-        Ok(unsafe { SlotHashes::new_unchecked(data_ref) })
+        Ok(unsafe { SlotHashes::new_unchecked(sysvar_data) })
     }
 }
 
 #[cfg(feature = "alloc")]
 impl SlotHashes<Box<[u8]>> {
```
febo, line 287: The changes on this `impl` are simply simplifications of the code.

joncinque (approved):
Very nice! Big fan of the uninit box slice again

PR #363, opened 2026-03-06
Bump dependencies

febo:
### Problem

There are new versions of solana-account-view and solana-instruction-view, which require updating the entrypoint signature and `invoke_signed_with_bounds` call sites.

### Solution

Bump the dependencies versions and update the required code.

### DIFF
--- a/programs/token-2022/src/instructions/extensions/cpi_guard/disable.rs
+++ b/programs/token-2022/src/instructions/extensions/cpi_guard/disable.rs
@@ -115,7 +115,7 @@ impl<'a, 'b, 'c> Disable<'a, 'b, 'c> {
             account.write(signer);
         }
 
-        invoke_signed_with_bounds::<{ 2 + MAX_MULTISIG_SIGNERS }, &AccountView>(
+        invoke_signed_with_bounds::<{ 2 + MAX_MULTISIG_SIGNERS }, _>(
             &InstructionView {
                 program_id: self.token_program,
                 // SAFETY: instruction accounts has `expected_accounts` initialized.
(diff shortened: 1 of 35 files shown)


programs/system/src/lib.rs
```diff
@@ -81,6 +81,8 @@ pub fn create_account_with_minimum_balance_signed(
         //
         // SAFETY: There are no active borrows of the `account`.
         // This was checked by the `Assign` CPI above.
-        unsafe { account.resize_unchecked(space) }
+        // TODO: re-enable this when `Reize` trait is available.
+        //unsafe { account.resize_unchecked(space) }
```
febo, line 85: The resize change will be done in a follow-up PR, so it is commented out for now.

programs/token-2022/src/instructions/extensions/cpi_guard/disable.rs
```diff
@@ -115,7 +115,7 @@ impl<'a, 'b, 'c> Disable<'a, 'b, 'c> {
             account.write(signer);
         }
 
-        invoke_signed_with_bounds::<{ 2 + MAX_MULTISIG_SIGNERS }>(
+        invoke_signed_with_bounds::<{ 2 + MAX_MULTISIG_SIGNERS }, &AccountView>(
```
joncinque: Is it possible to use a `_` here like in the memo program? Probably easier to not specify since it shouldn't be needed

febo: Good call! I made it consistent. When we specify the type on the parameter, then we can use a `_`. So I applied this pattern in all of them.

joncinque: Ok great, thanks!

PR #364, opened 2026-03-06
More `&mut` accounts updates

febo:
### Problem

PR #363 bumped the account-view dependency, which requires using `&mut` references for the accounts slice. There are many places in the documentation that have not been updated. Additionally, the `&mut` slice is being created from the `*const` pointer instead of `*mut` one.

### Solution

Update the documentation and code to use `&mut` references.

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -205,7 +205,7 @@ pub unsafe fn process_entrypoint<const MAX_ACCOUNTS: usize>(
     // they are initialized so we cast the pointer to a slice of `[AccountView]`.
     match process_instruction(
         program_id,
-        unsafe { from_raw_parts_mut(accounts.as_ptr() as _, count) },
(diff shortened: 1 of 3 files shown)


sdk/src/entrypoint/mod.rs
```diff
@@ -205,7 +205,7 @@ pub unsafe fn process_entrypoint<const MAX_ACCOUNTS: usize>(
     // they are initialized so we cast the pointer to a slice of `[AccountView]`.
     match process_instruction(
         program_id,
-        unsafe { from_raw_parts_mut(accounts.as_ptr() as _, count) },
+        unsafe { from_raw_parts_mut(accounts.as_mut_ptr() as _, count) },
```
febo, line 208: The compiler did not complain about this one.

joncinque: because it's just getting cast anyway?

febo: Yep!

PR #365, opened 2026-03-06
programs: Use `AsRef<AccountView>` for signers

febo:
### Problem

Instruction helpers that require multisig signers currently specify them as `&[&AccountView]`. This has the drawback that if signers are extracted from the accounts slice as a `[AccountView]` slice, then they need to be mapped to a `[&AccountView]` slice.

### Solution

This PR add a generic type `A: AsRef<AccountView>` to specify signers. With this change, signers can be specific either as `&[&AccountView]` or `&[AccountView]`. This is particularly useful when signers are passed as "remaining" accounts – e.g.:
```rust
let [from, to, authority, signers @ ..] = accounts else {
};

Transfer {
    from,
    to,
    authority,
    multisig_signers: signers,
    amount: 1_000_000_000,
}
.invoke()?;
```
The above example does not compile with the current version, but it will with the changes in this PR.

### DIFF
--- a/programs/token-2022/src/instructions/extensions/transfer_fee/withdraw_withheld_tokens_from_accounts.rs
+++ b/programs/token-2022/src/instructions/extensions/transfer_fee/withdraw_withheld_tokens_from_accounts.rs
@@ -28,7 +28,13 @@ use {
 ///   2. `[]` The mint's multisig `withdraw_withheld_authority`.
 ///   3. `..3+M` `[signer]` M signer accounts.
 ///   4. `3+M+1..3+M+N` `[writable]` The source accounts to withdraw from.
-pub struct WithdrawWithheldTokensFromAccounts<'a, 'b, 'c, A: AsRef<AccountView>> {
+pub struct WithdrawWithheldTokensFromAccounts<
+    'a,
+    'b,
+    'c,
+    MultisigSigner: AsRef<AccountView>,
+    Source: AsRef<AccountView>,
+> {
     /// The token mint.
     pub mint: &'a AccountView,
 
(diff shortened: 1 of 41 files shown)


programs/token-2022/src/instructions/extensions/transfer_fee/withdraw_withheld_tokens_from_accounts.rs
```diff
@@ -39,16 +39,16 @@ pub struct WithdrawWithheldTokensFromAccounts<'a, 'b, 'c> {
     pub authority: &'a AccountView,
 
     /// Multisignature signer accounts.
-    pub multisig_signers: &'c [&'a AccountView],
+    pub multisig_signers: &'c [A],
 
     /// Source accounts to withdraw from.
-    pub sources: &'c [&'a AccountView],
+    pub sources: &'c [A],
```
joncinque: I think this is fine, and can be addressed in a follow-up, but is there a case where we would mix two different types that both implement `AsRef<AccountView>`?

febo: Good point. We could add a separate generic. Let me add one so we can see how it would look like.

febo: Added another generic:
```rust
pub struct WithdrawWithheldTokensFromAccounts<
    'a,
    'b,
    'c,
    MultisigSigner: AsRef<AccountView>,
    Source: AsRef<AccountView>,
> {
...
}

PR #366, opened 2026-03-07
add decode as default features

AkliluYirgalem:
Problem:

`pinocchio` depends on `solana-address`, but it does not re-export the `decode` feature as default nor as optional. Because of this, decoding-related functionality from `solana-address`, such as `Address::from_str_const` is not available when using `pinocchio`.

Solution:

Expose the `decode` feature in `pinocchio` and forward it to `solana-address`, and include it in the default features.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -25,12 +25,15 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 alloc = ["solana-instruction-view?/slice-cpi"]
 copy = ["solana-account-view/copy", "solana-address/copy"]
 cpi = ["dep:solana-instruction-view"]
-default = ["alloc"]
+decode = ["solana-address/decode"]
+default = ["alloc", "decode"]
 
 [dependencies]
 solana-account-view = { workspace = true }
 solana-address = { workspace = true, features = ["syscalls"] }
-solana-instruction-view = { workspace = true, features = ["cpi"], optional = true }
+solana-instruction-view = { workspace = true, features = [
+    "cpi",
+], optional = true }
 solana-program-error = { workspace = true }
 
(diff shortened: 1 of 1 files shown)


febo:
You can add solana-address dependency with the `"decode"` feature to enabled it in your program. This is a preferred approach than exposing every single feature of dependencies in pinocchio.

```toml
solana-address = { version = "2.2", features = ["decode"] }
```

AkliluYirgalem:
> You can add solana-address dependency with the `"decode"` feature to enabled it in your program. This is a preferred approach than exposing every single feature of dependencies in pinocchio.
> 
> ```toml
> solana-address = { version = "2.2", features = ["decode"] }
> ```

interesting, but how about optionally than default

febo:
> > You can add solana-address dependency with the `"decode"` feature to enabled it in your program. This is a preferred approach than exposing every single feature of dependencies in pinocchio.
> > ```toml
> > solana-address = { version = "2.2", features = ["decode"] }
> > ```
> 
> interesting, but how about optionally than default

Same issue, if we start doing that where do we stop? We will end up with a salad of features in pinocchio, which will be difficult to trace what they do because they are not directly related to pinocchio.

AkliluYirgalem:
> > > You can add solana-address dependency with the `"decode"` feature to enabled it in your program. This is a preferred approach than exposing every single feature of dependencies in pinocchio.
> > > ```toml
> > > solana-address = { version = "2.2", features = ["decode"] }
> > > ```
> > 
> > 
> > interesting, but how about optionally than default
> 
> Same issue, if we start doing that where do we stop? We will end up with a salad of features in pinocchio, which will be difficult to trace what they do because they are not directly related to pinocchio.

you're totally right. I just did some refactoring on a side project and I see what you mean now!

PR #368, opened 2026-03-11
token, token-2022: Fix consistency in instructions

LStan:
1. Changed `&AccountView` to `_` in places missed in #363.
2. While doing that, I noticed that `from_raw_parts` is called differently in different instructions: in some `from_raw_parts`, in some `slice::from_raw_parts`, and in some both variants are used. Changed to `from_raw_parts` everywhere.

There are also `core::slice::from_raw_parts` in `Memo` and `sdk`, but I'm not sure if they need to be fixed.


LStan:
There is also inconsistency in creating `InstructionView`: in some places there is a variable `instruction`, and in some places it is inlined in `invoke*` calls. Should it be fixed?

PR #373, opened 2026-03-17
sdk: Remove unnecessary check in bump allocator

febo:
### Problem

Currently the bump allocator includes a test to check whether it is the first allocation or not. This can be avoided if we reserve the first `8` bytes at the start to store the current position.

### Solution

Remove the unnecessary check and reserve the first `8` bytes at the creation of the allocator.

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -656,7 +656,6 @@ mod alloc {
         crate::{entrypoint::MAX_HEAP_LENGTH, hint::unlikely},
         core::{
             alloc::{GlobalAlloc, Layout},
-            mem::size_of,
             ptr::null_mut,
         },
     };
@@ -696,6 +695,13 @@ mod alloc {
         /// runtime. The runtime will enforce the actual heap size
         /// requested by the program.
         pub const unsafe fn new_unchecked(start: usize, len: usize) -> Self {
+            if len < size_of::<usize>() {
+                panic!("heap length must be at least `size_of::<usize>()` bytes");
+            }
+            // Reserves the first `size_of::<usize>()` bytes to store the current position
+            // of the heap pointer.
+            *(start as *mut usize) = start + size_of::<usize>();
+
             Self {
                 start,
                 end: start + len,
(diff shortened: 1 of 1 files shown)


febo:
Since the allocator can only work if `len >= size_of<usize>()`, this change actually validates that the `len` value is valid.

sdk/src/entrypoint/mod.rs
```diff
@@ -764,6 +695,13 @@ mod alloc {
         /// runtime. The runtime will enforce the actual heap size
         /// requested by the program.
         pub const unsafe fn new_unchecked(start: usize, len: usize) -> Self {
+            if len < size_of::<usize>() {
+                panic!("heap length must be at least `size_of::<usize>()` bytes");
+            }
+            // Reserves the first `size_of::<usize>()` bytes to store the current position
+            // of the heap pointer.
+            *(start as *mut usize) = start + size_of::<usize>();
```
grod220, line 703: How can this be done at compile time? Doesn't that address only exist at runtime?

grod220: Probably not a likely scenario, but if new_unchecked() is ever called twice over the same heap, it resets the cursor and the next allocation overlaps the previous one.

joncinque: Yeah this doesn't work at compile time, since there's no address to write to -- in fact, it's dangerous to do this at compile-time and expect something to happen!

However, it's fine to do at runtime, but I'm not sure when allocator initialization happens during execution of a program

febo: Hmmm, right, I was overly optimistic since the way the tests are setup up, this was allowed. I will refactor it so we can see if makes sense to do this change – I am starting to thing that might be better to leave it as it is.

febo:
If we move the `*(start as *mut usize) = start + size_of::<usize>()` to the entrypoint, then it works, but it sort of decouples the initialization of the allocator and creates a dependency between two different macros. It would  make sense if we refactor the macros and, instead of requiring them to be called, make use of features to choose the allocator macro. That way when the bump allocator feature is enabled, we can also generate the necessary initialization.

I don't think it is worth doing it to save `1` CU per allocation because it would make harder for people to write their own entrypoint. Maybe when SIMD-0449 is active and the entrypoint is super simple, we can incorporate this.

I will close this for now.

PR #376, opened 2026-03-19
sdk: Add inline program entrypoint

febo:
### Problem

With the change proposed in SIMD-0321, the runtime will pass a pointer to the instruction data. This will allow programs to determine what instruction to execute before parsing the account data. Entrypoints can take advantage of this information to offer custom parsing logic, which is optimized by the number of accounts required by an instruction.

### Solution

Add a new entrypoint that makes use of the instruction data pointer. The rationale is that a program will inspect the instruction data to determine how many accounts and instruction to execute.

The `inline_program_entrypoint` macro is used to declare the entrypoint. The
only argument is the name of a function with this type signature:
```rust
fn process_instruction(input: ProgramInput) -> ProgramResult;
```
The `execute` macro takes three arguments: the number of accounts to parse from the input buffer, the program input, and the name of the processor function. The processor function has this type signature:
```rust
fn processor(
    program_id: &Address,
    accounts: &mut [AccountView],
    instruction_data: &[u8]
) -> ProgramResult;
```

#### Example

The example below consumes `30` CUs less than a similar program using the "standard" entrypoint.

```rust
use {
    pinocchio::{
        entrypoint::inline::ProgramInput, error::ProgramError, execute, inline_program_entrypoint,
        AccountView, Address, ProgramResult,
    },
    pinocchio_system::instructions::CreateAccount,
};

// Declares the entrypoint of the program.
inline_program_entrypoint!(process_instruction);

pub fn process_instruction(input: ProgramInput) -> ProgramResult {
    match input.data.first() {
        Some(&0) => execute!((3, input) => create),
        _ => Err(ProgramError::InvalidInstructionData),
    }
}

/// Instruction processor.
pub fn create(
    program_id: &Address,
    accounts: &mut [AccountView],
    _instruction_data: &[u8],
) -> ProgramResult {
    let [from, to, _system_program] = accounts else {
    };

    CreateAccount {
        from,
        to,
        lamports: 1_000_000_000,
        space: 10,
        owner: program_id,
    }
    .invoke()
}
```

### DIFF
--- a/sdk/src/entrypoint/inline.rs
+++ b/sdk/src/entrypoint/inline.rs
@@ -16,25 +16,27 @@ pub const MAX_INLINE_ACCOUNTS: usize = 10;
 /// This allows the entrypoint to read the instruction data and decide how many
 /// accounts it wants to parse from the program input.
 ///
-/// It offers two macros to declare the entrypoint: [`inline_program_entrypoint`]
-/// and [`execute`]. The former is used to declare the entrypoint and the
-/// latter is used to execute the program logic with a specified number of
-/// accounts and a processor function. The processor function is called with
-/// the program id, the accounts, and the instruction data.
+/// It offers two macros to declare the entrypoint:
+/// [`crate::inline_program_entrypoint!`] and [`crate::execute!`]. The former is
+/// used to declare the entrypoint and the latter is used to execute the program
+/// logic with a specified number of accounts and a processor function. The
+/// processor function is called with the program id, the accounts, and the
+/// instruction data.
 ///
-/// The [`inline_program_entrypoint`] macro is used to declare the entrypoint. The
-/// only argument is the name of a function with this type signature:
+/// The [`crate::inline_program_entrypoint!`] macro is used to declare the
+/// entrypoint. The only argument is the name of a function with this type
+/// signature:
 ///
 /// ```ignore
 /// fn process_instruction(input: ProgramInput) -> ProgramResult;
 /// ```
 ///
 /// [`ProgramInput`] offers a method to read the instruction data. Programs can
-/// then use the [`execute`] macro to execute the program logic for the
-/// corresponding instruction. The [`execute`] macro takes three arguments: the
-/// number of accounts to parse from the input buffer, the program input, and
-/// the name of the processor function. The processor function has this type
-/// signature:
+/// then use the [`crate::execute!`] macro to execute the program logic for the
+/// corresponding instruction. The [`crate::execute!`] macro takes three
+/// arguments: the number of accounts to parse from the input buffer, the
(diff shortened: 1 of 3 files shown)


sdk/src/entrypoint/inline.rs
```diff
@@ -0,0 +1,266 @@
+//! Defines the inline program entrypoint and associated types.
+
+use {core::slice::from_raw_parts, solana_address::Address};
+
+/// Maximum number of accounts that can be parsed inline in the entrypoint.
+///
+/// This is a hard limit to constrain the size of the parsing logic
+/// inlined in the entrypoint.
+pub const MAX_INLINE_ACCOUNTS: usize = 10;
```
febo, line 9: This is an arbitrary limit, it can be higher or lower. The trade-off here is the size of the binary, since the parsing will be inlined (no loops).

sdk/src/entrypoint/mod.rs
```diff
@@ -270,31 +270,52 @@ pub unsafe fn process_entrypoint<const MAX_ACCOUNTS: usize>(
 }
 
 /// Align a pointer to the BPF alignment of [`u128`].
-macro_rules! align_pointer {
+#[macro_export]
+macro_rules! __align_pointer {
```
febo, line 274: There was a need to export these "internal" macros, so I followed the pattern of adding `__` at the start of the name. They are not meant to be used by programs directly.

sdk/src/entrypoint/inline.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// Return the address of the program.
+    pub fn program_id(&self) -> &'static Address {
+        // SAFETY: The program id is located at the end of the program input buffer
+        // serialized by the SVM loader after the instruction data.
+        unsafe { &*self.data.as_ptr().add(self.data.len()).cast::<Address>() }
+    }
+}
+
+#[macro_export]
+macro_rules! execute {
```
febo, line 167: Currently there is no API to get remaining accounts – accounts beyond the number specified on `execute`. We could add something similar to the lazy entrypoint.

joncinque:
We discussed this a bit offline, and considering the new entrypoint requires making a lot of internal macros public, and will only be useful until SIMD-0449 is enabled, I lean towards keeping this as a draft (edit) and not merging it. People can fork Pinocchio if they truly want it

sdk/src/entrypoint/inline.rs
BretasArthur1: I remember to bench something like this here: https://github.com/anza-xyz/pinocchio/pull/319

and if I'm not mistaking we have a limit of accounts that can be inlined to avoid overflow

febo:
#421 replaces the need for this.

PR #377, opened 2026-03-23
token: Add `Batch` instruction

febo:
### Problem

[SIMD-0266](https://github.com/febo/solana-improvement-documents/blob/main/proposals/0266-efficient-token-program.md) was activated on testnet, so soon the token program will support [`Batch`](https://github.com/solana-program/token/blob/main/pinocchio/interface/src/instruction.rs#L532-L552) instructions. This significantly reduces the cost of executing multiple token instruction via cross-program invocations (CPIs).

### Solution

Add support for `Batch` instructions. To facilitate the implementation, 2 traits are introduced:
- `CpiWriter`: this trait is implemented by all instructions (apart from `Batch`) to write their CPI information (accounts and instruction data) to existing buffers.
- `IntoBatch`: this trait is implemented by all instructions (apart from `Batch`) to indicate that the instruction can be added to a `Batch`.

The `Batch` instruction works over `IntoBatch` instructions and prepares all accounts and instruction data for the cross-program invocation.
```rust
let mut batch_state = BatchState::new(Batch::MAX_ACCOUNTS_LEN, Batch::MAX_DATA_LEN);
let mut batch = batch_state.as_batch()?;

InitializeMint2 {
    mint,
    mint_authority: authority.address(),
    decimals: 9,
    freeze_authority: None,
}
.into_batch(&mut batch)?;

InitializeAccount3 {
    account,
    mint,
    owner: authority.address(),
}
.into_batch(&mut batch)?;

batch.invoke()?;
```

An example program using the `Batch` instruction is [here](https://github.com/febo/playground/blob/token-batch/programs/pinocchio/token/src/entrypoint.rs).

### DIFF
--- a/programs/token/src/instructions/amount_to_ui_amount.rs
+++ b/programs/token/src/instructions/amount_to_ui_amount.rs
@@ -54,10 +54,10 @@ impl<'account> AmountToUiAmount<'account> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; AmountToUiAmount::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; AmountToUiAmount::DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
--- a/programs/token/src/instructions/get_account_data_size.rs
+++ b/programs/token/src/instructions/get_account_data_size.rs
@@ -49,10 +49,10 @@ impl<'account> GetAccountDataSize<'account> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; GetAccountDataSize::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; GetAccountDataSize::DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
--- a/programs/token/src/instructions/initialize_account.rs
+++ b/programs/token/src/instructions/initialize_account.rs
@@ -76,10 +76,10 @@ impl<'account> InitializeAccount<'account> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; InitializeAccount::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; InitializeAccount::DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
--- a/programs/token/src/instructions/initialize_account_2.rs
+++ b/programs/token/src/instructions/initialize_account_2.rs
@@ -71,10 +71,10 @@ impl<'account> InitializeAccount2<'account> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; InitializeAccount2::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; InitializeAccount2::DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
--- a/programs/token/src/instructions/initialize_account_3.rs
+++ b/programs/token/src/instructions/initialize_account_3.rs
@@ -63,10 +63,10 @@ impl<'account, 'address> InitializeAccount3<'account, 'address> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; InitializeAccount3::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; InitializeAccount3::DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
--- a/programs/token/src/instructions/initialize_immutable_owner.rs
+++ b/programs/token/src/instructions/initialize_immutable_owner.rs
@@ -47,10 +47,10 @@ impl<'account> InitializeImmutableOwner<'account> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; InitializeImmutableOwner::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; InitializeImmutableOwner::DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
--- a/programs/token/src/instructions/initialize_mint.rs
+++ b/programs/token/src/instructions/initialize_mint.rs
@@ -79,10 +79,10 @@ impl<'account, 'address> InitializeMint<'account, 'address> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; InitializeMint::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; InitializeMint::MAX_DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::MAX_DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
--- a/programs/token/src/instructions/initialize_mint_2.rs
+++ b/programs/token/src/instructions/initialize_mint_2.rs
@@ -68,10 +68,10 @@ impl<'account, 'address> InitializeMint2<'account, 'address> {
         let written_instruction_accounts =
             self.write_instruction_accounts(&mut instruction_accounts)?;
 
-        let mut accounts = [UNINIT_CPI_ACCOUNT; InitializeMint2::ACCOUNTS_LEN];
+        let mut accounts = [UNINIT_CPI_ACCOUNT; Self::ACCOUNTS_LEN];
         let written_accounts = self.write_accounts(&mut accounts)?;
 
-        let mut instruction_data = [UNINIT_BYTE; InitializeMint2::MAX_DATA_LEN];
+        let mut instruction_data = [UNINIT_BYTE; Self::MAX_DATA_LEN];
         let written_instruction_data = self.write_instruction_data(&mut instruction_data)?;
 
         unsafe {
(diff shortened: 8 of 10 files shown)


programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        Self {
+            data,
+            instruction_accounts: Box::new_uninit_slice(MAX_CPI_ACCOUNTS),
+            accounts: Box::new_uninit_slice(MAX_CPI_ACCOUNTS),
+            data_len: 1,
+            accounts_len: 0,
+            instruction_accounts_len: 0,
+        }
+    }
```
febo: Currently working with the assumption that `alloc` will be available. We could add a lower default that fit the stack limit using generics.

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,131 @@
...
+        Ok(())
+    }
+
+    pub fn append(&mut self, instructions: &[&'a dyn Batchable]) -> ProgramResult {
+        for instruction in instructions {
+            self.push(*instruction)?;
+        }
+
+        Ok(())
+    }
```
febo: I am considering removing this since it is worse in terms of CU because of the dynamic dispatch.

programs/token/src/lib.rs
```diff
@@ -1,14 +1,29 @@
...
+#[cfg(feature = "batch")]
+extern crate alloc;
+
+//mod error;
+//mod instruction;
 pub mod instructions;
 pub mod state;
 
-use core::mem::MaybeUninit;
+//pub use {error::*, instruction::*};
```
grod220: nit: remove

febo:
@joncinque @grod220 Had an idea of a refactoring, so putting this back to draft.

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,144 @@
...
+            accounts: Box::new_uninit_slice(MAX_CPI_ACCOUNTS),
+            data_len: 1,
+            accounts_len: 0,
+            instruction_accounts_len: 0,
+        }
+    }
+
+    #[inline(always)]
+    pub fn push<T: Batchable + ?Sized>(&mut self, instruction: &'a T) -> ProgramResult {
+        self.accounts_len += instruction.write_accounts(&mut self.accounts[self.accounts_len..])?;
```
grod220: Could incrementing accounts_len before later fallible writes leave these cursors out of sync?

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        let instruction_accounts_len = instruction.write_instruction_accounts(
+            &mut self.instruction_accounts[self.instruction_accounts_len..],
+        )?;
+
+        let data_len = instruction.write_instruction_data(&mut self.data[self.data_len + 2..])?;
+
+        self.data[self.data_len].write(instruction_accounts_len as u8);
+        self.data[self.data_len + 1].write(data_len as u8);
```
grod220: The u8 casts will truncate values over 255. We need a bound check here right given an instruction with more bytes of data?

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    #[inline(always)]
+    pub fn push<T: Batchable + ?Sized>(&mut self, instruction: &'a T) -> ProgramResult {
+        self.accounts_len += instruction.write_accounts(&mut self.accounts[self.accounts_len..])?;
+
+        let instruction_accounts_len = instruction.write_instruction_accounts(
+            &mut self.instruction_accounts[self.instruction_accounts_len..],
+        )?;
+
+        let data_len = instruction.write_instruction_data(&mut self.data[self.data_len + 2..])?;
```
grod220: A guard here too maybe?
```rust
 if self.data_len + 2 > self.data.len() {
      return Err(ProgramError::InvalidArgument.into());
  }
```

programs/token/src/instructions/batch.rs
febo: Removed due to the latest refactoring.

programs/token/src/instructions/batch.rs
febo: Good point! Move the increment to the end.

programs/token/src/instructions/batch.rs
febo: No instruction writes more than `u8::MAX` data, so I think we are ok here.

febo:
@joncinque @grod220 It is <del>not</del> now ready for review. It includes a refactoring so the same logic can be used for `invoke` and `into_batch`.

programs/token/src/instructions/mod.rs
```diff
@@ -32,3 +36,66 @@ pub use {
...
+#[inline(always)]
+fn cpi_account<'cpi>(account: &AccountView) -> Result<CpiAccount<'cpi>, ProgramError> {
+    if account.as_ref().is_borrowed_mut() {
+        return Err(account_borrow_failed_error());
+    }
+    Ok(CpiAccount::from(account))
+}
+
+/// A trait for instructions that can be used in a CPI context.
+pub trait CpiWriter {
```
grod220, line 53: Should this be pub(crate)?

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,156 @@
...
+        let written_data = write_data(&mut self.data[self.data_len + IX_HEADER_SIZE..])?;
+
+        let written_accounts = write_accounts(&mut self.accounts[self.accounts_len..])?;
+
+        let written_instruction_accounts = write_instruction_accounts(
+            &mut self.instruction_accounts[self.instruction_accounts_len..],
+        )?;
+
+        // If all writres succeeded, update the lengths and write the instruction
+        // header.
```
grod220:
```diff
- // If all writres succeeded
+ // If all writes succeeded
```

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        self.accounts_len += written_accounts;
+        self.instruction_accounts_len += written_instruction_accounts;
+
+        self.data[self.data_len].write(written_instruction_accounts as u8);
+        self.data[self.data_len + 1].write(written_data as u8);
```
grod220, line 141: Looks like a truncation could happen through `UiAmountToAmount`? 

```rust
let amount = "1".repeat(u8::MAX as usize); // this + discriminator is 256
UiAmountToAmount::<255>::new(&mint, &amount)
   .into_batch(&mut batch)
   .unwrap();
```

programs/token/src/instructions/close_account.rs
```diff
@@ -72,57 +101,188 @@ impl<'a, 'b, MultisigSigner: AsRef<AccountView>> CloseAccount<'a, 'b, MultisigSi
...
+                    self.destination,
+                    self.authority,
+                    self.multisig_signers,
+                    accounts,
+                )
+            },
+            write_instruction_data,
+        )
+    }
+}
```
grod220, line 204: Wonder if this closure glue code can be eliminated if push() took &impl CpiWriter directly instead of three closures. It could then call the three write methods itself and then each instruction's into_batch would collapse to just batch.push(&self).

programs/token/src/instructions/transfer.rs
```diff
@@ -1,62 +1,95 @@
...
     solana_program_error::{ProgramError, ProgramResult},
 };
 
-/// Transfer tokens from one Token account to another.
+/// Maximum number of accounts expected by this instruction.
+///
+/// The required number of accounts will depend whether the
+/// source account has a single owner or a multisignature
+/// owner.
+const MAX_ACCOUNTS_LEN: usize = 4 + MAX_MULTISIG_SIGNERS;
```
grod220: Saw `3 + multisig_signers.len()` used below. Is this off by one?

programs/token/src/instructions/thaw_account.rs
```diff
@@ -1,57 +1,85 @@
...
+
+/// Thaw a Frozen account using the Mint's `freeze_authority` (if set).
 ///
-/// ### Accounts:
-///   * Single freeze authority
-///   0. `[WRITE]` The account to thaw.
+/// Accounts expected by this instruction:
+///
+///   * Single owner
+///   0. `[writable]` The account to freeze.
```
grod220:
```diff
- ///   0. `[writable]` The account to freeze.
+ ///   0. `[writable]` The account to thaw.
```

programs/token/src/instructions/thaw_account.rs
```diff
@@ -1,57 +1,85 @@
...
 ///   1. `[]` The token mint.
-///   2. `[SIGNER]` The mint freeze authority.
+///   2. `[signer]` The mint freeze authority.
 ///
-///   * Multisignature freeze authority
-///   0. `[WRITE]` The account to thaw.
+///   * Multisignature owner
+///   0. `[writable]` The account to freeze.
```
grod220:
```diff
- ///   0. `[writable]` The account to freeze.
+ ///   0. `[writable]` The account to thaw.
```

programs/token/src/instructions/thaw_account.rs
```diff
@@ -1,57 +1,85 @@
...
-pub struct ThawAccount<'a, 'b, MultisigSigner: AsRef<AccountView>> {
-    /// Token Account to thaw.
-    pub account: &'a AccountView,
-    /// Mint Account.
-    pub mint: &'a AccountView,
-    /// Mint Freeze Authority Account
-    pub freeze_authority: &'a AccountView,
+///   3. `..+M` `[signer]` M signer accounts.
+pub struct ThawAccount<'account, 'multisig, MultisigSigner: AsRef<AccountView>> {
+    ///  The account to freeze.
```
grod220:
```diff
-    ///  The account to freeze.
+    ///  The account to thaw.
```

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/// Implement `Sealed` for all types that implement `CpiWriter`.
+impl<T: CpiWriter> sealed::Sealed for T {}
+
+/// A module only accessible within this crate that contains the
+/// `Sealed` trait.
+pub(crate) mod sealed {
+    /// A sealed trait that prevents external implementations of the
+    /// `Batchable` trait.
```
grod220:
```diff
- /// `Batchable` trait.
+ /// `IntoBatch` trait.
```

programs/token/src/instructions/mod.rs
febo: The idea of keeping it public is that if someone wants to reuse the buffer arrays they could. For example, you can allocate the buffers in your program once and use the `CpiWriter` to fill them in and invoke the CPI "manually", reusing the same buffers if you are performing multiple CPIs.

programs/token/src/instructions/transfer.rs
febo: Yep, fixed.

programs/token/src/instructions/batch.rs
febo: Hmmm, good point. Added an explicit test against a `MAX_DATA_LEN` constant.

programs/token/src/instructions/thaw_account.rs
febo: This error is also present in SPL Token and pinocchio token interface. 😅

https://github.com/solana-program/token/pull/144

programs/token/src/instructions/close_account.rs
febo: We could and indeed simplifies the code, but it is a bit worse in terms of CU.

febo:
@grod220 Ready for a second look.

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn new() -> Self {
+        let mut data: Box<[MaybeUninit<u8>]> = Box::new_uninit_slice(MAX_CPI_INSTRUCTION_DATA_LEN);
+        // The first byte of the instruction data is reserved for the batch instruction
+        // discriminator.
+        data[0].write(Self::DISCRIMINATOR);
+
+        Self {
+            data,
+            instruction_accounts: Box::new_uninit_slice(MAX_CPI_ACCOUNTS),
+            accounts: Box::new_uninit_slice(MAX_CPI_ACCOUNTS),
```
joncinque: Doesn't this use up a huge amount of heap space? Just the instruction data would use up one third of the total allowed by default.

If we're using the heap at all, how about allowing people to specify the sizes to allocate in `new`? This could be a wrapper over a more generic `Batch` struct which just holds slices, and that way people could even stack-allocate everything if they prefer.

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,156 @@
...
+        let written_data = write_data(&mut self.data[self.data_len + IX_HEADER_SIZE..])?;
+
+        let written_accounts = write_accounts(&mut self.accounts[self.accounts_len..])?;
+
+        let written_instruction_accounts = write_instruction_accounts(
+            &mut self.instruction_accounts[self.instruction_accounts_len..],
+        )?;
+
+        // If all writes succeeded, update the lengths and write the instruction
+        // header.
```
joncinque, line 135: This is pretty neat! We don't often allow for recoverable situations, so it's nice when it's possible

programs/token/src/instructions/batch.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        Ok(())
+    }
+}
+
+impl Default for Batch<'_> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
joncinque: I'd probably remove this in favor of forcing sizes at creation time

programs/token/src/instructions/burn.rs
```diff
@@ -228,17 +227,21 @@ where
         return Err(invalid_argument_error());
     }
 
-    accounts[0].write(writable_cpi_account(account)?);
+    if account.is_borrowed() | mint.is_borrowed() {
```
joncinque, line 232: nit: since `is_borrowed()` returns a bool, the single `|` is a little misleading, a real `||` would make more sense, unless there's a performance reason to use `|`

joncinque (commented):
Just taking a look at the recent changes

programs/token/src/instructions/burn.rs
febo: Yeah, I should have put a comment on the PR. Surprisingly, a single | is more efficient.

programs/token/src/instructions/batch.rs
febo: This was a really good idea. I made the constants for `data_len` and `accounts_len` public on each instructions, so it is possible to statically calculate how much to allocate.

programs/token/src/instructions/set_authority.rs
```diff
@@ -115,14 +115,14 @@ impl<'account, 'address, 'multisig, MultisigSigner: AsRef<AccountView>>
             Err(ProgramError::InvalidArgument)?;
         }
 
-        let mut instruction_accounts = [UNINIT_INSTRUCTION_ACCOUNT; MAX_ACCOUNTS_LEN];
+        let mut instruction_accounts = [UNINIT_INSTRUCTION_ACCOUNT; SetAuthority::MAX_ACCOUNTS_LEN];
```
joncinque, line 118: nit: up to you, but it would be simpler to refer to these consts through `Self` rather than the full type name

joncinque (dismissed):
Beautiful! Just a nit that you can take or leave from my side

programs/token/src/instructions/set_authority.rs
febo: I did where it was possible. For instructions that have a multisig, it was not possible since the constant is defined on the `&AccountView` impl.

PR #380, opened 2026-03-24
token: Add `UnwrapLamports` instruction

febo:
### Problem

[SIMD-0266](https://github.com/febo/solana-improvement-documents/blob/main/proposals/0266-efficient-token-program.md) was activated on testnet, so soon the token program will support [`UnwrapLamports`](https://github.com/solana-program/token/blob/main/pinocchio/interface/src/instruction.rs#L508-L530) instruction. 

### Solution

Add support for `UnwrapLamports` instruction.

Note: This PR requires #377 to land first.

### DIFF
--- a/programs/token/src/instructions/unwrap_lamports.rs
+++ b/programs/token/src/instructions/unwrap_lamports.rs
@@ -14,6 +14,21 @@ use {
     solana_program_error::{ProgramError, ProgramResult},
 };
 
+/// Enum specifying the amount of lamports to transfer
+/// from a native SOL account.
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Amount {
+    /// Transfer the entire amount of the source account.
+    All,
+
+    /// Transfer a specified amount.
+    ///
+    /// The value must be less than or equal to the amount
+    /// of the source account.
+    Some(u64),
+}
+
 /// Transfer lamports from a native SOL account to a destination account.
 ///
 /// This is useful to unwrap lamports from a wrapped SOL account.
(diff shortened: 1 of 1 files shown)


programs/token/src/instructions/unwrap_lamports.rs
```diff
@@ -0,0 +1,324 @@
...
+    pub authority: &'account AccountView,
+
+    /// Multisignature signers.
+    pub multisig_signers: &'multisig [MultisigSigner],
+
+    /// The amount of lamports to transfer.
+    ///
+    /// When an amount is not specified, the entire balance of the source
+    /// account will be transferred.
+    pub amount: Option<u64>,
```
grod220: `None` meaning "all lamports" feels a bit mismatched no? What if we do this:

```rust
enum Amount {
  All,
  Partial(u64)
}

...

pub amount: Amount

```

Between these two, the second feels more declarative of what's happening:
```
UnwrapLamports::new(source, dest, auth, None)
  vs
UnwrapLamports::new(source, dest, auth, Amount::All)
```

grod220: Now seeing token-2022 shipped with the same `Option<u64>` field though so this would be inconsistent if changed

febo: Yeah, I like this approach since it communicates very clearly what the parameter is doing. I don't think we need to change the program implementation, although it would be nice to have it the same, but it can be something done by the client only. What do you think @joncinque ?

joncinque: Fine by me! I would go with the names `All` and `Some(u64)` for the variants, to make it look more like `SpendAmount` in the CLI, and also like the base `Option` type

febo: I think it looks better now.

PR #382, opened 2026-04-01
ci: pinning github actions + adding depandabot to bump the deps that we have

nbelenkov:
All actions were using floating tags (@v4, @master, etc.), which can be silently updated by the upstream repo.

**Changes:**

- Pinned every action in `main.yml`, `publish.yml`, and `.github/actions/setup/action.yml` to a specific commit SHA with the version tag as a comment
- Swapped `dtolnay/rust-toolchain@master` for `actions-rust-lang/setup-rust-toolchain@v1.9.0`. It is the same interface, but has proper versioned releases, so it can actually be pinned meaningfully. Added flags so it behaves the same as the old action.
- Added `dependabot.yml` with weekly grouped updates for Actions, Cargo, and npm. Actions updates have a 7-day cooldown, as we don't need the latest straight away and can let it bake for a bit.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,16 +13,6 @@ updates:
         patterns:
           - "*"
 
-  - package-ecosystem: cargo
-    directory: /
-    schedule:
-      interval: weekly
-      day: monday
-    groups:
-      cargo:
-        patterns:
-          - "*"
-
   - package-ecosystem: npm
     directory: /
     schedule:


.github/dependabot.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    groups:
+      cargo:
+        patterns:
+          - "*"
```
joncinque: Since pinocchio is a library, bumping dependencies should only be done if needed, ie if there's a security vulnerability or if we need to use a new / breaking API.

Otherwise, new versions create needless churn for downstream users. We should typically declare the lowest required deps.

nbelenkov: hmm, that is a very good point. I was mainly thinking of Dependabot in the context of actions, which should be bumped, but yeah, for Pinocchio itself, we should limit it to just security vulnerabilities. Will update with that, wdyt with npm side, should we also do just vulns?

joncinque: The npm packages are only used for internal scripts in the repo, so it should be fine to update those deps. In my opinion, those scripts should be removed, but that's a conversation for another day :smile:

nbelenkov: cool, sounds good, will purge the cargo dependabot and we enable db for security advisories (if it is not already)

febo: > The npm packages are only used for internal scripts in the repo, so it should be fine to update those deps. In my opinion, those scripts should be removed, but that's a conversation for another day 😄

100% agree.

PR #402, opened 2026-04-10
sdk: Enable default features

febo:
### Problem

Currently to use additional features from re-exported crates (e.g., solana-address), it is necessary to separately import the crate. There are features that are used by most programs, which could be beneficial to enable them by default.

### Solution

This PR adds a `"sha2"` feature to enable `"solana-address/sha2"`, and adds both `"copy"` and `"sha2"` to the default features of pinocchio. Program can opt0out by disabling default features.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -22,11 +22,12 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 ] }
 
 [features]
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Makes sense -- we'll just need to be careful about adding too many features to pinocchio.

PR #403, opened 2026-04-13
programs/system: Use invoke unchecked

febo:
### Problem

Currently the CPI helpers include more validation than needed. For example, when preparing for a CPI, helpers already guarantee that accounts are set in the correct order. Yet, they use `invoke_signed`, which will validate that accounts are in the expected order by performing an address comparison – this is unnecessary.

### Solution

Update the CPI logic to use `invoke_signed_unchecked`, which does not include any of the unnecessary validation. There is an improvement of about `~200` CUs over the current version, since they are also using the improved `CpiAccount` initialization from https://github.com/anza-xyz/solana-sdk/pull/650.

### DIFF
--- a/programs/system/src/instructions/advance_nonce_account.rs
+++ b/programs/system/src/instructions/advance_nonce_account.rs
@@ -1,7 +1,11 @@
-use pinocchio::{
-    cpi::{invoke_signed, Signer},
-    instruction::{InstructionAccount, InstructionView},
-    AccountView, ProgramResult,
+use {
+    core::{mem::MaybeUninit, slice::from_raw_parts},
+    pinocchio::{
+        cpi::{invoke_signed_unchecked, CpiAccount, Signer},
+        error::ProgramError,
+        instruction::{InstructionAccount, InstructionView},
+        AccountView, ProgramResult,
+    },
 };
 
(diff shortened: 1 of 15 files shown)


LStan:
Why
```
let mut instruction_accounts = [const { MaybeUninit::<InstructionAccount>::uninit() }; 1];
instruction_accounts[0].write(InstructionAccount::writable_signer(self.account.address()));
```
is faster than
```
let instruction_accounts: [InstructionAccount; 1] = [InstructionAccount::writable_signer(self.account.address())];
```
?
Is `instruction_accounts` initialized with zeros first in the second case?

febo:
> Why
> 
> ```
> let mut instruction_accounts = [const { MaybeUninit::<InstructionAccount>::uninit() }; 1];
> instruction_accounts[0].write(InstructionAccount::writable_signer(self.account.address()));
> ```
> 
> is faster than
> 
> ```
> let instruction_accounts: [InstructionAccount; 1] = [InstructionAccount::writable_signer(self.account.address())];
> ```
> 
> ? Is `instruction_accounts` initialized with zeros first in the second case?

Compiler quirks – code generation ends up more efficient in the first case.

joncinque (approved):
Looks great!

PR #404, opened 2026-04-13
programs/token-2022: Fix `Mint` owner check

febo:
### Problem

PR #400 wrongly removed the owner check instead of the length check in `Mint::from_account_view`.

### Solution

Add the owner check back and remove the length check, which is done by `validate_account_type`.


joncinque (approved):
Looks great!

PR #415, opened 2026-05-28
token-2022: add account-size helpers

grod220:
Adds `extension::account_len` module in `pinocchio-token-2022` for sizing a Token-2022 account. .

 ### Why

Programs like p-ATA want the ability to size token-2022 accounts without the need for paying the CU cost of a `GetAccountDataSize` CPI. Reference: https://github.com/solana-program/associated-token-account/pull/258.

### New method

`try_calculate_account_len_from_mint()` is added to be a utility for optimistic, fast path sizing. If mint data errs, the caller is expected to fallback to `GetAccountSize` CPI.

### DIFF
--- a/programs/token-2022/src/state/extension/account_len.rs
+++ b/programs/token-2022/src/state/extension/account_len.rs
@@ -136,20 +136,21 @@ impl SeenExtensions {
     }
 }
 
-/// Computes the byte length of a Token-2022 account for `mint`: the extensions
-/// every account of that mint requires plus any `additional_account_extensions`
-/// (e.g. the `ATA` program adds `ImmutableOwner`).
+/// Computes the byte length a Token-2022 account for `mint` needs, counting the
+/// extensions every account of that mint requires plus any
+/// `additional_account_extensions` (e.g. the `ATA` program adds
+/// `ImmutableOwner`).
 ///
-/// - `Ok(Some(len))`: Sized locally.
-/// - `Ok(None)`: Can't size locally (ext not recognized or don't track a size
-///   for). Defer to `GetAccountDataSize` CPI.
-/// - `Err(..)`: Invalid input (wrong owner, mint-only ext passed as account
-///   ext, invalid mint data bytes, etc).
+/// Returns `Err` whenever the size can't be computed locally. This can be due
+/// to invalid input (wrong owner, a mint-only ext passed as an account ext,
+/// non-mint data) or inputs it can't size (an extension it doesn't recognize
+/// or track a size for). Callers should defer to the token program's
+/// `GetAccountDataSize` if it's not recognized here.
 #[inline]
 pub fn try_calculate_account_len_from_mint(
     mint: &AccountView,
     additional_account_extensions: &[ExtensionType],
-) -> Result<Option<usize>, ProgramError> {
+) -> Result<usize, ProgramError> {
     if !mint.owned_by(&ID) {
         return Err(ProgramError::IncorrectProgramId);
     }
@@ -168,9 +169,7 @@ pub fn try_calculate_account_len_from_mint(
         if seen.insert(ext) {
             has_extensions = true;
             validate_extension_account_type(ext, AccountType::Account)?;
-            let Some(value_len) = extension_value_len(ext) else {
-                return Ok(None); // unrecognized ext
-            };
+            let value_len = extension_value_len(ext).ok_or(ProgramError::InvalidInstructionData)?;
             total_len = total_len
                 .checked_add(TLV_HEADER_LEN + value_len)
                 .ok_or(ProgramError::InvalidInstructionData)?;
@@ -185,32 +184,28 @@ pub fn try_calculate_account_len_from_mint(
 
         // Find what mint extensions are present
         let mut mint_exts = [ExtensionType::Uninitialized; MAX_EXTENSIONS];
-        let Ok(mint_ext_count) = state.write_extension_types(&mut mint_exts) else {
-            return Ok(None);
-        };
+        let mint_ext_count = state.write_extension_types(&mut mint_exts)?;
         for &mint_ext in &mint_exts[..mint_ext_count] {
-            let Some(required) = required_account_extensions_from_mint_extension(mint_ext) else {
-                return Ok(None); // unrecognized ext
-            };
+            let required = required_account_extensions_from_mint_extension(mint_ext)
+                .ok_or(ProgramError::InvalidAccountData)?;
             for &ext in required {
                 if seen.insert(ext) {
                     has_extensions = true;
-                    let Some(value_len) = extension_value_len(ext) else {
-                        return Ok(None); // unrecognized ext
-                    };
+                    let value_len =
+                        extension_value_len(ext).ok_or(ProgramError::InvalidAccountData)?;
                     total_len = total_len
                         .checked_add(TLV_HEADER_LEN + value_len)
-                        .ok_or(ProgramError::InvalidInstructionData)?;
+                        .ok_or(ProgramError::InvalidAccountData)?;
                 }
             }
         }
     }
 
     if !has_extensions {
-        return Ok(Some(Account::BASE_LEN));
+        return Ok(Account::BASE_LEN);
     }
 
-    Ok(Some(adjust_len_for_multisig(total_len)))
+    Ok(adjust_len_for_multisig(total_len))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


programs/token-2022/src/state/extension/account_len.rs
```diff
@@ -0,0 +1,389 @@
...
+    solana_program_error::ProgramError,
+};
+
+/// Returns the fixed byte length of the given extension's value payload,
+/// or `None` if the extension type is not yet supported.
+///
+/// Only a subset of extension types have known sizes registered here.
+/// Unsupported types will cause [`try_calculate_account_len`] to fail.
+#[inline(always)]
+pub const fn extension_value_len(extension_type: ExtensionType) -> Option<usize> {
```
grod220, line 25: moved from mod.rs

programs/token-2022/src/state/extension/account_len.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// Inserts `extension_type` into the seen struct. Returns `true` if it
+    /// wasn't already present.
+    #[inline(always)]
+    fn insert(&mut self, extension_type: ExtensionType) -> bool {
+        let bit = 1u32 << (extension_type as u32);
+        let is_new = self.0 & bit == 0;
+        self.0 |= bit;
+        is_new
+    }
+}
```
grod220, line 137: Exists to create a performant version of [try_get_total_tlv_len()](https://github.com/solana-program/token-2022/blob/876149c50c957f0a73ffe0349b20c44c4b5cb5f3/interface/src/extension/mod.rs#L1232-L1241) (doesn't require allocation + multiple iteration).

programs/token-2022/src/state/extension/account_len.rs
```diff
@@ -0,0 +1,383 @@
...
+        _ => None,
+    }
+}
+
+/// Returns the account data length needed for the given extension types.
+///
+/// Only extension types with sizes registered in [`extension_value_len`] are
+/// supported.
+#[inline]
+pub fn try_calculate_account_len<B: ExtensionBaseState>(
```
grod220, line 45: moved from mod.rs

programs/token-2022/src/state/extension/account_len.rs
```diff
@@ -0,0 +1,383 @@
...
+
+/// Computes the byte length a Token-2022 account for `mint` needs, counting the
+/// extensions every account of that mint requires plus any
+/// `additional_account_extensions` (e.g. the `ATA` program adds
+/// `ImmutableOwner`).
+///
+/// Returns `Err` whenever the size can't be computed locally. This can be due
+/// to invalid input (wrong owner, a mint-only ext passed as an account ext,
+/// non-mint data) or inputs it can't size (an extension it doesn't recognize
+/// or track a size for). Callers should defer to the token program's
```
grod220, line 147: Went back and forth whether this should should return a: `Result<usize>` or `Result<Option<usize>>`. I didn't see any real precedent for the latter so thought it simpler to not include the option. 

Consumers (like the ATA program) can't distinguish if the failure here is due to malformed bytes or that the extension isn't recognized in these helpers. In both cases, the consumer will have to do a fallback to `GetAccountSize` CPI. If we used another type, they could distinguish the cases and propagate the err instead of CPI. It may just be simpler to keep the return type simpler though 🤔

programs/token-2022/src/state/extension/shared_test_helpers.rs
```diff
@@ -0,0 +1,48 @@
...
+use {
+    super::{ExtensionType, ACCOUNT_TYPE_INDEX, TLV_START_INDEX},
+    crate::state::AccountType,
+    alloc::{vec, vec::Vec},
+    core::{mem::size_of, ptr::copy_nonoverlapping},
+    solana_account_view::{AccountView, RuntimeAccount, NOT_BORROWED},
+    solana_address::Address,
+};
+
+pub(super) fn push_tlv_entry(buffer: &mut Vec<u8>, extension_type: ExtensionType, value: &[u8]) {
```
grod220, line 12: All test helpers were not changed, just relocated so they can be shared

programs/token-2022/src/state/extension/account_len.rs
```diff
@@ -0,0 +1,383 @@
...
+        let is_new = self.0 & bit == 0;
+        self.0 |= bit;
+        is_new
+    }
+}
+
+/// Computes the byte length a Token-2022 account for `mint` needs, counting the
+/// extensions every account of that mint requires plus any
+/// `additional_account_extensions` (e.g. the `ATA` program adds
+/// `ImmutableOwner`).
```
grod220, line 142: This fn takes inspiration from [process_get_account_data_size()](https://github.com/solana-program/token-2022/blob/876149c50c957f0a73ffe0349b20c44c4b5cb5f3/program/src/processor.rs#L1505-L1530) ([get_required_account_extensions(](https://github.com/solana-program/token-2022/blob/876149c50c957f0a73ffe0349b20c44c4b5cb5f3/program/src/processor.rs#L2294-L2312)) & [try_calculate_account_len()](https://github.com/solana-program/token-2022/blob/876149c50c957f0a73ffe0349b20c44c4b5cb5f3/interface/src/extension/mod.rs#L1246-L1256)) and uses a similar sizing logic.

Decided to optimize and put the logic in the pinocchio crate because:

- token-2022-interface pulled in a ton of deps
- the current helpers are not efficient (allocates three vecs per call) and had worse CU benches
- the pinocchio repo works with AccountView

That said, there is an argument that these helpers can be relocated to token-2022-interface and replace the current size helpers there and accept mint_data versus AccountView/AccountInfo. But given that token-2022 will soon be pinocchio-ized, not sure if that is worth it as the work of this PR will eventually be used regardless.

joncinque (commented):
Before I read through this, is there a reason to use this over the logic that already exists in spl-token-2022-interface?

programs/token-2022/src/state/extension/account_len.rs
grod220: To give more details to ⬆️ 

We can't use the token-2022 work as is because: (a) it uses allocations and (b) some logic we need exposed is internal to the program (e.g. `get_required_account_extensions()`).

If we are willing to make new helper take raw data as a param (versus AccountInfo/View), we could put the dedicated helper in `spl-token-2022-interface`, and make the token-2022 processor call that helper for `GetAccountDataSize` instead of its internal logic.

But the question is sort of a larger one: where should new pinocchio token logic live? We seem to have a split story at the moment. It does seem that the direction is going with the pinocchio repo getting the optimized logic at first and as the program gets migrated, that logic moves to the new p-token-interface crates? If that's the case, that's an argument for keeping it here until t22's `GetAccountDataSize` gets migrated.

grod220:
@joncinque relevant thread here: https://github.com/anza-xyz/pinocchio/pull/415#discussion_r3357628430. Worth discussing for sure.

programs/token-2022/src/state/extension/account_len.rs
joncinque: Unless there's a compelling reason, rewriting existing audited code is a waste of time at best, and a security issue at worst.

Let's start from the beginning. We know that we want to do a Pinocchio migration for token-2022, so we have two approaches:

* rewrite everything from scratch. We did this for p-token and p-ata, and it's worked well. On the other hand, those are considerably smaller programs than token-2022  
* adapt the existing code piece by piece, Ship of Theseus-style. This could prevent optimizations, and require breaking changes to the interface crate

What are the current issues in spl-token-2022-interface?

* Uses solana-instruction for instruction creators, which requires allocations
* Uses solana-account-info for helper functions, which has issues 
* Uses bytemuck for `Pod` types, which adds a few unnecessary checks
* some helpers return allocated types
* Confidential-transfer / ZK deps also use these, so they need to get updated too

How can we adapt all of these for the pinocchio future?

* remove allocations. This one is pretty easy, since we can mostly work with slices and arrays. We can even keep functions that return allocations, and not call them in the program, maybe even gate them on an `alloc` feature in the crate.
* stop using bytemuck. We did a breaking release to stop using spl-pod, so I imagine we could do that in a normal breaking release for all crates
* stop using solana-account-info. Same here, we could transition to using solana-account-view in a breaking change since most of those functions are only used by the program
* stop using solana-instruction. This one gets tricky since there's downstream usage of the instruction creators. We could get around this by gating usage on an `alloc` feature on the crate.

I mostly wrote all of this out to see which way I would lean, and given everything that needs to change, I lean towards doing the adaptations that I've outlined, by adding an `alloc` feature on the crate, which would not be used by the pinocchio program.

Also tagging @febo to see what he thinks. Since the two of you will likely end up doing most of the implementation, I'll defer to what you both think, but we should make sure we're aligned on the path forward before we copy-pasta too much more code

febo: I think we should definitely aim for adapting the existing code, given that the program is upgradable. The tricky thing is that we have a chicken-and-egg situation. 😅

For example, if we want to start using `AccountView` and change the interface, we need to update the program; or if we change the program, then we need to update the interface. Having the changes in the pinocchio crate helps here, since the changes are isolated from the program. But agree that copying code around is not ideal.

One way that we could approach this is to pin the interface version for the program and start adapting the interface in-place. We can make the changes to the program independently as the interface is updated. And we can start moving the code from pinocchio to Token-2022 repository.

Another consideration is whether we will "drop" support for `AccountInfo` altogether or not. Since Token-2022 is upgradable, when we add new extensions, are programs forced to migrate to an interface that only uses `AccountView`? Otherwise it would make sense to have a separate crate, although we should still try to avoid duplicating logic that is independent of these types.

joncinque: > One way that we could approach this is to pin the interface version for the program and start adapting the interface in-place. We can make the changes to the program independently as the interface is updated. And we can start moving the code from pinocchio to Token-2022 repository.

That's a fine option too!

> Another consideration is whether we will "drop" support for AccountInfo altogether or not.

There's literally three functions in the interface that use `AccountInfo`, and they shouldn't be used outside of the program, so a normal breaking change will be good.

grod220: Helpful thread. Definitely worthwhile to align on the migration strategy before we find ourselves managing redundant logic. I see that ejecting from the other crate was premature. To unblock p-ATA, think it's best we close this PR and adapt token-2022-interface (pulling in optimizations from here). Will work on that!

PR #416, opened 2026-05-28
Add ATA-relevant ext structs

grod220:
Pulling out just the specific extension structs added to https://github.com/anza-xyz/pinocchio/pull/415. This multi-part PR series is to help p-ATA calculate token 2022 account lengths https://github.com/solana-program/associated-token-account/pull/258.

This PR just focuses on adding the ATA-relevant extension structs w/ length that can later be referenced.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /node_modules
 /target
+.idea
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 /node_modules
 /target
+.idea
```
febo, line 3: Just checking that this was intentional.

febo (approved):
Looks great! Just checking if the change to `.gitignore` was intentional.

.gitignore
grod220: Indeed, for us intellij ide users, we kinda need it: https://github.com/solana-program/stake/blob/main/.gitignore#L10

PR #418, opened 2026-05-28
fix(token-2022): use extension types for account sizing

dev-jodee:
## Summary
- Switch GetAccountDataSize to accept Token-2022 ExtensionType values instead of instruction ExtensionDiscriminator values.
- Apply the same ExtensionType API to Reallocate, which serializes the same extension list format.
- Share little-endian u16 extension-list encoding and cover it with focused unit tests.

## Testing
- cargo test -p pinocchio-token-2022
- pnpm cargo-audit
- pnpm miri
- pnpm format/clippy/doc/hack/build-sbf/test for all Main workflow matrix members


febo (approved):
Looks great!

PR #421, opened 2026-06-09
sdk: Use direct account pointers in program entrypoint

febo:
### Problem

[SIMD-0449](https://github.com/solana-foundation/solana-improvement-documents/pull/449) proposes to serialize pointers for instruction accounts to allow direct account access without requiring parsing of the account section. This is currently not being used in the entrypoint implementation.

### Solution

Use the direct account pointers in program entrypoint, removing the account parsing logic.

Considering a simple entrypoint:
```rust
entrypoint!(process_instruction);

pub fn process_instruction(
    program_id: &Address,
    accounts: &mut [AccountView],
    instruction_data: &[u8],
) -> ProgramResult {
    core::hint::black_box((program_id, accounts, instruction_data));
    Ok(())
}
```

Below are the updated values for entrypoint CU consumption, where the "Delta" column represents the difference in relation to the current implementation.
| Name | CUs | Delta |
|------|------|-------|
| Account (0) | 14 | -5 |
| Account (1) | 14 | -10 |
| Account (2) | 14 | -19 |
| Account (3) | 14 | -33 |
| Account (4) | 14 | -41 |
| Account (8) | 14 | -75 |
| Account (16) | 14 | -135 |
| Account (32) | 14 | -258 |
| Account (64) | 14 | -501 |

When the feature `"account-resize"` is enabled, there is still a need to iterate over all accounts to store the original `data_len` value but there are still gains in CUs over the current implementation:
| Name | CUs | Delta |
|------|------|-------|
| Account (0) | 22 | +2 |
| Account (1) | 24 | -1 |
| Account (2) | 30 | -5 |
| Account (3) | 36 | -15 |
| Account (4) | 39 | -22 |
| Account (8) | 54 | -49 |
| Account (16) | 84 | -96 |
| Account (32) | 144 | -190 |
| Account (64) | 264 | -377 |

Note that a similar modification will be done to the `lazy_program_entrypoint` in a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ The [`entrypoint!`](https://docs.rs/pinocchio/latest/pinocchio/macro.entrypoint.
 * [`default_allocator!`](https://docs.rs/pinocchio/latest/pinocchio/macro.default_allocator.html): declares the default (bump) global allocator
 * [`default_panic_handler!`](https://docs.rs/pinocchio/latest/pinocchio/macro.default_panic_handler.html): declares the default panic "hook" that works in combination with the `std` panic handler
 
-When all dependencies are `no_std`, you should use [`nostd_panic_handler!`](https://docs.rs/pinocchio/latest/pinocchio/macro.nostd_panic_handler.html) instead of `default_panic_handler!` to declare a rust runtime panic handler. There's no need to do this when any dependency is `std` since rust compiler will emit a panic handler.
+When all dependencies are `no_std`, you should use [`nostd_panic_handler!`](https://docs.rs/pinocchio/latest/pinocchio/macro.nostd_panic_handler.html) instead of `default_panic_handler!` to define a Rust runtime panic handler. There's no need to do this when any dependency is `std` since Rust compiler will emit a panic handler.
 
 To use the `entrypoint!` macro, use the following in your entrypoint definition:
 ```rust
@@ -79,10 +79,10 @@ pub fn process_instruction(
 }
 ```
 
-The information from the input is parsed into their own entities:
+The information from the input is split into their own entities:
 
-* `program_id`: the `ID` of the program being called
-* `accounts`: the accounts received
+* `program_id`: address of the program being called
+* `accounts`: accounts received by the instruction
 * `instruction_data`: data for the instruction
 
 `pinocchio` also offers variations of the program entrypoint (`lazy_program_entrypoint`) and global allocator (`no_allocator`). In order to use these, the program needs to specify the program entrypoint, global allocator and panic handler individually. The `entrypoint!` macro is equivalent to writing:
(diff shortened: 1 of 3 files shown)


sdk/src/entrypoint/mod.rs
```diff
@@ -291,258 +239,108 @@ macro_rules! align_pointer {
...
-    // Number of accounts to process.
-    let mut processed = *(input as *const u64) as usize;
-    // Skip the number of accounts (8 bytes).
-    input = input.add(size_of::<u64>());
-
-    if processed > 0 {
-        let mut accounts = accounts.as_mut_ptr() as *mut AccountView;
-        // Represents the beginning of the accounts slice.
-        let accounts_slice = accounts;
+pub unsafe fn process_entrypoint<F>(
```
febo: This is a breaking change since the parameter list changed. We could deprecate the old one and add this one with a different name, but that would result in keeping all the parsing logic.

sdk/src/entrypoint/mod.rs
```diff
@@ -291,258 +239,108 @@ macro_rules! align_pointer {
...
+    // The slice of account pointers is located right after the `program_id` +
+    // alignment padding. The length of the slice is determined by the first
+    // 8-bytes of `program_input`.
+
+    let slice_ptr = instruction_data.add(ix_data_len + size_of::<Address>());
+
+    let accounts_len = *(program_input as *const u64) as usize;
+    let accounts_ptr: *mut AccountView = align_pointer!(slice_ptr);
+
+    #[cfg(feature = "account-resize")]
```
febo, line 294: This is a bit annoying – we need to iterate over all accounts to store the original data length.

febo:
@grod220 Ready for review.

README.md
joncinque, line 1: I don't know how you want to surface the breaking change, but I think some line in the README about what changes people need to make would be helpful.

sdk/src/entrypoint/mod.rs
joncinque: Hm, it's up to you, but it certainly doesn't hurt to have a clear deprecation and new functionality in one release.

We can immediately remove it and publish a breaking change, but at least people get a message with some information about what to do.

The alternative is to do a separate release with *just* the deprecation, but that would probably be more annoying, since there wouldn't be another entrypoint to use 😅

joncinque (commented):
The code looks great to me! I'll let you choose how to proceed with the release, but I would go with a purely additive PR + deprecation of old entrypoint, release that, then do the breaking change

sdk/src/entrypoint/mod.rs
febo: Good point! I added back the code with a deprecated note. I used the name `process_program_input` instead for the new way to handle the program input, so we can deprecate the old one and eventually remove it.

README.md
febo: Last time I wrote the details on the release notes.

Update: For programs using the entrypoint macros, there are no changes; programs using `process_entrypoint` and `deserialize` should use `process_program_input` instead. Majority of programs probably fall into the first category.

README.md
```diff
@@ -114,16 +113,19 @@ no_allocator!();
...
+) -> u64 {
     // Fast path: check the number of accounts
-    let num_accounts = unsafe { *(input as *const u64) };
+    let num_accounts = unsafe { *(program_input as *const u64) };
     if num_accounts == 0 {
         log("Fast path - no accounts!");
         return 0;
     }
 
     // Standard path: delegate to `process_entrypoint`
```
grod220:
```diff
- // Standard path: delegate to `process_entrypoint`
+ // Standard path: delegate to `process_program_input`
```

README.md
```diff
@@ -95,16 +95,15 @@ Any of these macros can be replaced by alternative implementations.
 
 📌 Custom entrypoints with [`process_entrypoint`](https://docs.rs/pinocchio/latest/pinocchio/entrypoint/fn.process_entrypoint.html)
```
grod220: This needs updating to `process_program_input` right?

sdk/src/entrypoint/mod.rs
```diff
@@ -204,44 +202,155 @@ macro_rules! entrypoint {
...
+    ( $ptr:ident ) => {
+        // Integer-to-pointer cast: first compute the aligned address as a `usize`,
+        // since this is more CU-efficient than using `ptr::align_offset()` or the
+        // strict provenance API (e.g., `ptr::with_addr()`). Then cast the result
+        // back to a pointer. The resulting pointer is guaranteed to be valid
+        // because it follows the layout serialized by the runtime.
+        with_exposed_provenance_mut(
+            ($ptr.expose_provenance() + (BPF_ALIGN_OF_U128 - 1)) & !(BPF_ALIGN_OF_U128 - 1),
+        )
+    };
```
grod220: This appears to be a duplicate of an `align_pointer!` already in this file

grod220 (dismissed):
Small nits from me, but looks good!

joncinque (dismissed):
Looks good to me!

sdk/src/entrypoint/mod.rs
febo: Oh, weird, wondering why the compiler did not complain about it.

febo:
@joncinque @grod220 Fixed all the nits.

grod220 (dismissed):
💪

joncinque (dismissed):

joncinque (approved):

PR #424, opened 2026-06-09
ci: adding gh app for release flow

nbelenkov:
Swaps the publish workflow from the `ANZA_TEAM_PAT` to our GitHub App token (our normal setup), so publishing aligns with the rest of our release flows.

## What we need to do still
- Create the release github app
- Needs `vars.CLIENT_ID`, `vars.APP_ID`, and `secrets.PRIVATE_KEY` set in the `prod` environment.
- The App has to be in the branch-protection bypass list, since publish pushes the version commit + tag to a protected branch

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -130,17 +130,13 @@ jobs:
         run: pnpm test ${{ inputs.crate }}
 
       - name: Set Git Author
-        run: |
-          if [ "${{ inputs.dry_run }}" == "true" ]; then
-            git config --global user.email "github-actions@github.com"
-            git config --global user.name "github-actions"
-          else
-            git config --global user.email "${APP_ID}+${APP_SLUG}[bot]@users.noreply.github.com"
-            git config --global user.name "${APP_SLUG}[bot]"
-          fi
+        if: github.event.inputs.dry_run != 'true'
         env:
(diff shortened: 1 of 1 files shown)


nbelenkov:
cc @yihau on the github app side

.github/workflows/publish.yml
```diff
@@ -123,8 +131,16 @@ jobs:
...
-          git config --global user.email "github-actions@github.com"
-          git config --global user.name "github-actions"
+          if [ "${{ inputs.dry_run }}" == "true" ]; then
+            git config --global user.email "github-actions@github.com"
+            git config --global user.name "github-actions"
+          else
+            git config --global user.email "${APP_ID}+${APP_SLUG}[bot]@users.noreply.github.com"
+            git config --global user.name "${APP_SLUG}[bot]"
+          fi
+        env:
```
yihau, line 134: nig: we can just use `github-actions[bot]` to keep things simple.

nbelenkov: Simplified it here [6aa5ab8](https://github.com/anza-xyz/pinocchio/pull/424/commits/6aa5ab8058cdde447a673bf75619197bf59a6be0), would be nice to have the actual identity as we have it in all the other repos

yihau (approved):

PR #428, opened 2026-07-02
programs/token: Generic instruction builders

febo:
### Problem

Currently there are many instruction builders duplicated between token and token-2022. This is not ideal since it duplicates the maintenance effort.

### Solution

Introduce a generic parameter so that instruction builders can be used with different token programs without the need to duplicate the code.


programs/token/src/instructions/amount_to_ui_amount.rs
```diff
@@ -1,17 +1,32 @@
...
+    solana_address::Address,
     solana_instruction_view::{
         cpi::{invoke_unchecked, CpiAccount},
         InstructionAccount, InstructionView,
     },
     solana_program_error::{ProgramError, ProgramResult},
 };
 
+/// The instruction discriminator.
+const DISCRIMINATOR: u8 = 23;
```
febo, line 20: Because of the generic parameter, the constants need to be out of the type so they can be used in the implementation.

programs/token/src/instructions/amount_to_ui_amount.rs
```diff
@@ -23,47 +38,59 @@ use {
...
     }
 
     #[inline(always)]
     pub fn invoke(&self) -> ProgramResult {
-        let mut instruction_accounts = [UNINIT_INSTRUCTION_ACCOUNT; AmountToUiAmount::ACCOUNTS_LEN];
+        self.invoke_with_program(&Program::ID)
+    }
+
+    #[inline(always)]
+    pub fn invoke_with_program(&self, program: &Address) -> ProgramResult {
```
febo, line 81: Every instruction builder has a `*_with_program` variant that can be used with any program address.

programs/token/src/lib.rs
```diff
@@ -2,32 +2,25 @@
...
-};
-
 solana_address::declare_id!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
 
-const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
-
-const UNINIT_CPI_ACCOUNT: MaybeUninit<CpiAccount> = MaybeUninit::<CpiAccount>::uninit();
+/// A trait for token programs that can be used in a CPI with a staticly known
+/// program address.
+pub trait TokenProgram {
```
febo, line 17: Trait to provide the program address – this is used to maintain the current API.

programs/token/src/instructions/mod.rs
```diff
@@ -1,47 +1,513 @@
...
+///
+/// Fails on an invalid mint.
+///
+/// Return data can be fetched using `sol_get_return_data` and deserialized
+/// with `String::from_utf8`.
+///
+/// Accounts expected by this instruction:
+///
+///   0. `[]` The mint to calculate for.
+pub type AmountToUiAmount<'account> = amount_to_ui_amount::AmountToUiAmount<'account, Program>;
```
febo, line 62: Other token programs (e.g., token-2022) just need to export a type alias with their own `TokenProgram` impl.

febo:
@joncinque @grod220 This is an idea to reduce the duplication between token and token-2022 helpers.

programs/token/src/instructions/mod.rs
```diff
@@ -1,47 +1,513 @@
...
-    amount_to_ui_amount::*, approve::*, approve_checked::*, batch::*, burn::*, burn_checked::*,
-    close_account::*, freeze_account::*, get_account_data_size::*, initialize_account::*,
-    initialize_account_2::*, initialize_account_3::*, initialize_immutable_owner::*,
-    initialize_mint::*, initialize_mint_2::*, initialize_multisig::*, initialize_multisig_2::*,
-    mint_to::*, mint_to_checked::*, revoke::*, set_authority::*, sync_native::*, thaw_account::*,
-    transfer::*, transfer_checked::*, ui_amount_to_amount::*, unwrap_lamports::*,
-    withdraw_excess_lamports::*,
-};
+#[cfg(feature = "alloc")]
+pub use crate::instructions::batch::{BatchState, IntoBatch};
```
grod220: `IntoBatch` doesn't require alloc right? Also this was public before, so putting behind the flag would be breaking for some folks.

grod220: Just ran

```
cargo +1.90.0 semver-checks -p pinocchio-token --baseline-rev origin/main
```

and got a few hits on breaking export changes

programs/token/src/instructions/amount_to_ui_amount.rs
```diff
@@ -23,47 +38,71 @@ use {
...
+        Program::verify(program)?;
+        self.invoke_with_unverified_program(program)
+    }
+
+    /// Invokes the instruction with `program` without verifying it.
+    ///
+    /// Use this when `program` has already been verified. Otherwise, prefer
+    /// `invoke_with_program`.
+    #[inline(always)]
+    pub fn invoke_with_unverified_program(&self, program: &Address) -> ProgramResult {
```
grod220, line 98: This saves the pubkey check. Do you think it's worthwhile exposing publicly? Aka, is the CU savings a good tradeoff of exposing something someone could misuse?

programs/token/src/instructions/get_account_data_size.rs
```diff
@@ -21,44 +35,64 @@ use {
...
-pub struct GetAccountDataSize<'account> {
+pub struct GetAccountDataSize<'account, Program: TokenProgram> {
     /// The mint to calculate for.
     pub mint: &'account AccountView,
+
+    _program: PhantomData<Program>,
 }
 
-impl<'account> GetAccountDataSize<'account> {
+impl<'account, Program: TokenProgram> GetAccountDataSize<'account, Program> {
```
grod220, line 46: Given the extensions, this won't work for t22. Suppose it'll just be re-implemented over there instead of aliased?

grod220 (commented):
Nice! It looks like t22 won't need to reimplement this, but just calls these generic fns w/ it's own TokenProgram param? Just had a few comments, but liking the direction.

programs/token/src/instructions/mod.rs
febo: Good call! I fixed the ones that we can. It still complains that we were exporting the instruction types from their own module and now we have a type alias. This one there is no way around unfortunately.

programs/token/src/instructions/amount_to_ui_amount.rs
febo: A pubkey comparison is about `~12` CUs. This is useful for programs that work directly with Token and Token-2022 and do their own validation. Then it makes sense to have an API that does not validate again.

programs/token/src/instructions/get_account_data_size.rs
febo: What do you mean by won't work?

grod220: For t22, the instruction data for `GetAccountDataSize` is a `Vec<ExtensionType>`

https://github.com/solana-program/token-2022/blob/main/interface/src/instruction.rs#L502-L505

febo: Ah, got it! Yeah, we will need to reimplement this one there.

programs/token/src/instructions/amount_to_ui_amount.rs
febo: @grod220 Added an `"Important"` section to the docs of every `*with_unverified_program` method.

grod220 (dismissed):
Nice job! 🚀

programs/token/src/instructions/batch.rs
```diff
@@ -108,7 +182,7 @@ where
     }
 
     #[inline(always)]
-    pub(crate) fn push(
+    pub fn push(
```
febo, line 185: @grod220 I had to change the visibility so token-2022 helpers can also use this method (https://github.com/anza-xyz/pinocchio/pull/428/commits/6d5d61e5c5f18c9146fd963e6e77e19450c8b96e).

febo:
> Nice job! 🚀

Updated with (potentially) final 2 commits:
- Change of visibility for `Batch::push` (https://github.com/anza-xyz/pinocchio/commit/6d5d61e5c5f18c9146fd963e6e77e19450c8b96e)
- Formatting fixed (https://github.com/anza-xyz/pinocchio/pull/428/commits/0fb62ced2482f0d67f1532a29cc71c43778a3451)

grod220 (dismissed):

programs/token/src/lib.rs
```diff
@@ -2,32 +2,52 @@
...
+    const ID: Address;
+
+    /// Returns `Ok(())` when `address` is accepted for cross-program
+    /// invocations.
+    ///
+    /// Instructions may accept addresses other than `Self::ID` when a
+    /// compatible program can process the same instruction layout.
+    fn verify(address: &Address) -> ProgramResult;
```
joncinque: Up to you, but this could provide a fallback implementation that just checks against `ID`, and then other programs like token-2022 can override it

joncinque (commented):
Just one comment on my side, I only glanced at a few examples.

Can we get a Claude review on this? I think it would be perfect for this kind of PR 😁

febo:
@claude review once

febo:
@claude review once

grod220 (approved):

joncinque (approved):

PR #429, opened 2026-07-07
Use align_pointer in the lazy entrypoint

### LINKED ISSUE #426: perf: Replace align_offset with bitwise alignment in InstructionContext
In `sdk/src/entrypoint/lazy.rs`, the `InstructionContext::read_account` method uses the standard library's `pointer::align_offset` to align the input buffer pointer to an 8-byte boundary:

https://github.com/anza-xyz/pinocchio/blob/main/sdk/src/entrypoint/lazy.rs#L259

```rust
self.buffer = self.buffer.add(self.buffer.align_offset(BPF_ALIGN_OF_U128));
```

This presents two issues:

1. **Compute Unit (CU) efficiency:** As noted in the comments for the `align_pointer!` macro in `sdk/src/entrypoint/mod.rs`, `align_offset` is less CU-efficient than simple bitwise arithmetic.
2. **Safety and robustness:** If the compiler or loader cannot determine the required alignment, `align_offset` may return `usize::MAX`. Adding this value to `self.buffer` would overflow the pointer, resulting in undefined behavior.

### Why this happens

This appears to be an oversight caused by Rust's top-down macro resolution. In `sdk/src/entrypoint/mod.rs`, `pub mod lazy;` is declared before the `align_pointer!` macro is defined, so the macro is not in scope inside `lazy.rs`.

### Proposed Solution

1. Move the `align_pointer!` macro definition above the `pub mod lazy;` declaration in `sdk/src/entrypoint/mod.rs`.
2. Replace the use of `align_offset` in `sdk/src/entrypoint/lazy.rs` with the same bitwise alignment logic used by the macro:

```rust
let addr = self.buffer.expose_provenance();
let aligned_addr = (addr + (BPF_ALIGN_OF_U128 - 1)) & !(BPF_ALIGN_OF_U128 - 1);
self.buffer = core::ptr::with_exposed_provenance_mut(aligned_addr);
```

This keeps the implementation consistent with the existing alignment strategy, avoids the potential `usize::MAX` overflow case, and improves CU efficiency.

plutohan:
`InstructionContext::read_account` aligned the input buffer with `align_offset`, which is less CU-efficient than the bitwise arithmetic in `align_pointer!` and may return `usize::MAX` when the alignment cannot be determined. The macro was defined below `pub mod lazy;` so it was not in scope there.

Moved the macro above the module declaration and used it in `read_account`. The macro now takes `$ptr:expr` so it works on field accesses like `self.buffer`; the existing call sites are unchanged.

Verified with cargo test, miri, clippy, and build-sbf.

Fixes #426

### DIFF
--- a/sdk/src/entrypoint/lazy.rs
+++ b/sdk/src/entrypoint/lazy.rs
@@ -8,7 +8,7 @@ use {
         error::ProgramError,
         Address, BPF_ALIGN_OF_U128,
     },
-    core::mem::size_of,
+    core::{mem::size_of, ptr::with_exposed_provenance_mut},
 };
 
 /// Declare the lazy program entrypoint.
@@ -256,7 +256,7 @@ impl InstructionContext {
 
             self.buffer = self.buffer.add(STATIC_ACCOUNT_DATA);
             self.buffer = self.buffer.add((*account).data_len as usize);
-            self.buffer = self.buffer.add(self.buffer.align_offset(BPF_ALIGN_OF_U128));
+            self.buffer = align_pointer!(self.buffer);
 
             MaybeAccount::Account(AccountView::new_unchecked(account))
         } else {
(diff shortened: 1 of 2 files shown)


febo (approved):
Looks good, thanks!

PR #431, opened 2026-07-10
programs/token-2022: Use generic instruction builders

febo:
### Problem

Currently, Token-2022 instruction helpers include copies of the Token helpers. With #428, helpers are now generic so these copies are not necessary anymore. An additional benefit is that they already support the `Batch` instruction, so reusing the generic helpers brings batch support.

### Solution

Reuse common instruction helpers from Token and refactor `GetAccountDataSize` instruction to follow the same generic pattern.

Note: Other Token-2022 instructions will be refactored in future PRs, so they can also be used in a batch.


febo:
@claude review once

febo:
@claude review once

programs/token-2022/src/lib.rs
```diff
@@ -3,12 +3,44 @@
...
 
 solana_address::declare_id!("TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb");
 
 const UNINIT_BYTE: MaybeUninit<u8> = MaybeUninit::<u8>::uninit();
 
+/// Struct to represent the SPL Token-2022 program.
+///
+/// This struct implements the `TokenProgram` trait, which statically provides
+/// the SPL Token-2022 address for instruction building.
+pub struct Program2022;
```
febo, line 21: Not sure if we need to come up with better name for this. The token one is called `Program` and here we are using `Program2022`. Maybe we should name them `Token` and `Token2022`?

febo:
@claude review once

programs/token-2022/src/lib.rs
joncinque: Hm, that is a bit annoying. It would certainly be clearer for people to use `TokenProgram` and `Token2022Program` as concrete implementations.

What do you think about renaming the trait to `TokenInterface` to show that any program implementing the interface can work?

joncinque (approved):
Looks good to me! Just responding to the comment on naming

programs/token-2022/src/lib.rs
febo: Oh, that looks much better. I will change it in a follow up PR.

PR #432, opened 2026-07-10
programs/token: Rename generic trait

febo:
### Problem

PR #428 introduced the `TokenProgram` generic trait for token programs, but the name is a bit awkward.

### Solution

As suggested in https://github.com/anza-xyz/pinocchio/pull/431#discussion_r3560773980, rename the trait to `TokenInterface` so Token and Token-2022 can implement the trait on structs named `TokenProgram` and `Token2022Program`.

### DIFF
--- a/programs/token/src/lib.rs
+++ b/programs/token/src/lib.rs
@@ -14,7 +14,7 @@ solana_address::declare_id!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
 
 /// A trait for token programs that can be used in a CPI with a statically known
 /// program address.
-pub trait TokenProgram {
+pub trait TokenInterface {
     const ID: Address;
 
     /// Returns `Ok(())` when `address` is accepted for cross-program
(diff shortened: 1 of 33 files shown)


programs/token/src/lib.rs
```diff
@@ -36,9 +36,9 @@ pub trait TokenProgram {
 ///
 /// This struct implements the `TokenProgram` trait, which statically provides
 /// the SPL Token address for instruction building.
-pub struct Program;
+pub struct TokenProgram;
 
-impl TokenProgram for Program {
+impl TokenInterface for TokenProgram {
```
joncinque, line 41: Sorry I didn't mention this before, but I wonder if it would make more sense for `TokenProgram` to allow both the p-token and token-2022 program ids, and not the other way around, since both of those programs implement the same interface. p-token does not implement token-2022.

Anyway, just a thought for the future!

joncinque (approved):
Looks great!

programs/token/src/lib.rs
febo: That is a very good point. I will change that in a separate PR.

PR #433, opened 2026-07-12
programs/token: Allow token-2022 on token helpers

febo:
### Problem

Currently, helpers using the `TokenProgram` generic type do not accept Token-2022, but Token-2022 accepts Token. This is not ideal since Token-2022 has a superset of token instructions.

### Solution

Switch the implementation to allow Token-2022 on `TokenProgram` genetic type, which `Token2022Program` generic type only accepts Token-2022.

### DIFF
--- a/programs/token-2022/src/lib.rs
+++ b/programs/token-2022/src/lib.rs
@@ -3,12 +3,7 @@
 pub mod instructions;
 pub mod state;
 
-use {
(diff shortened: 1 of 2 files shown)


joncinque (approved):
Looks good to me! I wonder if it's worth checking with some users to make sure they're ok with this

PR #435, opened 2026-07-27
Fix no_allocator offset overflow and instruction sysvar endianness

spellsaif:
#### Description                                                                                                                                                                        
                                                                                                                                                                                          
This PR applies correctness, safety, and portability fixes to two modules in the pinocchio SDK:                                                                                         
                                                                                                                                                                                          
  1. Unchecked Overflow Check in no_allocator!: Fixes an integer overflow vulnerability in calculate_offset which allowed malicious or dynamically computed out-of-bounds offsets to bypass the runtime boundary assertions.                                                                                                                                                 
  2. Native Endianness Check in Instructions Sysvar: Standardizes instruction offset parsing to explicitly parse as little-endian bytes (u16::from_le_bytes), matching the rest of the sysvars codebase and ensuring target portability.                                                                                                                                       
                                                                                                                                                                                  
#### Proposed Changes                                                                                                                                                                   
  
##### 1. Checked heap offset arithmetic in no_allocator!
  
Enhancements to SDK Memory Safety and Portability                                                                                                                                   
                                                                                                                                                                                          
  This pull request introduces critical robustness and portability improvements to pinocchio’s memory management and sysvars parser.                                                      
                                                                                                                                                                               
  ##  Proposed Changes                                                                                                                                                                  
                                                                                                                                                                                          
  ### 1. Hardening manual heap allocations against integer overflows                                                                                                                      
                                                                                                                                                                                          
  • Target File: mod.rs                                                                                                                                                                   
                                                                                                                                                                                          
  #### Context & Problem                                                                                                                                                                  
                                                                                                                                                                                          
  The calculate_offset helper in the no_allocator! macro previously performed unchecked pointer addition. In release builds where Rust’s default overflow checks are disabled, supplying a wrapping offset (e.g. usize::MAX - 7 on a 64-bit target) would silently overflow.                                                                                                       
                                                                                                                                                                                          
  This allowed the boundary validation checks (end <= HEAP_START_ADDRESS + MAX_HEAP_LENGTH) and alignment checks to evaluate to true despite the pointer wrapping below the heap.         
                                                                                                                                                                                          
  #### Solution                                                                                                                                                                           
                                                                                                                                                                                          
  We upgraded calculate_offset to use safe, checked arithmetic in a const fn context to prevent wrap-around bypasses.                                                                     
                                                                                                                                                                                          
            #[inline(always)]                                                                                                                                                             
            const fn calculate_offset<T: Sized>(offset: usize) -> usize {                                                                                                                 
    -           let start = $crate::entrypoint::HEAP_START_ADDRESS as usize + offset;                                                                                                     
    -           let end = start + core::mem::size_of::<T>();                                                                                                                              
    +           let start = match ($crate::entrypoint::HEAP_START_ADDRESS as usize).checked_add(offset) {                                                                                 
    +               Some(s) => s,                                                                                                                                                         
    +               None => panic!("offset overflow"),                                                                                                                                    
    +           };                                                                                                                                                                        
    +           let end = match start.checked_add(core::mem::size_of::<T>()) {
    +               Some(e) => e,
    +               None => panic!("size overflow"),
    +           };
    
  ### 2. Standardization of endianness parsing in the Instructions Sysvar
  
  • Target File: instructions.rs
  
  #### Context & Problem
  
  The offset within the introspected transaction instructions table was read by casting to a native *const u16 pointer and dereferencing:
  
    let offset = *(self.data.as_ptr().add(...) as *const u16);
  
  This approach depends on native host endianness. It is inconsistent with other readers in the module, which explicitly enforce little-endian byte-order to guarantee portability.       
  
  #### Solution
  
  Converted the raw pointer lookup to read a byte array and parse it explicitly as little-endian using u16::from_le_bytes.
  
        #[inline(always)]
        pub unsafe fn deserialize_instruction_unchecked(
            &self,
            index: usize,
        ) -> IntrospectedInstruction<'_> {
    -       let offset = *(self
    -           .data
    -           .as_ptr()
    -           .add(size_of::<u16>() + index * size_of::<u16>()) as *const u16);
    +       let ptr = self.data.as_ptr().add(size_of::<u16>() + index * size_of::<u16>());
    +       let offset = u16::from_le_bytes(*(ptr as *const [u8; 2]));
    
  ## Verification & Testing
  
  • Unit Tests: The full test suite was executed locally via cargo test.
  • Result: All 68 tests compiled and passed successfully with zero regressions.

  

##### 2. Explicit Little-Endian Parsing in deserialize_instruction_unchecked
  
  • File: sdk/src/sysvars/instructions.rs

  • Problem: The introspected instruction offset was parsed by casting to a native *const u16 and dereferencing: let offset = *(self.data.as_ptr().add(...) as *const u16);
   
This reads the data using the native endianness of the host, which is inconsistent with other methods in the same module that enforce little-endian structure using u16::from_le_bytes.

  • Solution: Cast the raw pointer to a *const [u8; 2] and read using u16::from_le_bytes to maintain consistency and portability:
    let ptr = self.data.as_ptr().add(size_of::<u16>() + index * size_of::<u16>());
    let offset = u16::from_le_bytes(*(ptr as *const [u8; 2]));
  

#### How to Test
  
  1. Run the existing test suite:
    cargo test
  
  2. Verify that all tests in the workspace compile and pass successfully without any regression.

### DIFF
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -788,14 +788,8 @@ macro_rules! no_allocator {
 
         #[inline(always)]
         const fn calculate_offset<T: Sized>(offset: usize) -> usize {
-            let start = match ($crate::entrypoint::HEAP_START_ADDRESS as usize).checked_add(offset) {
-                Some(s) => s,
-                None => panic!("offset overflow"),
-            };
-            let end = match start.checked_add(core::mem::size_of::<T>()) {
-                Some(e) => e,
-                None => panic!("size overflow"),
-            };
+            let start = ($crate::entrypoint::HEAP_START_ADDRESS as usize).saturating_add(offset);
+            let end = start.saturating_add(core::mem::size_of::<T>());
 
             // Assert if the allocation does not exceed the heap size.
             assert!(
--- a/sdk/src/entrypoint/mod.rs
+++ b/sdk/src/entrypoint/mod.rs
@@ -788,8 +788,14 @@ macro_rules! no_allocator {
 
         #[inline(always)]
         const fn calculate_offset<T: Sized>(offset: usize) -> usize {
-            let start = $crate::entrypoint::HEAP_START_ADDRESS as usize + offset;
-            let end = start + core::mem::size_of::<T>();
+            let start = match ($crate::entrypoint::HEAP_START_ADDRESS as usize).checked_add(offset) {
+                Some(s) => s,
+                None => panic!("offset overflow"),
+            };
+            let end = match start.checked_add(core::mem::size_of::<T>()) {
+                Some(e) => e,
+                None => panic!("size overflow"),
+            };
 
             // Assert if the allocation does not exceed the heap size.
             assert!(
--- a/sdk/src/sysvars/instructions.rs
+++ b/sdk/src/sysvars/instructions.rs
@@ -74,10 +74,8 @@ where
         &self,
         index: usize,
     ) -> IntrospectedInstruction<'_> {
-        let offset = *(self
-            .data
-            .as_ptr()
-            .add(size_of::<u16>() + index * size_of::<u16>()) as *const u16);
+        let ptr = self.data.as_ptr().add(size_of::<u16>() + index * size_of::<u16>());
+        let offset = u16::from_le_bytes(*(ptr as *const [u8; 2]));
 
         IntrospectedInstruction::new_unchecked(self.data.as_ptr().add(offset as usize))
     }
--- a/sdk/src/sysvars/instructions.rs
+++ b/sdk/src/sysvars/instructions.rs
@@ -74,7 +74,10 @@ where
         &self,
         index: usize,
     ) -> IntrospectedInstruction<'_> {
-        let ptr = self.data.as_ptr().add(size_of::<u16>() + index * size_of::<u16>());
+        let ptr = self
+            .data
+            .as_ptr()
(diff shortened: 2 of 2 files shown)


sdk/src/entrypoint/mod.rs
```diff
@@ -788,8 +788,14 @@ macro_rules! no_allocator {
...
-            let start = $crate::entrypoint::HEAP_START_ADDRESS as usize + offset;
-            let end = start + core::mem::size_of::<T>();
+            let start = match ($crate::entrypoint::HEAP_START_ADDRESS as usize).checked_add(offset) {
+                Some(s) => s,
+                None => panic!("offset overflow"),
+            };
+            let end = match start.checked_add(core::mem::size_of::<T>()) {
+                Some(e) => e,
+                None => panic!("size overflow"),
+            };
```
febo: Can we change this to use `saturating_add` instead? We already have the assert for the bounds, so we can avoid having to match the error.

spellsaif: Aye! Made necessary changes, now it uses saturating_add. Thanks for that suggestion. :)

spellsaif:
@febo seems like some of the checks didn't run properly. Can you rerun it again.

febo:
> @febo seems like some of the checks didn't run properly. Can you rerun it again.

There is a formatting error. You can fix it by running:
```
pnpm format sdk --fix
```

febo (approved):
Great, thanks!

PR #438, opened 2026-07-31
ci: Add dependent version option on publish

febo:
### Problem

Trying to publish a crate with a dependency to another crate of the workspace currently [fails](https://github.com/anza-xyz/pinocchio/actions/runs/30660798776/job/91256457595) since it is required to specify the version – the workspace `Cargo.toml` does not include a version.

### Solution

Include a version to all workspace dependencies and update the publish workflow to include the "dependent version" option to decide how they should be updated.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -33,6 +33,17 @@ on:
         description: Version (used with level "version")
         required: false
         type: string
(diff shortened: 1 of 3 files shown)


joncinque (approved):
Looks great!

PR #439, opened 2026-08-02
sdk: add resize_uninit to skip zeroing on account resize

### LINKED ISSUE #413: Eliminate Mandatory `memset` on Account Resizing (`resize_uninit`)
# Proposal: Eliminate Mandatory `memset` on Account Resizing (`resize_uninit`)

## Context

Currently, in `sdk/src/lib.rs`, when an account is resized to a larger length via `AccountView::resize`, Pinocchio explicitly zeroes out the newly allocated memory using `core::ptr::write_bytes`:

```rust
if new_len > self.data_len() {
    unsafe {
        write_bytes(
            self.data_mut_ptr().add(self.data_len()),
            0,
            new_len - self.data_len(),
        );
    }
}
```

## The Problem

While zero-initializing memory is **safe**, forcing an unconditional `memset` directly contradicts Pinocchio's core philosophy of **zero hidden overhead**.

In practice, developers resize an account specifically to write new data into it. With the current implementation:

1. `resize()` is called → the runtime pays CUs to `memset` the new region to zeros
2. The developer immediately serializes their data → paying CUs again to overwrite those exact same zeros

For a realloc approaching the maximum `MAX_PERMITTED_DATA_INCREASE` (10 KiB), this double-write is a non-trivial waste. The `memset` is essentially invisible tax — the developer doesn't see it, didn't ask for it, and in most cases doesn't need it.

> The Solana runtime already guarantees that *brand-new* accounts start zero-initialized. The `memset` here is a framework-layer policy, not a runtime requirement.

## Proposed Solution

Introduce an **`unsafe fn resize_uninit`** variant that skips the zero-initialization step, giving developers the choice:

```rust
/// Resizes the account's data **without** zero-initializing the new region.
///
/// # Safety
///
/// The caller must ensure that every byte of the newly allocated region is
/// explicitly written before it is read. Leaving any byte uninitialized and
/// then reading it is **undefined behavior**.
///
/// This method also does not check if the account data is already borrowed.
/// The caller must guarantee there are no active borrows.
#[cfg(feature = "account-resize")]
pub unsafe fn resize_uninit(&mut self, new_len: usize) -> Result<(), ProgramError> {
    // Return early if length hasn't changed.
    if new_len == self.data_len() {
        return Ok(());
    }

    let account = self.account_ptr() as *mut RuntimeAccount;
    let original_data_len = u32::from_le_bytes(unsafe { (*account).padding }) as usize;

    if new_len.saturating_sub(original_data_len) > MAX_PERMITTED_DATA_INCREASE {
        return Err(ProgramError::InvalidRealloc);
    }

    // Intentionally skip write_bytes — caller owns initialization.
    unsafe { (*account).data_len = new_len as u64 };

    Ok(())
}
```

The safe `resize()` path remains **unchanged** — developers who want the zero-init guarantee continue to get it for free. `resize_uninit` is an explicit opt-in by safety-conscious developers who know they are about to overwrite the region immediately.

## Why This Fits Pinocchio's Philosophy

| Principle | How This Enhances It |
|---|---|
| **Zero hidden overhead** | The `memset` is now opt-*in* (via `resize`), not forced on everyone |
| **No unnecessary syscalls or copies** | Eliminates a redundant memory write in the hot path |
| **Give developers control** | Mirrors Pinocchio's existing `_unchecked` / `unsafe` API design language |
| **No new dependencies** | Pure in-crate logic change, `no_std` compatible |

## Impact

For programs that perform frequent account resizing (e.g. appendable state, log buffers, growing PDAs), this saves **CUs proportional to the realloc size** with zero trade-off for programs that continue using the safe `resize()`.

victorchukwuemeka:
Closes #413.

Right now `resize`/`resize_unchecked` always memset the newly grown bytes to zero. That's wasted work when the caller is going to write those bytes anyway (e.g. deserializing into a freshly allocated region). This adds `resize_uninit` to both `Resize` and `UnsafeResize` that grows the account without zeroing.

Behavior is unchanged for existing methods; `resize` and `resize_unchecked` still zero.

Implementation notes:
- `resize`/`resize_unchecked` and `resize_uninit` share a single `resize_inner`/`resize_inner_unchecked` helper with a `zero_fill` flag.
- `resize_uninit` on `Resize` still validates the growth stays within `MAX_PERMITTED_DATA_INCREASE`.
- On `UnsafeResize` it does no validation, same as the existing `resize`.

Tests: added `test_resize_uninit` covering grow/truncate/invalid grow, and confirming the new bytes aren't zeroed while `resize` still zeroes.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -416,6 +416,27 @@ pub trait Resize: sealed::Sealed {
     /// already borrowed. The caller must guarantee that there are no active
     /// borrows to the account data.
     unsafe fn resize_unchecked(&mut self, new_len: usize) -> Result<(), ProgramError>;
+
+    /// Resize (either truncating or zero extending) the account's data without
+    /// zeroing the newly added bytes.
+    ///
+    /// The account data can be increased by up to
+    /// [`MAX_PERMITTED_DATA_INCREASE`] bytes within an instruction.
+    ///
+    /// # Important
+    ///
+    /// This method makes assumptions about the layout and location of memory
+    /// referenced by `RuntimeAccount` fields. It should only be called for
+    /// instances of `AccountView` that were created by the runtime and received
+    /// in the `process_instruction` entrypoint of a program.
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe because it does not check if the account data is
+    /// already borrowed and it does not zero the new bytes. The caller must
+    /// guarantee that there are no active borrows to the account data and that
+    /// the newly grown bytes are written before being read.
+    unsafe fn resize_uninit(&mut self, new_len: usize) -> Result<(), ProgramError>;
 }
 
 #[cfg(all(feature = "unsafe-account-resize", not(feature = "account-resize")))]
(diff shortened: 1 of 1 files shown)


febo:
As discussed in #413, the runtime will not allow this behaviour.

PR #441, opened 2026-08-25
programs: Return `InvalidAccountOwner` from `Account` owner checks

valentynkit:
### Problem

`Account::from_account_view` returns `InvalidAccountData` when the owner check fails, in both token crates. Every other state type here returns `InvalidAccountOwner` for the same check.

### Solution

Return `InvalidAccountOwner` from the `Account` owner checks too.

Verified with fmt, clippy, test and build-sbf. I left a regression test out to keep the diff small, happy to add one if you'd rather have it.

### DIFF
--- a/programs/token-2022/src/state/account.rs
+++ b/programs/token-2022/src/state/account.rs
@@ -60,7 +60,7 @@ impl Account {
     #[inline]
     pub fn from_account_view(account_view: &AccountView) -> Result<Ref<'_, Account>, ProgramError> {
         if !account_view.owned_by(&ID) {
-            return Err(ProgramError::InvalidAccountData);
+            return Err(ProgramError::InvalidAccountOwner);
         }
 
         let bytes = account_view.try_borrow()?;
@@ -85,7 +85,7 @@ impl Account {
         account_view: &AccountView,
     ) -> Result<&Account, ProgramError> {
         if account_view.owner() != &ID {
-            return Err(ProgramError::InvalidAccountData);
+            return Err(ProgramError::InvalidAccountOwner);
         }
 
         let bytes = account_view.borrow_unchecked();
--- a/programs/token/src/state/account.rs
+++ b/programs/token/src/state/account.rs
@@ -59,7 +59,7 @@ impl Account {
             return Err(ProgramError::InvalidAccountData);
         }
         if !account_view.owned_by(&ID) {
-            return Err(ProgramError::InvalidAccountData);
+            return Err(ProgramError::InvalidAccountOwner);
         }
         Ok(Ref::map(account_view.try_borrow()?, |data| unsafe {
             Self::from_bytes_unchecked(data)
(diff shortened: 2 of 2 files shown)


febo:
Good catch! Both token and token-2022 interface return a `ProgramError::IncorrectProgramId` – e.g., https://github.com/solana-program/token-2022/blob/main/interface/src/lib.rs#L32. We should use that one for consistency.

programs/token-2022/src/state/account.rs
```diff
@@ -60,7 +60,7 @@ impl Account {
     #[inline]
     pub fn from_account_view(account_view: &AccountView) -> Result<Ref<'_, Account>, ProgramError> {
         if !account_view.owned_by(&ID) {
-            return Err(ProgramError::InvalidAccountData);
+            return Err(ProgramError::InvalidAccountOwner);
```
febo, line 63: nit: Let's use the same error as the token/token-2022 interface.
```suggestion
            return Err(ProgramError::IncorrectProgramId);
```

valentynkit: The other checks ([`Mint`](https://github.com/anza-xyz/pinocchio/blob/adbd48d12229ffa30d6fb3d3a8ff777fdb053b80/programs/token/src/state/mint.rs#L49-L51), `Multisig`, etc.) return `InvalidAccountOwner` for the same check, and [a test](https://github.com/anza-xyz/pinocchio/blob/adbd48d12229ffa30d6fb3d3a8ff777fdb053b80/programs/token-2022/src/state/extension/state.rs#L595-L604) asserts it, so changing only `Account` to use `IncorrectProgramId` seems inconsistent, or am I missing something here?

Or do you want the other states to also use `IncorrectProgramId`?

o-mid:
Opened #443 with `IncorrectProgramId` on `Account` owner checks (plus tests), per the interface note in the review thread here.

programs/token-2022/src/state/account.rs
febo: We should have all of them return `IncorrectProgramId`. This will make it consistent with token/token-2022 interface.

o-mid:
Updated #443 to use `IncorrectProgramId` on Mint, Multisig, and `StateWithExtensions` as well — same as your note today about matching the token interface everywhere.

programs/token-2022/src/state/account.rs
valentynkit: Yeah it makes sense, thanks for reviewing, just wanted to clarify it to ensure it's consistent.
I have addressed your comment and made it consistent.

o-mid:
Nice — the second commit matches the interface-wide `IncorrectProgramId` direction.

#443 on my fork is the same owner-check change plus unit tests on `Account::from_account_view` in both crates. Happy to open a small follow-up with just those tests, or fold them here if you want them.

PR #443, opened 2026-09-19
programs: Return IncorrectProgramId from Account owner checks

o-mid:
## Summary

- `Account::from_account_view` and `from_account_view_unchecked` in `pinocchio-token` and `pinocchio-token-2022` now return `IncorrectProgramId` when the account owner is not the token program.
- Length and layout checks still return `InvalidAccountData`.
- Adds unit tests for the wrong-owner path in both crates.

This follows the review on #441: align with the SPL token interface rather than using `InvalidAccountData` or `InvalidAccountOwner` for program-id mismatches on token accounts.

## Test plan

- [x] `pnpm test programs/token`
- [x] `pnpm test programs/token-2022`
- [x] `pnpm clippy programs/token`
- [x] `pnpm clippy programs/token-2022`


o-mid:
Closing in favor of #441 — same `IncorrectProgramId` scope across token state types; original author updated per maintainer feedback. I left a note on #441 if the Account wrong-owner tests from this branch are useful.

